PR #13, opened 2015-01-11
Implement basic `route get` functionality

dave-tucker:
This provides the basic functionality of the iproute2 `route get`
command. It allows you to query netlink for the route to a destination
IP.Net

While implementing I noticed that the `s.Receive()` function was hanging
after the netlink response was recieved, so I added an additional clause
to break from the loop and return the decoded message if the NLM_F_MULTI
flag is not set.

Signed-off-by: Dave Tucker <dave@dtucker.co.uk>

### DIFF
--- a/nl/nl_linux.go
+++ b/nl/nl_linux.go
@@ -248,6 +248,9 @@ done:
 				continue
 			}
 			res = append(res, m.Data)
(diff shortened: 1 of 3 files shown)


erikh:
+1, would like to see this too.

PR #21, opened 2015-02-10
Hairpin support

LK4D4:
This adds basic support for hairpin mode. Deserializing is pretty ugly, but I don't know how to get `IFLA_PROTINFO` without `NLM_F_DUMP` :(

### DIFF
--- a/nl/link_linux.go
+++ b/nl/link_linux.go
@@ -51,6 +51,25 @@ const (
 )
 
 const (
+	BRIDGE_MODE_UNSPEC  = iota
+	BRIDGE_MODE_HAIRPIN = iota
+)
+
+const (
+	IFLA_BRPORT_UNSPEC        = iota
+	IFLA_BRPORT_STATE         = iota
+	IFLA_BRPORT_PRIORITY      = iota
+	IFLA_BRPORT_COST          = iota
+	IFLA_BRPORT_MODE          = iota
+	IFLA_BRPORT_GUARD         = iota
+	IFLA_BRPORT_PROTECT       = iota
+	IFLA_BRPORT_FAST_LEAVE    = iota
+	IFLA_BRPORT_LEARNING      = iota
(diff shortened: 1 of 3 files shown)


vishvananda:
LinkList already calls NLM_F_DUMP so you should be able to deserialize directly there. For the LinkByName and LinkByIndex calls is it possible to add NLM_F_DUMP to those to get it to return PROTINFO?

LK4D4:
@vishvananda Thanks for answer, you're right about List, I'll try to rewrite. If I'll add `NLM_F_DUMP` to calls with `Index`, then index seems ignored and there will be returned all devices with `AF_UNSPEC`, so our assumption about `len(msgs) == 1` would be wrong. I can rewrite all calls to use `NLM_F_DUMP`, but the there will be overhead on receiving info about all devices. I wonder if there is way to get `PROTINFO` without `NLM_F_DUMP`, but I am pretty bad kernel-source reader and can't find anything :(

LK4D4:
@vishvananda Hmm, problem with calling `LinkList` is that it uses `AF_UNSPEC` and we need `AF_BRIDGE` for `PROTINFO`.

vishvananda:
@LK4D4 wow looking at the iproute2 code for bridge/link.c it does appear that it just always dumps the whole list even when a single item is requested. This is nasty. I'm wondering if we should just have a separate setHairpinMode and getHairpinMode in this case instead of trying to stick it as an attribute of link

LK4D4:
@vishvananda might be cleaner, you're right. But now question, that `PROTINFO` contains of several fields. Do you want me to create `Protinfo` datastructure for return from some `LinkGetProtinfo`?

LK4D4:
@vishvananda I changed code to treat `IFLA_PROTINFO` as separate entity.

vishvananda:
nice that seems better, will merge it

PR #22, opened 2015-02-11
IPVlan support

LK4D4:


nl/link_linux.go
```diff
@@ -70,6 +70,20 @@ const (
 )
 
 const (
+	IFLA_IPVLAN_UNSPEC = iota
+	IFLA_IPVLAN_MODE   = iota
+	IFLA_IPVLAN_MAX    = IFLA_IPVLAN_MODE
+)
+
+type IPVlanMode uint16
```
vishvananda: This type should be in link.go since it is a higher level type used by the netlink library not the lower level nl library (see type Mode uint8 in xfrm.go for example).

link_test.go
```diff
@@ -5,6 +5,7 @@ import (
 	"net"
 	"testing"
 
+	"github.com/vishvananda/netlink/nl"
```
vishvananda: shouldn't need to import nl in the test here once you have made the change below

link.go
```diff
@@ -2,6 +2,8 @@ package netlink
 
 import (
 	"net"
+
+	"github.com/vishvananda/netlink/nl"
```
vishvananda: no need to import nl here

vishvananda:
thanks for the code. A few suggestions inline

LK4D4:
@vishvananda Fixed

link_test.go
```diff
@@ -81,6 +81,16 @@ func testLinkAddDel(t *testing.T, link Link) {
 		compareVxlan(t, vxlan, other)
 	}
 
+	if ipv, ok := link.(*IPVlan); ok {
+		other, ok := result.(*IPVlan)
+		if !ok {
+			t.Fatal("Result of create is not a vxlan")
```
vishvananda: small typo here but good enough. I will merge it

PR #24, opened 2015-02-12
Use tristate(negative/0/positive) fields for Protinfo

LK4D4:
I noticed that it is not very convenient to use default bool value,
because IFLA_PROTINFO fields is not always false by default.

### DIFF
--- a/protinfo.go
+++ b/protinfo.go
@@ -7,27 +7,45 @@ import (
 	"github.com/vishvananda/netlink/nl"
 )
 
+// Protinfo is abstraction under IFLA_PROTINFO.
+// Negative fields is treated as "unset" and won't be sent to netlink.
 type Protinfo struct {
-	Hairpin   bool
-	Guard     bool
-	FastLeave bool
-	RootBlock bool
-	Learning  bool
-	Flood     bool
+	Hairpin   int
+	Guard     int
+	FastLeave int
+	RootBlock int
+	Learning  int
+	Flood     int
 }
 
-func boolToByte(x bool) []byte {
-	if x {
+// NewProtinfo returns Protinfo with all fields unset
+// Example of usage(enables hairpin, disables learning):
+// pi := NewProtinfo()
+// pi.Hairpin = 1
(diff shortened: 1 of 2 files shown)


vishvananda:
I don't love the -1s representing not set. It seems like it would be more accurate to use *bool and leave them null if they are not set. Thoughts?

LK4D4:
@vishvananda Maybe you're right that -1 is not cool, but pointers is not very convenient to use. Then we probably need make setters and getters for each field. Or I can create tristate type with `Unset, Yes, No` values or something like this. What do you think?

LK4D4:
Now I think that setters/getters + MarshalJSON method maybe not bad option. Let me know your opinion about that.

vishvananda:
how about something like a specialized type like:

https://code.google.com/p/go/source/browse/cmd/vet/main.go?repo=tools#53

LK4D4:
@vishvananda Seems like not convenient to use, compare
```
protinfo.Hairpin = netlink.PROTINFO_TRUE
protinfo.Hairpin = netlink.PROTINFO_FALSE
protinfo.Hairpin = netlnik.PROTINFO_UNSET
```
with
```
protinfo.SetHairpin(true)
protinfo.SetHairpin(false)
protinfo.UnsetHairpin()
```

vishvananda:
the issue with setters is that none of the other fields have Setters and it is confusing that link.SetMTU(9000) actually makes a netlink call to set the mtu whereas protinfo.SetHairpin(true) would just be setting the field. Maybe what we really want is a command on link for each of the fields:

link.GetHairpin() *bool (or tristate)
link.SetHairpin(true)
link.SetHairpin(false)
...

rather than link.SetProtInfo/link.GetProtInfo

LK4D4:
@vishvananda Maybe you're right, personally for my case I don't care about +-3 calls to netlink for setting multiple fields in Protinfo. I'll open another PR about that. Should simplify interface for user. Problem only that getting that attributes will be different from for example `MTU` anyway :(

PR #25, opened 2015-02-17
Change logic of setting protinfo attrs

LK4D4:
Now interface is supposing that you setting protinfo attrs for link one
by one. But you can get all protinfo attrs with one call.


vishvananda:
I like this much better

PR #27, opened 2015-02-24
Added ability to specify net ns at link creation

eyakubovich:
Creating a link in one namespace and then moving can
be problematic since there could be an iface name conflict.
However, it is not always possible to switch into the target
namespace either -- e.g. creating a macvlan interface must
be done in the namespace of the parent interface.

### DIFF
--- a/link.go
+++ b/link.go
@@ -10,6 +10,11 @@ type Link interface {
 	Type() string
 }
 
+type (
+	NsPid int
+	NsFd  int
+)
+
 // LinkAttrs represents data shared by most link types
 type LinkAttrs struct {
 	Index        int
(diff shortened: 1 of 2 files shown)


LK4D4:
@eyakubovich Is it possible to create veth peer inside namespace directly too?

eyakubovich:
@LK4D4 Not that I know of.

eyakubovich:
Just found out that while this is PR helps with not moving the interface into its destined namespace, it does not help with name conflict. When adding a new link, the kernel will first lookup the device by name (if ifi_index is not given) in the current net namespace. It will then error out with EEXIST if there's one and NLM_F_EXCL is set. I view this as a bug in the kernel but I don't have a good proposal on how to fix it.

LK4D4:
@eyakubovich Pretty inconvenient thing :/

vishvananda:
merged as 1fd3169564abccd94f0e2786160ab294832a498b

PR #29, opened 2015-02-25
added support for MACVLAN mode

eyakubovich:


link_linux.go
```diff
@@ -13,6 +13,14 @@ import (
 var native = nl.NativeEndian()
 var lookupByDump = false
 
+var macvlanModes = [...]uint32{
```
LK4D4: do we really need this? Why not just explicitly cast `MacvlanMode` to `uint32`?

eyakubovich: I'm just following the convention where the constants from Linux headers are in `nl` package. We also need to ensure that 0 value maps to a noop. That's technically possible in this case as the constants start with 1 (for some reason they look like they can be combined in a bitmask), however it seemed a bit magical to do that. But I'm open to getting rid of double constants.

LK4D4: Hehe, actually I already broke that convention in ipvlan patch :) I personally prefer to keep code simpler. But it is up to @vishvananda of course

vishvananda:
One advantage of separating it out is constants are not the same across architectures. Even 386 vs. x86_64 has different constants for syscalls in linux.

eyakubovich:
Other than syscalls, most constants are the same on different architectures. However, it's best to keep them separate for those rare cases.

LK4D4:
I looked at `iproute2` source and noticed, that there is one more mode:
```
MACVLAN_MODE_SOURCE  = 16
```
Also, seems like they use hardcoded values for constants.

eyakubovich:
Added support for MACVLAN_MODE_SOURCE

eyakubovich:
@vishvananda Ping

PR #34, opened 2015-05-10
prefer IFA_LOCAL address when dumping addresses

eyakubovich:
IFA_LOCAL and IFA_ADDRESS are the same for bcast interfaces,
however for point-to-point links, IFA_ADDRESS is the remote
address while IFA_LOCAL is the interface address.


eyakubovich:
ping @vishvananda

PR #35, opened 2015-06-12
Add LinkAttrs constructor NewLinkAttrs

### LINKED ISSUE #26: TxQLen is zero by default
But actually seems like it should be 1000. I'm not sure how I can fix this, should I implement `SetTxQLen` and move it out of `Attrs`?
@vishvananda WDYT?

LK4D4:
It looks little awkward, but should be enough for people who care about
defaults from kernel.

Fix #26

### DIFF
--- a/link.go
+++ b/link.go
@@ -19,13 +19,20 @@ type (
 type LinkAttrs struct {
 	Index        int
 	MTU          int
-	TxQLen       uint32 // Transmit Queue Length
+	TxQLen       int // Transmit Queue Length
 	Name         string
 	HardwareAddr net.HardwareAddr
 	Flags        net.Flags
 	ParentIndex  int         // index of the parent link device
 	MasterIndex  int         // must be the index of a bridge
 	Namespace    interface{} // nil | NsPid | NsFd
+}
+
+// NewLinkAttrs returns LinkAttrs structure filled with default values
+func NewLinkAttrs() LinkAttrs {
+	return LinkAttrs{
+		TxQLen: -1,
+	}
 }
 
 // Device links cannot be created via netlink. These links
--- a/README.md
+++ b/README.md
@@ -43,13 +43,19 @@ import (
 )
 
 func main() {
-    mybridge := &netlink.Bridge{netlink.LinkAttrs{Name: "foo"}}
+    la := netlink.NewLinkAttrs()
+    la.Name = "foo"
+    mybridge := &netlink.Bridge{la}}
     _ := netlink.LinkAdd(mybridge)
     eth1, _ := netlink.LinkByName("eth1")
     netlink.LinkSetMaster(eth1, mybridge)
 }
 
 ```
+Note `NewLinkAttrs` constructor, it sets default values in structure. For now
+it sets only `TxQLen` to `-1`, so kernel will set default by itself. If you're
+using simple initialization(`LinkAttrs{Name: "foo"}`) `TxQLen` will be set to
+`0` unless you specify it like `LinkAttrs{Name: "foo", TxQLen: 1000}`.
 
 Add a new ip address to loopback:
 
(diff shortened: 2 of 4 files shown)


LK4D4:
ping @vishvananda @eyakubovich

link.go
```diff
@@ -28,6 +28,14 @@ type LinkAttrs struct {
...
 }
 
+// NewLinkAttrs returns LinkAttrs structure filled with default values
+func NewLinkAttrs() LinkAttrs {
+	return LinkAttrs{
+		TxQLen: -1,
+		MTU:    -1,
+	}
+}
+
```
eyakubovich: You're checking for `> 0` below so why not have "not set" be `== 0`. This way, default initialization still works and `NewLinkAttrs()` is not needed.

eyakubovich:
+1 for checking for `> 0` before setting the attribute.

link.go
LK4D4: Hmm, my mistake. I need `>=0`. Is it true for MTU too?

LK4D4: No, I got it. For `MTU` 0 is sorta invalid.

eyakubovich: `TxQLen` can be 0?

LK4D4: Yeah, at least I can set it with `iproute2`.

LK4D4: But that's weird, I can't set 0 actually with our lib :/

eyakubovich: Looks like TxQLen == 0 has important special meaning in some drivers. Actually, same for MTU (basically says default to lower dev MTU). I guess we need `NewLinkAttrs`. @vishvananda thoughts?

LK4D4: Hmm, I'm not sure, but seems like we can't set `TxQLen` to `0` with netlink. `iproute2` uses `ioctl`.

vishvananda: If we can't actually set zero for either then we could just leave zero as
the default, but I thought the initial bug was that it doesn't default to
10000

Vish

LK4D4: https://github.com/torvalds/linux/blob/master/net/core/rtnetlink.c#L1702

LK4D4: I was pretty sure, that we can :/ Actually with `TxQLen` all works as expected, apart from setting it to `0`.

eyakubovich: But from that kernel code, it looks like 0 is ok and should be set. `tb[IFLA_TXQLEN` is just `rtattr *`

LK4D4: Yes, sorry, I'll retry my tests. TxQLen should be set.

LK4D4:
Okay, I think I found it. For veth pair, `TxQLen` is set to 1000 by default, but with netlink you're changing only peer `TxQLen`. For `dummy` type default is 0.

LK4D4:
@eyakubovich @vishvananda Thanks for help!
Seems like it never worked! I changed code and added some tests.
MTU seems can't be zero, it returns `EINVAL` for Veth creation for example.

eyakubovich:
With veth, the main device and its peer are two separate devices (they each have ifindex). So you can't change TxQLen on them individually? Or did you mean that at creation time, TxQLen affects only peer.

LK4D4:
No, I was wrong. Logic of txqlen was broken for main device. 
Also you can set different qlens for main and peer,  you can try with iproute2.

LK4D4:
And original test for txqlen was false positive. I think all should be OK now.

vishvananda:
The code in the tests and the readme should be changed to use NewLinkAttrs. Unfortunately this means the nice one liners have to go away.

One option for allowing one liners would be to allow the New to take and return a pointer to a LinkAttrs that modifies the default:

http://play.golang.org/p/d4ZYrukfDE

but it has the drawback that NewLinkAttrs(&LinkAttrs{TxQLen: 0}) doesn't work as expected.

LK4D4:
People still can use oneliners if they will set TxQLen explicitly. I'll write about this in Readme.

LK4D4:
@vishvananda I added info to readme, but changing tests is a lot of work and not really worth, because we really don't care about `TxQLen` in most tests.

PR #36, opened 2015-06-14
Use sync.Pool instead of each time allocation

LK4D4:
We allocate 4kb of space each time and this space won't be collected by GC until we remove references to syscall.NetlinkMessages, because it stores parts of original slice. Also allocation of 4kb is unneeded every time and in common cases makes call slower.
This patch though makes receiving big netlink messages slightly slower(because of copying answer to new slice), but allows to save significant amount of memory.
Here results of comparing benchmarks:
```
benchmark                 old ns/op     new ns/op     delta
BenchmarkLinkAddDel-4     6860622       7510755       +9.48%
BenchmarkLinkList-4       59623         52799         -11.45%

benchmark                 old allocs     new allocs     delta
BenchmarkLinkAddDel-4     138            144            +4.35%
BenchmarkLinkList-4       39             41             +5.13%

benchmark                 old bytes     new bytes     delta
BenchmarkLinkAddDel-4     34632         14237         -58.89%
BenchmarkLinkList-4       13824         8301          -39.95%
```

### DIFF
--- a/nl/nl_linux.go
+++ b/nl/nl_linux.go
@@ -6,6 +6,7 @@ import (
 	"encoding/binary"
 	"fmt"
 	"net"
+	"sync"
 	"sync/atomic"
 	"syscall"
 	"unsafe"
@@ -329,8 +330,18 @@ func (s *NetlinkSocket) Send(request *NetlinkRequest) error {
 	return nil
 }
 
+var rcvPool = &sync.Pool{
+	New: func() interface{} {
+		return make([]byte, syscall.Getpagesize())
+	},
+}
+
 func (s *NetlinkSocket) Receive() ([]syscall.NetlinkMessage, error) {
-	rb := make([]byte, syscall.Getpagesize())
+	rb, ok := rcvPool.Get().([]byte)
+	if !ok {
+		return nil, fmt.Errorf("receive buffers pool error")
+	}
+	defer rcvPool.Put(rb)
 	nr, _, err := syscall.Recvfrom(s.fd, rb, 0)
 	if err != nil {
 		return nil, err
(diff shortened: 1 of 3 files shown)


nl/nl_linux.go
```diff
@@ -338,8 +349,9 @@ func (s *NetlinkSocket) Receive() ([]syscall.NetlinkMessage, error) {
 	if nr < syscall.NLMSG_HDRLEN {
 		return nil, fmt.Errorf("Got short response from netlink")
 	}
-	rb = rb[:nr]
-	return syscall.ParseNetlinkMessage(rb)
+	data := make([]byte, nr)
```
vishvananda: Doesn't seem like we are saving much if we have to make a new slice right here. Can't you just return syscall.ParseNetlinkMessage(rb[:nr])?  I'm not sure it is worth it to use a pool for this otherwise.

LK4D4: Yeah, that's what's bad about `syscall.ParseNetlinkMessage`. It holds references to original slice in returned arguments and original slice won't be collected by garbage collector until those references not disappear. I'm not sure, maybe you can tell, but seems like typical netlink responses are much smaller than `syscall.Getpagesize()` and I show in benchmark that at least for `Add` and `Del` it's true. I can test `nr` and see if it's worth to create new slice and return original to pool, or just use it and never return to pool if you think that it's worth it.
Do you know some netlink call with massive response so I can write benchmark for that case.

LK4D4:
I did some measurements with docker and it seems that we have 70-99% of thrash memory on each call, which is pretty bad in Go, because memory released to OS not immediately. What if I introduce threshold(like 1.5k) for using slice from pool, so in worst case we will have 60% of thrash memory with decent performance(I think it should be pretty equal to old performance).

vishvananda:
I'm starting to think the optimization isn't really worth it if we are just saving a few bytes of memory usage.

LK4D4:
@vishvananda It's almost always around 4kb per netlink call. But maybe you're right, it's too slippery way.

PR #40, opened 2015-06-24
New netlink functionality

krhubert:
- Bond Interface
- Veth link peer index
- RouteFilter
- add Rule
- Docs
- More option for address and routes


addr.go
```diff
@@ -34,6 +41,30 @@ func ParseAddr(s string) (*Addr, error) {
...
 }
 
+// ParseAddr parses the string representation of an address in the form $ip/$netmask.
+func ParseAddr(s string) (*Addr, error) {
+	m, err := ParseIPNet(s)
+	if err != nil {
+		return nil, err
+	}
+	return &Addr{IPNet: m}, nil
+}
```
vishvananda: Why modify this when the old ParseAddr worked just fine without a label?

addr_test.go
```diff
@@ -27,7 +27,7 @@ func TestAddrAddDel(t *testing.T) {
 		t.Fatal(err)
 	}
 
-	if len(addrs) != 1 || !addr.Equal(addrs[0]) || addrs[0].Label != addr.Label {
+	if len(addrs) != 1 || !addr.Equal(addrs[0]) {
```
vishvananda: seems odd to remove the address part of the test. If we really are renaming ParseAddr to ParseAddrWithLabel, we chould use ParseAddrWithLabel in the test, but I'm not convinced it needs to change

link.go
```diff
@@ -41,10 +49,12 @@ type Device struct {
 	LinkAttrs
 }
 
+// Attrs implementation.
 func (device *Device) Attrs() *LinkAttrs {
 	return &device.LinkAttrs
 }
 
+// Type implementation fro Device.
```
vishvananda: typo

link.go
```diff
@@ -78,19 +92,23 @@ func (bridge *Bridge) Type() string {
...
-	VlanId int
+	Id int
 }
 
+// Attrs implementation.
 func (vlan *Vlan) Attrs() *LinkAttrs {
 	return &vlan.LinkAttrs
 }
 
+// Type implementation fro vlan.
```
vishvananda: typo fro

link.go
```diff
@@ -117,13 +137,17 @@ func (macvlan *Macvlan) Type() string {
...
+	// TODO: implelemnt (via IOCTL) get index while parse veth
+	PeerIndex int // veth: read only
 }
 
+// Attrs implementation.
 func (veth *Veth) Attrs() *LinkAttrs {
 	return &veth.LinkAttrs
 }
 
+// Type implementation fro Veth.
```
vishvananda: looks like a s/fro/for would be good on this file

protinfo_linux.go
```diff
@@ -19,38 +20,38 @@ func LinkGetProtinfo(link Link) (Protinfo, error) {
 		return pi, err
 	}
 
-	for _, m := range msgs {
-		ans := nl.DeserializeIfInfomsg(m)
+	for i := range msgs {
```
vishvananda: is there a reason you changed all of these? I don't see any functional changes.

xfrm_state_linux.go
```diff
@@ -119,8 +119,8 @@ func XfrmStateList(family int) ([]XfrmState, error) {
 	}
 
 	var res []XfrmState
-	for _, m := range msgs {
-		msg := nl.DeserializeXfrmUsersaInfo(m)
+	for i := range msgs {
```
vishvananda: once again, why change this?

route.go
```diff
@@ -3,33 +3,130 @@ package netlink
...
-// Scope is an enum representing a route scope.
-type Scope uint8
-
-const (
-	SCOPE_UNIVERSE Scope = syscall.RT_SCOPE_UNIVERSE
-	SCOPE_SITE     Scope = syscall.RT_SCOPE_SITE
-	SCOPE_LINK     Scope = syscall.RT_SCOPE_LINK
-	SCOPE_HOST     Scope = syscall.RT_SCOPE_HOST
-	SCOPE_NOWHERE  Scope = syscall.RT_SCOPE_NOWHERE
-)
```
vishvananda: Is there a reason you changed Scope from an enum? I like having the values here so the caller doesn't have to use the syscall package to pass values in. Tables and Protocols can be done the same way and the string methods can be directly on the enum. See how it is done in xfrm.go: https://github.com/vishvananda/netlink/blob/master/xfrm.go#L39

route.go
```diff
@@ -3,33 +3,130 @@ package netlink
...
 // gateway. Advanced route parameters and non-main routing tables are
 // currently not supported.
 type Route struct {
-	LinkIndex int
-	Scope     Scope
-	Dst       *net.IPNet
-	Src       net.IP
-	Gw        net.IP
+	Iif      int
+	IifName  string
```
vishvananda: If i understand correctly IifName and OifName are not settable properties on the route, so I'm thinking they don't belong here. Thoughts?

vishvananda:
Thanks a lot for all of this work! I have some comments and questions inline. The changes everywhere from for _, x := range to for i := range seem odd. Is there any way you could squash down the commits and propose the changes to existing files separately from new functionality? That would make it much easier to review.

addr.go
krhubert: When you ricive data from other source (like rpc call, shell input etc...), you want to check if the network address is correct and you dont want to use net.ParseIP followed with netlinka.Addr{...}, or netlink.ParseAddr. I just want to keep all network stuff in netlink package, this was the purpose of spliting this into two functions, what do you think about this ?

protinfo_linux.go
krhubert: Just to speed it up a littel bit and avoid copies. When you have for example few hundrest interfaces, or 300k routes, it speed up a littel bit, and it was really small change.

xfrm_state_linux.go
krhubert: same as above, i've changed all "for _, m :=range msgs" for the same reason

route.go
krhubert: I just don't want to cover the syscall pkg, becouse there are way more constans in syscall package that must be put in netlink pkg (for routes, rules, interfaces etc...), and i just thougth that netlink should be cover all syscall constans or none of them. But none of them is easier and i think more intuitive. For example values for FAMILY_V4, V6 are for me unintuitive, everyone who has programming some network stuff used to use AF_INET, AF_INET6 etc in all languages. Why introduce new naming convention for this values? In fact I want to remove them as well, but thought that maybe more peopel use it (more then Scope enum).

krhubert:
You're welcome. I know code isn't finished yet, there are many things to refactor, and few bugs was found since I put the pull request here, but I want to share it with others so it could be tested and develpoed more and more. About spliting, it would be hard for few days but I have some work todo, but as soon as I'll be free i will write to you about this commits. You can ask me any time if you have some questions about those commits :)

addr.go
krhubert: For example valid network address for interface "eth0" , user may enter:
10.16.0.1/32 xxx
And netlink.ParseAddr would worked fine, but AddrAdd will not. And there is need for just valid network address only

krhubert:
Function uses ioctl needs a lot of refactor and I'm working on it as well. Also I thought about spliting functions into more files (like link_eth, link_veth, link_bond for example) becouse it becomes to large and to illegible.

protinfo_linux.go
LK4D4: msgs has reference type anyway, there is no copy here
and you should avoid premature optimizations, sacrificing readability

vishvananda: I looked this up today and according to the internet it isn't a reference
http://stackoverflow.com/questions/20185511/golang-range-references-instead-values

route.go
krhubert: If i understand correctly IifName and OifName are not settable properties on the route => right
so I'm thinking they don't belong here. Thoughts? => that's true, but i don't know where to put it. Maybe it shouldn't be there and only index should be left as It was.

protinfo_linux.go
LK4D4: Underlying type is []byte, which is reference. Same will be for []*anyType.

route.go
vishvananda: My general hope is that people could use the library without importing syscall. I'd be ok with a pr to remove the shadowing if people think it is a good idea, but we should do everything at once if we do it.

protinfo_linux.go
vishvananda: Ah, good point. I'm thinking there is no reason to change it in that case

addr.go
vishvananda: I'm still not following the logic here. Which AddrAdd takes a string? and why would you pass the same string to both of them?

vishvananda:
@krhubert any chance of getting an update on this?

krhubert:
Hi, sorry for so long answer, 
now i try to find some time to split this functionality into smaller pull request and push it to you

PR #41, opened 2015-07-10
fixing issue with setting big endian

brahmaroutu:
Signed-off-by: Srini Brahmaroutu <srbrahma@us.ibm.com>

### DIFF
--- a/nl/nl_linux.go
+++ b/nl/nl_linux.go
@@ -39,8 +39,9 @@ func NativeEndian() binary.ByteOrder {
 		var x uint32 = 0x01020304
 		if *(*byte)(unsafe.Pointer(&x)) == 0x01 {
 			nativeEndian = binary.BigEndian
+		} else {
+			nativeEndian = binary.LittleEndian
 		}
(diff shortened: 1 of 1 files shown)


vishvananda:
Can you explain what the deal is here? Is the Big Endian handling completely unnecessary? should we remove it altogether?

brahmaroutu:
@vishvananda Without the else block it is never set to Big Endian when testing Docker on IBM Z. This change is helping me to set to that value.

vishvananda:
Is the ibmz big endian? If so I'm thinking that the big endian code is
unnecessary and we should just remove all the endianness checks.

brahmaroutu:
yep I need to support big endian. I am trying to follow why you did not want to use else block? Not sure why we need endian check, I do not know this code well enough to comment. For sure the code is not working on IBM Z without setting it.

brahmaroutu:
@vishvananda I hope you can merge this PR, we need this for Docker work on Z. Please let me know if you see any issues.

brahmaroutu:
Thanks for merging this PR

Issue #38, opened 2015-06-19
msg.Change value


LK4D4:
0207b6b6339670b6f4e9ddd2c2e72ce8d567969e changed behavior a little.
From `man 7 rtnetlink`:
```
ifi_change is reserved for future use and should be always set to 0xFFFFFFFF.
```
Also when set to `0xFFFFFFFF` any netlink call sets interface up(have no idea why). So, I'm not sure what sense make set `msg.Change` to anything. It works with any value, but with `0xFFFFFFFF` it also sets interface up :)
ping @vishvananda @marek-polewski

LK4D4:
I looked in libnl code and `ifi_change` isn't used there at all.

vishvananda:
That is odd. I'm curious about the issue with it overwriting flags on SetLink though that @marek-polewski referenced in the pull request

LK4D4:
@vishvananda Odd indeed, and I'm not sure what expected behavior.

vishvananda:
Fixed by #44

PR #44, opened 2015-08-14
Do not set nlmsghdr data values in ifinfomsg fields

mrjana:
Currently the netlink link_linux.go sets all sorts of nlmsghdr
values like RTM_NETLINK/RTM_SETLINK, NLM_F_REQUEST and IFLA_*
values in ifinfomsg.if_type, ifinfomsg.if_flags and ifinfomsg.if_change respectively.
This is totally incorrect since they are meant to send through to
the kernel interface specific information and not generic netlink
data. Because of this the kernel is treating them for their
expected meaning and more than likely causing a lot of unintentional
consequences.

Signed-off-by: Jana Radhakrishnan <mrjana@docker.com>

### DIFF
--- a/link_linux.go
+++ b/link_linux.go
@@ -73,10 +73,7 @@ func LinkSetMTU(link Link, mtu int) error {
 	req := nl.NewNetlinkRequest(syscall.RTM_SETLINK, syscall.NLM_F_ACK)
 
 	msg := nl.NewIfInfomsg(syscall.AF_UNSPEC)
-	msg.Type = syscall.RTM_SETLINK
-	msg.Flags = syscall.NLM_F_REQUEST
 	msg.Index = int32(base.Index)
-	msg.Change = syscall.IFLA_MTU
 	req.AddData(msg)
 
 	b := make([]byte, 4)
@@ -97,10 +94,7 @@ func LinkSetName(link Link, name string) error {
 	req := nl.NewNetlinkRequest(syscall.RTM_SETLINK, syscall.NLM_F_ACK)
 
 	msg := nl.NewIfInfomsg(syscall.AF_UNSPEC)
-	msg.Type = syscall.RTM_SETLINK
-	msg.Flags = syscall.NLM_F_REQUEST
 	msg.Index = int32(base.Index)
-	msg.Change = syscall.IFLA_IFNAME
 	req.AddData(msg)
 
 	data := nl.NewRtAttr(syscall.IFLA_IFNAME, []byte(name))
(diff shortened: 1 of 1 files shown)


mavenugo:
ping @vishvananda. This seems to be a basic issue. Can you please review it ?

LK4D4:
Btw travis not running tests. And `TestLinkAddDelVeth` failing in master.

mrjana:
@LK4D4 Is that test failing on master or my code?

vishvananda:
Travis cannot run tests because they require sudo. I'm all for removing the change values, as mentioned in #38. I'm not so sure about removing the other values, so I'm going to check. The veth test works fine for me.

vishvananda:
so as far as i can tell, setting NLM_F_REQUEST and RTM_SETLINK is correct. The change code was modified earlier in 0207b6b6339670b6f4e9ddd2c2e72ce8d567969e and it appears that was incorrect.

mrjana:
@vishvananda Yes, please do check. But the ifinfomsg hdr has the following C structure:
```
         struct ifinfomsg {
                  unsigned char  ifi_family; /* AF_UNSPEC */
                  unsigned short ifi_type;   /* Device type */
                  int            ifi_index;  /* Interface index */
                  unsigned int   ifi_flags;  /* Device flags  */
                  unsigned int   ifi_change; /* change mask */
              };
```
And the ifi_flags and ifi_type are meant to be interface specific type and flags. Not nlmsghdr type(like RTM_SETLINK) and nlmsghdr flags (like NLM_F_REQUEST)

mrjana:
@vishvananda Here's the code snippet which sets the nlmsghdr:
```
func NewNetlinkRequest(proto, flags int) *NetlinkRequest {
        return &NetlinkRequest{
                NlMsghdr: syscall.NlMsghdr{
                        Len:   uint32(syscall.SizeofNlMsghdr),
                        Type:  uint16(proto),
                        Flags: syscall.NLM_F_REQUEST | uint16(flags),
                        Seq:   atomic.AddUint32(&nextSeqNr, 1),
```
That is the only place NLM_F* and RTM* should be set. Not in ifinfomsg.

vishvananda:
Ah i see that now. I missed that it was setting it in the ifinfomsg. I stand corrected. Your code looks correct.

LK4D4:
If there is `unshare` on travis, then we can run tests under `unshare -Urn`. For me test failing with
```
--- FAIL: TestLinkAddDelVeth (0.05s)
        link_test.go:49: Created link has a Parent but it shouldn't
```
:/

mrjana:
@LK4D4 That test didn't fail for me when I ran the tests locally on my machine against my code

mavenugo:
works for me too.

LK4D4:
@mrjana Yeah, it fails for me both in master and here. I don't know why, because I remember it worked for me too. Maybe it's kernel magic again, I'm using 4.1.3

mrjana:
@LK4D4 Ah that might be it. Is it possible for you to debug and see where the failure is so that we can save some pain in the future? :-)

LK4D4:
@mrjana Yeah, the problem that newly created veth has ParentIndex of its peer, but test checking that it should be 0 if it wasn't specified.

PR #48, opened 2015-09-12
Htb Qdisc and Class

chantra:
Add support for HTB Qdisc and tc class. The class diff currently only add support for HTB class.

The Qdisc/Class have unittests. I also added unittests for some of the tc_linux structs.

### DIFF
--- a/class.go
+++ b/class.go
@@ -1,0 +1,110 @@
+package netlink
+
+import (
+	"fmt"
+)
+
+type Class interface {
+	Attrs() *ClassAttrs
+	Type() string
+}
+
+// Class represents a netlink class. A filter is associated with a link,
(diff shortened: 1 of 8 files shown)


chantra:
I have another diff in flight to add support for FW filter: https://github.com/chantra/netlink/tree/fw_filter 

AHappy to keep it for a separate PR if needed.

vishvananda:
fw filter code is looking good. A separate pr is the way to go. FYI, the htb code and especially the u32 code is incomplete. Actions should be a separate type with more than just Mirred available, but I did the minimum to support the functionality I needed. Feel free to add anything that you need.

chantra:
:+1:

PR #50, opened 2015-09-19
Add FW filter support

chantra:
Add support for FW filter.

### DIFF
--- a/filter.go
+++ b/filter.go
@@ -52,7 +52,7 @@ type FilterFwAttrs struct {
 	Rate      uint32
 	AvRate    uint32
 	PeakRate  uint32
-	Action    int32
+	Action    int
 	Overhead  uint16
 	LinkLayer int
 }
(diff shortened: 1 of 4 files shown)


chantra:
I will add documentation at a later stage but on this PR. Still, I created the PR so feedback can be given in the meantime.

filter.go
```diff
@@ -39,6 +41,89 @@ func (filter *U32) Type() string {
...
+	InDev     string
+	Mask      uint32
+	Index     uint32
+	Buffer    uint32
+	Mtu       uint32
+	Mpu       uint16
+	Rate      uint32
+	AvRate    uint32
+	PeakRate  uint32
+	Action    int32
```
vishvananda: generally i try to keep the signed ints at this layer as "int" instead of int32 but that is minor

filter_linux.go
```diff
@@ -1,10 +1,11 @@
 package netlink
 
 import (
+	"bytes"
+	"encoding/binary"
 	"fmt"
-	"syscall"
-
 	"github.com/vishvananda/netlink/nl"
+	"syscall"
```
vishvananda: this is out of order

vishvananda:
looks good overally. Some minor stuff inline

PR #51, opened 2015-09-25
Add basic support for TUN/TAP link type and creation.

xnox:


xnox:
All green now.

vishvananda:
Thanks for adding this! I'd like to keep netlink able to build without cgo, so you will need to move the Ifreq  structure definition into pure go and make a definition in go (with serialize/deserialize probably). There are tons of examples of this in nl/xfrm_linux.go. I'm also thinking that that should be part of the low level code but it doesn't belong in nl. Maybe ultimately it could move into an ioctl library.

It does muddy things a bit that it has to use ioctl which isn't exactly in the spirit of netlink, but i see that iproute2 added it because it makes it more usable so I'm ok with following their lead and adding it here.

xnox:
Sounds good, thanks for comments. I'm a golang beginner, but I have experts who can help me out.

PR #52, opened 2015-10-09
Add support for setting nexthop flags for routes

aakso:
There are few nexthop flags that are usable in some scenarios: onlink
and pervasive. This patch allows to set, clear and list them.

These flags make sence only when gateway is set. We will get an error from netlink if we set these flags without a gw. I'm unsure whether this is an ok behavior or should there be some validation before sending the request.


vishvananda:
Validation is a good idea if we know it is going to fail, but it can be added later.

PR #55, opened 2015-10-23
[netem] minimalist support for netem

chantra:
Support for tc_netem_qopt options, e.g: latency, limit, loss, gap, duplicate and jitter

### DIFF
--- a/qdisc.go
+++ b/qdisc.go
@@ -53,7 +53,11 @@ func HandleStr(handle uint32) string {
 	}
 }
 
-func Percentage(percentage float32) uint32 {
+func Percentage2u32(percentage float32) uint32 {
+	// FIXME this is most likely not the best way to convert from % to uint32
+	if percentage == 100 {
+		return math.MaxUint32
+	}
 	return uint32(math.MaxUint32 * (percentage / 100))
 }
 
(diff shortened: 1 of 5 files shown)


qdisc.go
```diff
@@ -120,6 +125,73 @@ func (qdisc *Htb) Type() string {
 	return "htb"
 }
 
+// Netem is a classless qdisc that rate limits based on tokens
+
+type NetemQdiscAttrs struct {
+	Latency   uint32 // in us
```
zfjagann: Maybe consider using [time.Duration](https://golang.org/pkg/time/#Duration)?

chantra: I am not sure it is worth it given that it will need to be casted back to a uint32 in netlink.

zfjagann: The reason being time.Duration is unit agnostic and more go idiomatic. See #56

qdisc_test.go
```diff
@@ -88,13 +88,6 @@ func TestHtbAddDel(t *testing.T) {
 		t.Fatal(err)
 	}
 
-	/*
-		cmd := exec.Command("tc", "qdisc")
-		out, err := cmd.CombinedOutput()
-		if err == nil {
-			fmt.Printf("%s\n", out)
-		}
-	*/
```
vishvananda: thanks for removing this debug code

qdisc_linux.go
```diff
@@ -166,6 +201,16 @@ func QdiscList(link Link) ([]Qdisc, error) {
 					if err := parseHtbData(qdisc, data); err != nil {
 						return nil, err
 					}
+				case "netem":
+					/*
+						data, err := nl.ParseRouteAttr(attr.Value)
+						if err != nil {
+							return nil, err
+						}
+					*/
```
vishvananda: this looks like debug code. Do you mind removing it?

vishvananda:
looks good. Small suggestion inline

chantra:
I `git push --force` with the comment removed.

PR #70, opened 2015-11-26
Add support for rules

krhubert:


vishvananda:
see comment on #72

krhubert:
updated, also when doing rule I thoguht that every netlink object (like link, router, rule, address) should have RtMsg header in struct. It can be useful for some users

PR #72, opened 2015-11-26
Add scope and flags support for netlink address

krhubert:


vishvananda:
Thanks for the work on this and the other package. I want to discuss the idea of having a MaskField. I'm not sure i like the interface that much. There are two fields in link.go that are optionally set. We have MTU which is only set it it is greater than 0, this approach makes sense if zero isn't a valid value for the option. We also have TxQLen where 0 is a valid value. For this field we default it to -1 and only set it if the value is positive. This works if we use a signed 64 bit value (i just noticed that TxQLen should probably be int64 instead of int for proper building on 32 bit platforms). The advantage of this method is that we don't need a separate field that we have to remember to set when we are setting optional values. The drawback is having to create the object with a specific method that defaults values to -1, and using slightly more space. I think i prefer this method to the FlagsMask approach that you use in this (and other) patches. Thoughts?

krhubert:
I think i prefer this method to the FlagsMask approach that you use in this (and other) patches. Thoughts? => you have rigth, I will patch the source,

I'm only affraid that in some cases FlagMask filed will be used. For example in Rule Table ID must be uint32, so type of Table filed must be int64 to be portable and to be able to set default value to e.g. -1

vishvananda:
Yes int64 is default on most platforms anyway, but we should explicitly set
it for portability.

krhubert:
Ok, addr pull request updated,doing rule :)

PR #73, opened 2015-11-26
New route fields, and route filtration 

marek-polewski:
Hi, 

This is route related part of changes that hkrauze and I were using in our projects. 

I think route.LinkIndex should be renamed to Oif for more like netlink msg feel. 
I kept here the LinkIndex name for backward compatibility.
RouteListFiltred with filtration struct in the second argument was added as a replacement for RouteList, but once again I left the previous call to not break anyone code.

Feel free to comment.

### DIFF
--- a/route_linux.go
+++ b/route_linux.go
@@ -29,8 +29,6 @@ func routeHandle(route *Route, req *nl.NetlinkRequest, msg *nl.RtMsg) error {
 		return fmt.Errorf("one of Dst.IP, Src, or Gw must not be nil")
 	}
 
-	msg.Scope = uint8(route.Scope)
-	msg.Flags = uint32(route.Flags)
 	family := -1
 	var rtAttrs []*nl.RtAttr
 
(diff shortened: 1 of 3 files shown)


route_linux.go
```diff
@@ -102,22 +157,28 @@ func routeHandle(route *Route, req *nl.NetlinkRequest, msg *nl.RtMsg) error {
...
+			Oif:      link.Attrs().Index,
+			FlagMask: RT_FILTER_OIF,
+		}
+	}
+	return RouteListFiltered(family, rf)
+}
+
+// RouteListFiltered gets a list of routes in the system filtered with specified rules.
+// All rules must be defined in RouteFilter struct
+func RouteListFiltered(family int, filter *RouteFilter) ([]Route, error) {
```
vishvananda: Rather than having a separate route filter object that duplicates all the fields, I think it makes more sense to have two parameters,  RouteListFiltered(family int, filter *Route, filterMask uint64)

vishvananda:
See my comment. I agree that Oif would make more sense, but yeah there is the problem of breaking existing code. I used LinkIndex since all the other stuff does. Maybe ILinkIndex and OLinkIndex? You could call the new parameter ILinkIndex to keep the same rough format if you want.

marek-polewski:
Hi,
I like the idea of using route and mask instead of additional structure.
Just applied those changes.

PR #74, opened 2015-11-30
Add support for creating gretap interfaces and setting and retrieving interface by alias.

mcastelino:
Added support for creating gretap interfaces. 
Added support for alias. This capability may be useful to set meta information to be associated with interfaces. The interface names are of limited length. The alias field allows for significantly more descriptive interface details.


vishvananda:
looks like one of the other merges conflicted with this. Could you rebase it?

PR #78, opened 2015-12-09
add bond interface support

marek-polewski:
Hi, 

This is our implementation of support for bond interfaces.

Earlier we were using FlagMask for distinguishing undefined fields, but I removed it for the negative values in fields (discussion on #72).

All empty fields are by default equal -1 (or nil). In my opinion, this  results in a faster understanding of which fields are undefined, instead of the rule: undefined := minValue(field) - 1  No need to know the minimal supported value.

Please tell me how you feel about that.


link.go
```diff
@@ -240,6 +241,289 @@ func (ipvlan *IPVlan) Type() string {
...
+	return &Bond{
+		LinkAttrs:       atr,
+		Mode:            -1,
+		ActiveSlave:     -1,
+		Miimon:          -1,
+		UpDelay:         -1,
+		DownDelay:       -1,
+		UseCarrier:      -1,
+		ArpInterval:     -1,
+		ArpIpTargets:    nil,
```
vishvananda: fyi the default value for a slice is nil

rade:
This PR is missing some code. `NewBond` is not defined anywhere. Cue [build failures](https://travis-ci.org/weaveworks/weave/builds/96004463#L299).

marek-polewski:
True, my mistake - sorry. #79 fix this.

PR #79, opened 2015-12-10
Fix build error

krhubert:
How this happend that CI dosen't catch build error?

### DIFF
--- a/link_linux.go
+++ b/link_linux.go
@@ -890,7 +890,7 @@ func parseVxlanData(link Link, data []syscall.NetlinkRouteAttr) {
 }
 
 func parseBondData(link Link, data []syscall.NetlinkRouteAttr) {
-	bond := NewLinkBond(LinkAttrs{})
+	bond := NewLinkBond(NewLinkAttrs())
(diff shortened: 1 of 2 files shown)


rade:
> How this happend that CI dosen't catch build error?

Looks like the CI is configured to simply run 'make'. And the default target in the Makefile... just runs `gofmt`.

andrewtchin:
+1 please merge this

vishvananda:
yeah unfortunately the tests require sudo so they don't really work on the ci server. I should at least modify it to do some kind of sanity build.

chantra:
Re ci, there is way to run as root when using vms. You may need to use the trusty image to get a kernel that is somehow newish.

LK4D4:
Also, it's possible with `unshare -Urn`, but you need newer util-linux.

vishvananda:
FYI i got the tests running in travis. I had to skip the ipvlan tests because the kernel is too old. Also there is a failure in the addr_linux tests. It looks like the older kernel doesn't set IFA_F_PERMANENT on the loopback device?

PR #84, opened 2016-01-04
Htbclass support for 2.6 kernel

chantra:
The 2.6 kernel series were making use of c/r tabs. In 3.10+, those are ignored for ethernet links type at least, so it is mostly a noop for 3.10+ and make the feature work on 2.6.
Also this is in line with iproute2 logic, I just did not get to code it in the first place while adding this feature.

More info in 
https://github.com/facebook/augmented-traffic-control/issues/225#issue-123893823

### DIFF
--- a/class.go
+++ b/class.go
@@ -1,10 +1,7 @@
 package netlink
 
 import (
-	"errors"
 	"fmt"
-
-	"github.com/vishvananda/netlink/nl"
 )
 
 type Class interface {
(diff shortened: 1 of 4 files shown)


class.go
```diff
@@ -41,21 +44,29 @@ func (q HtbClassAttrs) String() string {
...
 	Cbuffer uint32
 	Quantum uint32
 	Level   uint32
 	Prio    uint32
+	Rtab    [256]uint32
+	Ctab    [256]uint32
 }
 
-func NewHtbClass(attrs ClassAttrs, cattrs HtbClassAttrs) *HtbClass {
+func NewHtbClass(attrs ClassAttrs, cattrs HtbClassAttrs) (*HtbClass, error) {
```
zfjagann: I think it would be best to change the `HtbClassAttrs` struct to make Rate and Ceil both `nl.TcRateSpec` that way the signature of this function can remain unchanged.

chantra: So the error being returned is mainly due to `CalcRtable`. I had thought about it, but from a usage standpoint, I think it is better to provide the rates as attribute rather than the opaque `nl.TcRateSpec`.

That being said, I would be rather kin on aligning this with the rest of the code base for the sake of uniformity though. Will check how this is handle in other places.

class.go
```diff
@@ -41,21 +44,29 @@ func (q HtbClassAttrs) String() string {
 // Htb class
 type HtbClass struct {
 	ClassAttrs
-	Rate    uint64
-	Ceil    uint64
+	Rate    nl.TcRateSpec
+	Ceil    nl.TcRateSpec
```
vishvananda: in general i try to avoid nl specific types outside of the netlink library. It should be possible to use netlink without importing nl. If you must expose it this way, please creete a new type in class.go that mirrors the lower level nl.TcRateSpec, although i would prefer it if we could keep the types simple in netlink

chantra:
@vishvananda I moved the logic into class_linux.go instead. In the end, the interface has not change from before this PR.

PR #88, opened 2016-01-22
Do not send nl request data in XfrmStateList

ibuildthecloud:
I kept seeing `netlink: 16 bytes leftover after parsing attributes` messages when calling `XfrmStateList()` and from my (very naive) understanding of the kernel code it didn't seem like `XFRM_MSG_GETSA` w/ `NLM_F_DUMP` excepts any attributes.  The passed in `family` argument seems to be used by the go code and not the kernel itself.


vishvananda:
Thanks, I never noticed this error before.

Issue #89, opened 2016-01-27
How to subscribe to netlink messages about changes of interface address?


udhos:
How to subscribe to netlink messages about changes of interface address?

Is it currently possible? How? Please advise.

PR #90, opened 2016-01-28
AddrSubscribe(): receive updates for changes in link addresses.

udhos:
Fix for https://github.com/vishvananda/netlink/issues/89


addr_linux.go
```diff
@@ -140,3 +141,81 @@ func AddrList(link Link, family int) ([]Addr, error) {
...
+				return
+			}
+			for _, m := range msgs {
+				msgType := m.Header.Type
+				if msgType != syscall.RTM_NEWADDR && msgType != syscall.RTM_DELADDR {
+					log.Printf("netlink.AddrSubscribe: bad message type: %d", msgType)
+					continue
+				}
+
+				msg := nl.DeserializeIfAddrmsg(m.Data)
```
dcbw: This code seems almost exactly the same as AddrList(), it should probably get pulled out into its own function which is then called from both AddrList and AddrSubscribe.

vishvananda:
I agree with the other comment. Best to minimize duplicate code

udhos:
Moved common code from AddrList,AddrSubscribe into new helper func parseAddr.

PR #91, opened 2016-01-30
RFC: force AddList to get rid of unexpected information

liusdu:
#`AddList` returned all addresses when the protocol family is not supported. For example, With a kernel with IPV6 disabled. if I call `netlink.AddrList(link, netlink.FAMILY_V6)`., it will return all ips of this interface.  This behavior may confuse user. After this patch it will return []  instead.
 
When I ran docker daemon on a machine with ipv6 disabled, I encounterd the following error. And this patch made it workable.
```
FATA[0000] Error starting daemon: Error initializing network controller: Error creating default "bridge" network: Failed to setup IP tables, cannot acquire Interface address: Interface docker0 has no IPv4 addresses
```

### DIFF
--- a/addr_linux.go
+++ b/addr_linux.go
@@ -101,6 +101,10 @@ func AddrList(link Link, family int) ([]Addr, error) {
 			continue
 		}
 
+		if family != FAMILY_ALL && msg.Family != uint8(family) {
+			continue
+		}
+
 		attrs, err := nl.ParseRouteAttr(m[msg.Len():])
 		if err != nil {
 			return nil, err
--- a/addr_test.go
+++ b/addr_test.go
@@ -77,6 +77,25 @@ func TestAddr(t *testing.T) {
 			t.Fatalf("Address scope not set properly, got=%d, expected=%d", addrs[0].Scope, tt.expected.Scope)
 		}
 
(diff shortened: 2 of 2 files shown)


addr_linux.go
```diff
@@ -90,6 +90,10 @@ func AddrList(link Link, family int) ([]Addr, error) {
 			continue
 		}
 
+		if msg.Family != uint8(family) {
```
vishvananda: I think this will cause passing in FAMILY_ALL to fail. You probably want: 'if family != FAMILY_ALL && msg.Family != uint8(family) {'

vishvananda:
Thanks for the change. One fix suggested inline.

addr_linux.go
liusdu: This makes sense.  Thanks for you advice.

liusdu:
I run the UT. addaddr test passed.
```
go test -v github.com/vishvananda/netlink/./
=== RUN   TestAddrAddDel
--- PASS: TestAddrAddDel (0.00s)
=== RUN   TestClassAddDel
--- FAIL: TestClassAddDel (0.00s)
	class_test.go:11: operation not supported
=== RUN   TestFilterAddDel
--- FAIL: TestFilterAddDel (0.00s)
	filter_test.go:14: operation not supported
```

PR #92, opened 2016-02-15
Add support for setting route flags back

aakso:
Removed in dfdad47 without any comments


vishvananda:
seems reasonable. Was this causing a noticeable problem? If so it would be nice to add it to the commit message or the issue in case people are searching for it.

aakso:
I added an issue #93 for this.

aakso:
I also amended the commit message

PR #94, opened 2016-02-29
Add support for BPF filters and actions

drzaeus77:
This pull request adds support for BPF filters and actions, and refactors slightly the U32 parsing to support future extensions. Each commit in the series should be independently testable and reviewable.


vishvananda:
This is awesome, thanks for this. I think we could even remove RedirIndex because I suspect one of my other projects is the only user of that hack :). Thanks for putting in the backwards compatibility though.

PR #95, opened 2016-03-04
netlink with gcc-go and powerpc

michelmno:
part of changes to build docker on openSUSE https://build.opensuse.org/request/show/364697


michelmno:
my patch was most probably not good enough,  would it be sufficient to add more +build lines ?

PR #96, opened 2016-03-04
netlink with gcc-go and powerpc

michelmno:
version 2


michelmno:
not the correct way to solve, the issue is in fact tracked and solved by https://github.com/golang/go/issues/11707

Issue #99, opened 2016-03-18
go get error


udhos:
I'm getting the go get error below:

```
lab@ubu10:~/devel$ go get github.com/vishvananda/netlink
# github.com/vishvananda/netlink
nexthop/src/github.com/vishvananda/netlink/bpf_linux.go: In function ‘load_simple_bpf’:
nexthop/src/github.com/vishvananda/netlink/bpf_linux.go:33:12: warning: cast from pointer to integer of different size [-Wpointer-to-int-cast]
   .insns = (__u64)&insns,
            ^
nexthop/src/github.com/vishvananda/netlink/bpf_linux.go:34:14: warning: cast from pointer to integer of different size [-Wpointer-to-int-cast]
   .license = (__u64)&license,
              ^
lab@ubu10:~/devel$
lab@ubu10:~/devel$ go version
go version go1.6 linux/386
lab@ubu10:~/devel$
```

drzaeus77:
I will look into that, the fix shouldn't be too difficult. And is it a
warning like it said or does it fail to run afterwards.

udhos:
> And is it a warning like it said or does it fail to run afterwards.

It seems to cause actual issues:

```
root@ubu10:/home/lab/devel# go test github.com/vishvananda/netlink
# github.com/vishvananda/netlink
nexthop/src/github.com/vishvananda/netlink/bpf_linux.go: In function ‘load_simple_bpf’:
nexthop/src/github.com/vishvananda/netlink/bpf_linux.go:33:12: warning: cast from pointer to integer of different size [-Wpointer-to-int-cast]
   .insns = (__u64)&insns,
            ^
nexthop/src/github.com/vishvananda/netlink/bpf_linux.go:34:14: warning: cast from pointer to integer of different size [-Wpointer-to-int-cast]
   .license = (__u64)&license,
              ^
--- FAIL: TestLinkAddDelGretap (0.08s)
        link_test.go:123: Link not removed properly
--- FAIL: TestXfrmPolicyAddDel (0.62s)
panic: runtime error: slice bounds out of range [recovered]
        panic: runtime error: slice bounds out of range

goroutine 52 [running]:
panic(0x81fbac0, 0x1850a018)
        /usr/local/go/src/runtime/panic.go:464 +0x326
testing.tRunner.func1(0x1856be60)
        /usr/local/go/src/testing/testing.go:467 +0x14f
panic(0x81fbac0, 0x1850a018)
[...]

PR #100, opened 2016-03-18
Fix 32bit warnings in bpf_linux.go

drzaeus77:
The underlying struct type is still u64 regardless of architecture, but
the local cast should be to uintptr_t to squelch the warning.


udhos:
I confirm this change did fix the compiler warning I reported in https://github.com/vishvananda/netlink/issues/99

PR #101, opened 2016-03-21
Fix return of address family in parseAddr helper.

udhos:
I broke return of family in commit https://github.com/vishvananda/netlink/commit/9462941794a2fe8f0a0ed34ba8afcc3cf627cb77

PR #102, opened 2016-04-05
addr_linux: copy family from msg in parseAddr

unclejack:
A recent refactoring of the code has broken this. This PR makes parseAddr not return family=-1 all the time. This change fixes AddrList().

I've investigated this issue together with @mapuri.


unclejack:
I've missed #101. I'm closing this in favour of #101.

PR #104, opened 2016-04-15
Provide netlink handle

aboch:
This is a proposal for Netlink library to provide a netlink handle.
I would like to gather some early feedback on the design.
I'd prefer to get an approval before working further on the change.
Thanks.

It allows a client which performs a series of netlink operations to skip the overhead of socket creation/deletion with each API call. 
It also allows the client to specify on which network namespace the socket(s) should be opened.
This way the go client does not need to manage the compexity of setting to a netns for performing a netlink call because the socket would already be bound to the netns and provides a good abstraction for support netns aware netlink programming.

- Netlink socket(s) are created at handle request and released at handle deletion.
  All netlink requests on the handle will re-use same socket(s).
- Network namespace can be specified during handle creation. 
  Socket will be opened on the specified network namespace.
- The change is backward-compatible: Current package API behavior is unchanged.

Example of flow:
```
nlh, _ := netlink.NewHandle()
...
iface, _ := nlh.LinkByName(name)
nlh.LinkSetDown(iface)
nlh.LinkSetName(iface, newName)
nlh.LinkSetHardwareAddr(iface, mac)
nlh.AddrAdd(iface, ip)
nlh.LinkSetUp(iface)
...
nlh.Delete()
```

### DIFF
--- a/addr_linux.go
+++ b/addr_linux.go
@@ -16,24 +16,45 @@ const IFA_FLAGS = 0x8
 // AddrAdd will add an IP address to a link device.
 // Equivalent to: `ip addr add $addr dev $link`
 func AddrAdd(link Link, addr *Addr) error {
+	h, err := NewHandle()
+	if err != nil {
+		return err
+	}
+	defer h.Delete()
+	return h.AddrAdd(link, addr)
+}
 
-	req := nl.NewNetlinkRequest(syscall.RTM_NEWADDR, syscall.NLM_F_CREATE|syscall.NLM_F_EXCL|syscall.NLM_F_ACK)
-	return addrHandle(link, addr, req)
+// AddrAdd will add an IP address to a link device.
+// Equivalent to: `ip addr add $addr dev $link`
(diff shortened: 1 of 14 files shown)


vishvananda:
This seems like a reasonable approach. I like that you only have to lock the os thread once while you are getting the socket and for future requests you can use the socket without having to re-enter the namespace.

aboch:
Thanks for the feedback @vishvananda. Will polish the code, API and add a handle_test.go in the coming days.

mcastelino:
@aboch @vishvananda, I am  looking forward to this. I have been benchmarking how the network interface creation scales specially when multiple netlink calls are made in parallel, and found that I run into significant latency issues when making calls in parallel. I can easily measure how much this change will help address these issues. I have a simple test setup  to measure the latency impact `https://github.com/01org/ciao/blob/master/networking/libsnnet/tests/parallel/parallel_test.go` which I can

PR #108, opened 2016-04-28
Proposal to add multi queue support to tuntap

mcastelino:
This is a draft proposal to add multi queue support to tuntap without breaking legacy users of tuntap (i.e. not leaking fds) based on [https://www.kernel.org/doc/Documentation/networking/tuntap.txt]

To be able to leverage multiqueue the caller needs to have the fds available. Adding the Queue attribute to Tuntap and passing the fds back only when the Queue attribute was explicitly set seems to be the least intrusive way to add this capability.

Please do not merge this. The multi-queue tap has not been tested.

### DIFF
--- a/link_linux.go
+++ b/link_linux.go
@@ -433,6 +433,12 @@ func addBondAttrs(bond *Bond, linkInfo *nl.RtAttr) {
 	}
 }
 
+func cleanupFds(fds []*os.File) {
+	for _, f := range fds {
+		f.Close()
+	}
+}
+
 // LinkAdd adds a new link device. The type and features of the device
 // are taken fromt the parameters in the link object.
 // Equivalent to: `ip link add $link`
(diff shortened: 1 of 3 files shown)


link_linux.go
```diff
@@ -455,23 +461,43 @@ func LinkAdd(link Link) error {
...
+		for i := 0; i < queues; i++ {
+			file, err := os.OpenFile("/dev/net/tun", os.O_RDWR, 0)
+			if err != nil {
+				cleanupFds(tuntap.Fds[:i])
+				return err
+			}
+			tuntap.Fds = append(tuntap.Fds, file)
+			var req ifReq
+			//req.Flags |= syscall.IFF_ONE_QUEUE
+			//req.Flags |= syscall.IFF_TUN_EXCL
```
vishvananda: is there a reason these are commented out instead of deleted?

vishvananda:
This seems like a reasonable approach. Let me know how the testing goes.

link_linux.go
mcastelino: No. I plan to remove them in the final update. The original logic was setting them. This was just for reference. libvirt does not set them either.

vishvananda:
any progress on the testing?

mcastelino:
Will be done testing in the next couple of days. Will let you know as soon as possible.

vishvananda:
No updates in almost a year. Closing for now. Please rebase and reopen if you have had time to test it.

PR #110, opened 2016-04-29
Add Mark field to xrfm state and policy

aboch:
Tested on Ubuntu Linux 3.13, 3.19, 4.4

### DIFF
--- a/nl/xfrm_state_linux_test.go
+++ b/nl/xfrm_state_linux_test.go
@@ -205,3 +205,29 @@ func TestXfrmEncapTmplDeserializeSerialize(t *testing.T) {
 	msg := DeserializeXfrmEncapTmpl(orig)
 	testDeserializeSerialize(t, orig, safemsg, msg)
 }
+
+func (msg *XfrmMark) write(b []byte) {
+	native := NativeEndian()
+	native.PutUint32(b[0:4], msg.Value)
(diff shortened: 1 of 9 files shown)


nl/xfrm_state_linux_test.go
```diff
@@ -205,3 +205,12 @@ func TestXfrmEncapTmplDeserializeSerialize(t *testing.T) {
 	msg := DeserializeXfrmEncapTmpl(orig)
 	testDeserializeSerialize(t, orig, safemsg, msg)
 }
+
+func TestSerializeDeserializeXfrmMark(t *testing.T) {
```
vishvananda: Any reason you didn't follow the format of all the other DeserializeSerialize tests?

vishvananda:
minor comment inline. If there isn't a specific reason to have a unique test for mark it is probably best to stick with the standard format for serialization tests. Otherwise it all looks good

nl/xfrm_state_linux_test.go
aboch: No valid reason: I did not pay attention to the existing test code. Thanks to your comment, I just realized there is a template to automate the serialization/deserialization check of the structure filled with random data. 

I will push now the proper test code for the new structure. Thanks.

PR #113, opened 2016-05-06
Allow to program L4 fields in policy selector

aboch:
Tested on Ubuntu Linux 3.13, 3.19, 4.4

### DIFF
--- a/xfrm_policy.go
+++ b/xfrm_policy.go
@@ -52,6 +52,9 @@ type XfrmPolicyTmpl struct {
 type XfrmPolicy struct {
 	Dst      *net.IPNet
 	Src      *net.IPNet
+	Proto    Proto
+	DstPort  int
+	SrcPort  int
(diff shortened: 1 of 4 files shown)


xfrm_policy.go
```diff
@@ -52,6 +52,9 @@ type XfrmPolicyTmpl struct {
 type XfrmPolicy struct {
 	Dst      *net.IPNet
 	Src      *net.IPNet
+	Proto    Proto
+	DPort    uint16
```
vishvananda: All of the other port constructs use int instead of u1nt16 for the higher level representation, so these should be changed to match. I'm also wondering if they should  be called DstPort and SrcPort to match State.

aboch: Makes sense, I will follow the same convention for the parameters's type. 
Agree on the names, no need to be too concise.

PR #117, opened 2016-05-09
Use package empty handle for pkg methods

aboch:
#104 introduced the Handle structure with all the netlink pkg methods over it. The existing package methods were reworked to call the counterpart Handle methods for code re-use.

For this, in each pkg method a new handle is created, but this causes a couple of sockets (ROUTE and XFRM) be created and then closed, when it is not really needed...

Original intention was for the above change to be much less pervasive, in fact nl.Execute() was already modified to expect an handle with no allocated socket, causing the socket to be allocated on demand, therefore working as if the handle is not there, as it was doing before #104 

Unfortunately I oversaw this. With this change I am fixing the damage (runtime wise pkg method would behave as before #104) and makes the code more readable IMO.


vishvananda:
Thanks much cleaner

PR #118, opened 2016-05-11
Parse link statistics

aboch:
So that `ip -s link show <dev>` info is accessible

Tested on ubuntu 3.13, 3.19, and 4.4


link.go
```diff
@@ -40,6 +41,19 @@ func NewLinkAttrs() LinkAttrs {
 	}
 }
 
+type LinkStatistics struct {
+	rx_packets uint32
```
LK4D4: I wonder what the point of private fields here? Or is it a bug?

aboch: A bug. Thanks. Will fix

PR #119, opened 2016-05-11
XFRM Get/Delete state/policy should share same code

aboch:
- Currently they are not and GET methods are passing
  the wrong structure. Also they are setting the incorrect
  XFRM_F_DUMP flag. Because of this, current get methods
  do not return expected error when query target is not found.


vishvananda:
looks like there is a minor conflict with one of the other merges. Can you rebase?

aboch:
Thanks. Rebased and fixed the conflict.

PR #121, opened 2016-05-12
Properly pass IP address flags

### LINKED ISSUE #105: IFA_F_NODAD is not programmed on 3.13 kernel
https://github.com/vishvananda/netlink/pull/72 added the support for flags in netlink address. 
We use it to disable the ipv6 duplicate address detection. 
But netlink.AddrAdd() does not program the passed flag on 3.13 kernels.
Works fine on 3.19 and later.

Related to https://github.com/docker/docker/issues/22157

aboch:
Older kernel expect the flags in the main netlink message, not as attribute.
Once flag number increased beyond 8 in newer kernels, they moved them into the attributes.

Tested on ubuntu 3.13

Fixes #105

### DIFF
--- a/addr_linux.go
+++ b/addr_linux.go
@@ -69,10 +69,14 @@ func (h *Handle) addrHandle(link Link, addr *Addr, req *nl.NetlinkRequest) error
 	req.AddData(addressData)
 
 	if addr.Flags != 0 {
-		b := make([]byte, 4)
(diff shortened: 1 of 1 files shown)


vishvananda:
I was thinking we should always pass the lower bits in flags, but i guess iproute2 does the same, so lgtm.

aboch:
Right, I saw the way iproute2 handles this and replicated here.

PR #129, opened 2016-05-26
add generic actions and simplify action type

vishvananda:


vishvananda:
@drzaeus77 Can you take a look at these changes and make sure it doesn't mess up anything regarding your bpf handling? In general I'm trying to avoid the types nl library to the netlink public api. Ideally a user of netlink never has to import nl. Also I tend to prefer straight int values in the public api. I therefore moved some of the definitions for actions. There is still one reference to nl in the fw definition, but we can handle that later.

drzaeus77:
Looks good!

Tested and they work as before.

PR #139, opened 2016-06-16
Modprobe required modules in newHandle

aboch:
- but do not fail creation if some are missing

Some Linux distributions do not have autoload of modules configured and some new minimal ones do not even contain `xfrm_algo` and `xfrm_user` modules required to open an XFRM netlink socket.

In order to address the above and be able to initialize the Handle, modprobe for the modules, and attempt xfrm socket creation only if the modprobe was successful.

An example, this is from an Antergos 4.4.12-1-lts: 
```
modprobe -va xfrm_user xfrm_algo                             
modprobe: WARNING: Module xfrm_user not found in directory /lib/modules/4.4.12-1-lts
modprobe: WARNING: Module xfrm_algo not found in directory /lib/modules/4.4.12-1-lts
```

### DIFF
--- a/handle_linux.go
+++ b/handle_linux.go
@@ -1,6 +1,9 @@
 package netlink
 
 import (
+	"fmt"
+	"os/exec"
+	"strings"
 	"sync/atomic"
 	"syscall"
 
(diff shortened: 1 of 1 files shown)


aboch:
Superseded by #143

Issue #144, opened 2016-06-22
GBP field on netlink.Vxlan{}, if set to true, throws an error "numerical result out of range"


evanfarrar:
We have observed this in a live environment when experimenting with adding GBP to flannel. And by adding this to the netlink.Vxlan{} in unit tests:
```
diff --git a/link_test.go b/link_test.go
index ca11bb2..b654e4a 100644
--- a/link_test.go
+++ b/link_test.go
@@ -588,6 +588,7 @@ func TestLinkAddDelVxlan(t *testing.T) {
 		Learning:     true,
 		L2miss:       true,
 		L3miss:       true,
+		GBP: true,
 	}

 	testLinkAddDel(t, &vxlan)
```
We see in the test results:
```
--- FAIL: TestLinkAddDelVxlan (0.45s)
	link_test.go:27: numerical result out of range
```
The flannel was running on kernel version 4.4.0
The unit tests were on kernel version 4.4.13

PR #145, opened 2016-06-23
Fix VXLAN GBP support: Changes GBP field to set a value of NULL when set.

evanfarrar:
This is a proposed fix for #144 

We examined the values passed to netlink by the iproute2 implementation of GBP and were able to see that NULL was the correct value to pass for  when you want to enable group based policy on a device.

See: http://fossies.org/linux/iproute2/ip/iplink_vxlan.c

We also added this field to one of the tests to replicate the error message we were seeing and test the correct unmarshaling from null to "true".

Signed-off-by: Jay Dunkelberger <ldunkelberger@pivotal.io>

### DIFF
--- a/link_linux.go
+++ b/link_linux.go
@@ -425,7 +425,7 @@ func addVxlanAttrs(vxlan *Vxlan, linkInfo *nl.RtAttr) {
 		nl.NewRtAttrChild(data, nl.IFLA_VXLAN_UDP_CSUM, boolAttr(vxlan.UDPCSum))
 	}
 	if vxlan.GBP {
-		nl.NewRtAttrChild(data, nl.IFLA_VXLAN_GBP, boolAttr(vxlan.GBP))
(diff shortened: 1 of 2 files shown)


evanfarrar:
Our test would only pass on 4.4 kernels and above, we will break that into a separate test case and guard it from running on travis.

evanfarrar:
@tgraf can you review this?

rosenhouse:
@vishvananda VXLAN [GBP](https://tools.ietf.org/html/draft-smith-vxlan-group-policy-02) is currently broken in netlink, this PR fixes it.  Would you mind taking a look?

vishvananda:
yeah it looks fine to me, i see that @evanfarrar asked @tgraf to take a look so I was waiting for him to comment before I merge it.

tgraf:
I missed the ping. Looking into this now.

tgraf:
LGTM

rosenhouse:
Thanks!

PR #149, opened 2016-07-10
Add proxy support to the neighbor functions

CtrlZvi:
- Don't require a MAC address for a neighbor proxy
- Include proxies in the list of neighbors

Signed-off-by: Zvi "CtrlZvi" Effron <viz+GitHub@flippedperspective.com>

### DIFF
--- a/neigh_test.go
+++ b/neigh_test.go
@@ -10,6 +10,11 @@ type arpEntry struct {
 	mac net.HardwareAddr
 }
 
+type proxyEntry struct {
+	ip  net.IP
+	dev int
+}
+
 func parseMAC(s string) net.HardwareAddr {
 	m, err := net.ParseMAC(s)
 	if err != nil {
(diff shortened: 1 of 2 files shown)


neigh_test.go
```diff
@@ -102,3 +126,78 @@ func TestNeighAddDel(t *testing.T) {
...
+			Flags:     NTF_PROXY,
+			IP:        entry.ip,
+		})
+
+		if err != nil {
+			t.Errorf("Failed to NeighDel: %v", err)
+		}
+	}
+
+	// TODO: seems not working because of cache
```
vishvananda: probably should fix this todo before merging

vishvananda:
Thanks for adding this. One other thought. The iproute2 package allows you to list proxies separately. I'm wondering if we really want to include proxies by default or if it should be two separate commands? NeighList and NeighProxyList. Thoughts?

CtrlZvi:
I actually was thinking the same thing after I submitted the PR. Plus, no chance of breaking someone's current user cases. I'll make the changes and update the PR.

PR #154, opened 2016-07-21
Tuntap: support for non-default flags

lukasredynk:
I've added additional field in Tuntap struct which allow creating tuntap device with non-default flag.

### DIFF
--- a/link.go
+++ b/link.go
@@ -171,11 +171,13 @@ func (macvtap Macvtap) Type() string {
 }
 
 type TuntapMode uint16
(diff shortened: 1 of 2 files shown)


vishvananda:
Code looks correct but I think we need some way for the defaults to be set automatically if a user doesn't fill out flags. Would a user ever purposefully set no flags = 0 ? If not, we can just use _DEFAULTS if flags == 0

lukasredynk:
Ok, fixed.

jellonek:
Bump up.

PR #155, opened 2016-07-21
Fix bug in xfrmStateGetOrDelete

aboch:
- It fails if source address attribute is passed in 4 byte notation 

It fails with `numerical result out of range`

### DIFF
--- a/xfrm_state_linux.go
+++ b/xfrm_state_linux.go
@@ -213,7 +213,7 @@ func (h *Handle) xfrmStateGetOrDelete(state *XfrmState, nlProto int) (*XfrmState
 		req.AddData(out)
 	}
 	if state.Src != nil {
-		out := nl.NewRtAttr(nl.XFRMA_SRCADDR, state.Src)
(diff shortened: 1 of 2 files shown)


xfrm_state_linux.go
```diff
@@ -213,7 +213,9 @@ func (h *Handle) xfrmStateGetOrDelete(state *XfrmState, nlProto int) (*XfrmState
 		req.AddData(out)
 	}
 	if state.Src != nil {
-		out := nl.NewRtAttr(nl.XFRMA_SRCADDR, state.Src)
+		sAddr := nl.XfrmAddress{}
+		sAddr.FromIP(state.Src)
+		out := nl.NewRtAttr(nl.XFRMA_SRCADDR, sAddr.ToIP())
```
aboch: Wanted to reuse the `nl.XfrmAddress` structure. But if you are ok with it, I can simply just change this line to:

```
out := nl.NewRtAttr(nl.XFRMA_SRCADDR, state.Src.To16())
```

aboch:
Tests pass on my machine. Not sure why that test fails on CI.

PR #156, opened 2016-08-05
Add support for XDP link option

drzaeus77:
This series cleans up a few minor definitions in the IFLA definitions, and then adds support for IFLA_XDP. The use case and explanation for such option can be found at:
http://www.spinics.net/lists/netdev/msg387119.html

### DIFF
--- a/bpf_linux.go
+++ b/bpf_linux.go
@@ -8,11 +8,11 @@ package netlink
 #include <stdint.h>
 #include <unistd.h>
 
-static int load_simple_bpf(int prog_type) {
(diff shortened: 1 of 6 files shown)


vishvananda:
looks good. Can you do a quick rebase and squash it into one commit?

drzaeus77:
squashed+rebased

vishvananda:
Thanks. Am I correct in understanding that this is functionality that has not yet been merged into the kernel? I'd like it to be in the mainline before we merge this.

drzaeus77:
This is available in 4.8.

PR #157, opened 2016-08-09
[WIP] Add functions to turn on/off promiscuous mode for link

dshulyak:
Analogs for:
  ip l set dev eth0 promisc off
  ip l set dev eth0 promisc on

### DIFF
--- a/link.go
+++ b/link.go
@@ -31,6 +31,7 @@ type LinkAttrs struct {
 	Namespace    interface{} // nil | NsPid | NsFd
 	Alias        string
 	Statistics   *LinkStatistics
+	Promisc      int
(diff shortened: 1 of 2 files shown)


luxas:
Hi @vishvananda!
I would be glad to see this feature merged into your awesome project!
Right now we're tracking down the kubelet PATH dependencies, and in this case we want to use netlink to toggle promiscuous mode instead of exec'ing `ip` in PATH.

Looking forward to getting this merged. It would be nice to get in this week, so we can have it in v1.4.
Cheers!

luxas:
Great! That was faster than I'd imagined :)

PR #159, opened 2016-08-16
Multipath route is not installed due to gwData is duplicated in block.

kishiguro:
Multipath route with Gw is not properly installed due to gwData is newly allocated in block.
With this fix, multipath route with gateway address is properly installed into the kernel.


ishidawataru:
My bad.
Tested the fix, and properly working.

@vishvananda If you need a unit test for this fix to merge, I can open a PR. (https://github.com/ishidawataru/netlink/commit/1857146852c682f7b41f1465a5c98cfcf514b3d5)

PR #164, opened 2016-09-04
Supersides int with uint32 on MTU

### LINKED ISSUE #163: Use uint32 on MTU
Hi there,

As per https://en.wikipedia.org/wiki/Maximum_transmission_unit

Even when we under IPv6, the maximum MTU will not over 4294967296 (2^32-1), so I wonder if it make sense to use `uint32` instead of `int` on MTU?

HuKeping:
Closes #163 

Signed-off-by: Hu Keping <hukeping@huawei.com>

### DIFF
--- a/link.go
+++ b/link.go
@@ -21,7 +21,7 @@ type (
 // LinkAttrs represents data shared by most link types
 type LinkAttrs struct {
 	Index        int
-	MTU          int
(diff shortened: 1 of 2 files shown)


vishvananda:
In addition to keeping the interface simplified, this would break existing users of the library.

PR #169, opened 2016-11-02
Fix bug in ruleHandle: allocate different buffers for each rtattr

### LINKED ISSUE #150: rule priority defaults to table id value when used with table id higher than 255
I'm trying to set a rule which look like this:
`sudo ip rule add to 200.200.200.200 lookup 500 priority 3000`

produces the following ip rule output:
`3000:  from all to 200.200.200.200 lookup 500`

If the same operation applied with `netlink`:

> _, dstIpNet, _ := net.ParseCIDR("200.200.200.200/32")
> _, srcIpNet, _ := net.ParseCIDR("0.0.0.0/0")
> 
> rule := netlink.NewRule()
> rule.Src = dstIpNet
> rule.Dst = srcIpNet
> rule.Priority = 3000
> rule.Table = 500
> netlink.RuleAdd(rule)`

It produces the following `ip rule` output:
`500:   from 200.200.200.200 lookup 500`

I'm not sure what is the official support status for table ids higher than 255 in the recent kernels? It works fine with CLI and It looks like IDs > 255 work fine in `netlink` if we don't specify the `priority` field. 

I tested it on `3.13.0-76-generic #120-Ubuntu`

Any ideas?

tkojima:
This PR resolves #150.

Here is an example code:

```go
package main

import (
	"fmt"

)

func main() {
	rule := netlink.NewRule()

	rule.IifName = "enp4s0"
	rule.Priority = 6789
	rule.Table = 12345

	err := netlink.RuleAdd(rule)

	if err != nil {
		fmt.Println(err)
		return
	}
}
```
 
### before applying the patch

```
% go build -o example
% ./example
0:      from all lookup local 
12345:  from all iif enp4s0 lookup 12345 
32766:  from all lookup main 
32767:  from all lookup default 
```

priority equals to table id.

### after applying the patch

```
% go build -o example
% ./example
0:      from all lookup local 
6789:   from all iif enp4s0 lookup 12345 
32766:  from all lookup main 
32767:  from all lookup default 
```

It seems to work well.

### DIFF
--- a/rule_linux.go
+++ b/rule_linux.go
@@ -82,41 +82,46 @@ func ruleHandle(rule *Rule, req *nl.NetlinkRequest) error {
 		req.AddData(rtAttrs[i])
 	}
 
-	var (
-		b      = make([]byte, 4)
-		native = nl.NativeEndian()
-	)
+	native := nl.NativeEndian()
 
 	if rule.Priority >= 0 {
+		b := make([]byte, 4)
 		native.PutUint32(b, uint32(rule.Priority))
 		req.AddData(nl.NewRtAttr(nl.FRA_PRIORITY, b))
 	}
 	if rule.Mark >= 0 {
+		b := make([]byte, 4)
 		native.PutUint32(b, uint32(rule.Mark))
 		req.AddData(nl.NewRtAttr(nl.FRA_FWMARK, b))
 	}
 	if rule.Mask >= 0 {
+		b := make([]byte, 4)
 		native.PutUint32(b, uint32(rule.Mask))
 		req.AddData(nl.NewRtAttr(nl.FRA_FWMASK, b))
 	}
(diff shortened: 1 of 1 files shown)


rule_linux.go
```diff
@@ -82,7 +82,7 @@ func ruleHandle(rule *Rule, req *nl.NetlinkRequest) error {
 		req.AddData(rtAttrs[i])
 	}
 
-	native = nl.NativeEndian()
+	native := nl.NativeEndian()
```
aboch: This change is part of the previous commit. Same PR.
You do not need an extra commit, just amend the previous one.

tkojima: OK, I got it.

tkojima:
Thanks for merging!

Issue #131, opened 2016-06-07
netlink.LinkSetName is hung for ever, never coming out....


manoj0077:
netlink.LinkSetName(iface, i.SrcName()) , This is never returning and waiting indefinitely at 
nr, _, err := syscall.Recvfrom(s.fd, rb, 0) ....Above situation is not happening every time...but it is happening in some situation in production machine...

is there a way we can timeout here ? 

kernel version:- 3.19.0-58-generic #64~14.04.1-Ubuntu

vishvananda:
I haven't seen yet. We could timeout on recvfrom the kernel but it shouldn't fail. Not sure what might be causing that.

vishvananda:
That said, I am seeing the tests ocassionally hang on travis: https://travis-ci.org/vishvananda/netlink/builds/137837585

I do see the following:

https://lkml.org/lkml/2015/9/25/3
There are quite a few comments in that thread about patches that may need to be applied. It is possible that travis has an affected kernel as well. I would suggest updating your kernel or reporting the issue to ubuntu if you need to keep using that 3.19

manoj0077:
Hi, Thanks for looking into this....unfortunately we updated kernel to 4.2.0-36-generic #42~14.04.1-Ubuntu and seeing this issue now in LinkSetNsFd...where it is hanging for ever.....any pointers on higher kernel patch that went to resolve this....?

manoj0077:
If possible can you also give some pointers on how the time out is implemented fro recvfrom ?

manoj0077:
We even closed the socket file descriptor after some time it is still in hanging state...we are now making recvfrom a non blocking call ..and will let you know the results...any suggestions you have for us ?

PR #170, opened 2016-11-08
Add timeout for send and receive

aboch:
Because of kernel bugs, we have seen cases where the call to netlink socket send or receive will wait indefinitely causing docker engine to hang.

This PR allows client to set a timeout for socket send and receive for a netlink handle via `Handle.SetTimeout()`, in order to gracefully workaround the above situation.

A timeout logic was also requested in #131 

### DIFF
--- a/handle_linux.go
+++ b/handle_linux.go
@@ -1,7 +1,9 @@
 package netlink
 
 import (
+	"fmt"
 	"syscall"
+	"time"
 
 	"github.com/vishvananda/netlink/nl"
 	"github.com/vishvananda/netns"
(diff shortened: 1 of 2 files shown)


cpuguy83:
+1

aboch:
Opened https://github.com/vishvananda/netlink/pull/170 for this.

nl/nl_linux.go
```diff
@@ -238,54 +246,82 @@ func (req *NetlinkRequest) Execute(sockType int, resType uint16) ([][]byte, erro
...
+	go func() {
+		sch <- s.Send(req)
+	}()
+
+	select {
+	case err := <-sch:
+		if err != nil {
+			return nil, err
+		}
+	case <-time.After(req.Timeout):
```
LK4D4: this will leak a lot of timers. You should use `NewTimer` and stop them.

handle_linux.go
```diff
@@ -62,6 +66,15 @@ func newHandle(newNs, curNs netns.NsHandle, nlFamilies ...int) (*Handle, error)
 	return h, nil
 }
 
+// SetTimeout sets the amount of time the library will wait for the socket send
+// and for the socket receive on this handle to terminate before declaring a
+// a timeout failure to the client.
+func (h *Handle) SetTimeout(to time.Duration) {
```
LK4D4: I'd better go with `NewHandleWithTimeout` to avoid mutexes. But it's up to @vishvananda for decide of course.

cpuguy83:
Seems a little funky that this is implemented using just go with a timeout waiting for the goroutine to return rather on the socket itself.

nl/nl_linux.go
```diff
@@ -238,54 +246,82 @@ func (req *NetlinkRequest) Execute(sockType int, resType uint16) ([][]byte, erro
...
-				break done
 			}
 		}
+		rch <- err
+	}()
+
+	select {
+	case err := <-rch:
+		return res, err
+	case <-time.After(req.Timeout):
```
LK4D4: I think send and read should be in the same goroutine for having one timeout for whole Execute and simplify code.

LK4D4:
@cpuguy83 yeah, there is `syscall.SO_RCVTIMEO`.

nl/nl_linux.go
aboch: I have seen reports where the hung happen in both the send and receive phase, I wanted the client to know during which phase the timeout happened.

handle_linux.go
aboch: I wanted to avoid a precedence for adding a new constructor each time we add a new option.
This is more flexible and client will likely set the timeout once at handle creation. After that it is only going to be read locks.

aboch:
Thanks @cpuguy83 @LK4D4 
Agree it would be ideal to let the socket handle the timeout.
I will take a look at it.

aboch:
@cpuguy83 @LK4D4 Addressed the comments. Thanks. PTAL.

cpuguy83:
🎉  thanks @vishvananda!

Sn0rt:
it may a kernel bug.

you can try insert a probe into `nettling_recvmsg` and  `nettling_recvmsg.return` by systemtap , then count the number of call and return .

the count will be stable temporary , you will the its different between the number of call and return .

PR #172, opened 2016-11-14
Experimental support for GRE tunnels

justinsb:
Right now this is little more than a copy-and-paste of gretap, with the "extra" gretap fields removed.

Works, but wondering if you had a different preference for how this should be structured?  Will then clean it up accordingly...

### DIFF
--- a/link.go
+++ b/link.go
@@ -589,6 +589,33 @@ func (gretap *Gretap) Type() string {
 	return "gretap"
 }
 
+// Gre devices must specify LocalIP and RemoteIP on create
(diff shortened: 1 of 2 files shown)


vishvananda:
I think I'm fine with it being separate like this with duplicated code. You could experiement with  GreTap embedding a Gre, but it might end up being uglier that way.

vishvananda:
Any updates on this?

vishvananda:
closing this for now. Feel free to update and reopen if you want to move it forward

PR #173, opened 2016-11-18
Expose LinkDeserialize function

lebauce:


vishvananda:
I'm ok with this change, but can you explain why you need it?

lebauce:
Of course, it should have been part of the pull request description.

I'm using `nl.Subscribe` to listen for netlink events and use `Epoll` to get messages from the file descriptor. I need 2 methods to parse link message (as done in LinkSubscripbe) :  `DeserializeIfInfomsg` that is already exposed and `LinkDeserialize`.

PR #176, opened 2016-11-20
Add support for VTI and IPIP

jc-m:
Adding support for IPIP (IPTUN)  and VTI.


jc-m:
The test fail but is it expected. The first time either VTI or Tun is invoked a new interface is created by the kernel module. Since the test simplistically checks for the number of interface - the one run effectively create 2 but deletes only 1. I'll make a custom test or fix the base one.

link_test.go
```diff
@@ -120,6 +119,20 @@ func testLinkAddDel(t *testing.T, link Link) {
...
+		_, ok := result.(*Vti)
+		if !ok {
+			t.Fatal("Result of create is not a vti")
+		}
+	}
+
+	if _, ok := link.(*Iptun); ok {
+		_, ok := result.(*Iptun)
+		if !ok {
+			t.Fatal("Result of create is not a vti")
```
vishvananda: minor typo in the fatal here

jc-m:
Fixed

PR #189, opened 2017-01-13
fix a bug for net statistics when the whole network flow bytes is beyond 4GB, i.e. uint32

KevinTHU:
when the network flow bytes is growing up beyond 4GB, i.e. uint32, the network flow bytes will return to count from zero, this is not correct, and some golang program which depends on this package will have a wrong return, like "Docker", you know, the docker command docker stats will return a wrong result for network flow monitor.

I hope this can be fix, and reference to Docker community. thx~

Signed-off-by: Xin He <hexin@jiedaibao.com>

### DIFF
--- a/link.go
+++ b/link.go
@@ -83,29 +83,29 @@ func NewLinkAttrs() LinkAttrs {
 Ref: struct rtnl_link_stats {...}
 */
 type LinkStatistics struct {
-	RxPackets         uint32
-	TxPackets         uint32
-	RxBytes           uint32
-	TxBytes           uint32
-	RxErrors          uint32
-	TxErrors          uint32
-	RxDropped         uint32
-	TxDropped         uint32
-	Multicast         uint32
-	Collisions        uint32
-	RxLengthErrors    uint32
-	RxOverErrors      uint32
-	RxCrcErrors       uint32
-	RxFrameErrors     uint32
-	RxFifoErrors      uint32
-	RxMissedErrors    uint32
-	TxAbortedErrors   uint32
-	TxCarrierErrors   uint32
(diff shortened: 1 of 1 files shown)


KevinTHU:
I think the CI test is not quit right, need some one to comment, thx~

aboch:
@KevinTHU 
Take a look at `parseLinkStatistics()`, it operates on `rtnl_link_stats` data returned by the kernel as `syscall.IFLA_STATS` atribute. Those are all `uint32`. That is correct.

What you may want to do is to see if the uint64 based stats are already returned under a different attribute and parse them accordingly. You may also want to add a new API for the long statistics values.

KevinTHU:
@aboch attr.Value[:] is []byte, not Uint32, and LinkStatistics is for data from []byte to be a struct. I read this code like that, maybe I can rebuild docker to find out wether I am right.

KevinTHU:
@aboch ok, I see you point, but why kernel `rtnl_link_stats` for netfilter is Uint32, I'm a little confuse for that.

aboch:
My guess is that support for uint64 stats was added later.
So I guess user space programs which prints the stats will first check whether the netlink msg contains `syscall.IFLA_STATS64` attribute, otherwise they will fall back to print the uint32 stats.

KevinTHU:
@aboch maybe you are right, but that need to be proved by testing or dig the code. I will test it next week by rebuiding docker, if I do not have time to read kernel.

aboch:
@KevinTHU It does. I opened #190 for this. PTAL.

vishvananda:
superseded by #191

PR #190, opened 2017-01-13
Provide access to the 64 bits link statistics

aboch:
This is an example of statistics for my `lo` interface being flooded by ping:

```
Statistics32: &{61691558 61691558 887395372 887395372 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0}
Statistics64: &{61691558 61691558 5182362668 5182362668 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0}
```

Note the byte rx count on the 32 stats has reset: `5182362668 - 2^32 = 887395372`

### DIFF
--- a/link_linux.go
+++ b/link_linux.go
@@ -13,7 +13,11 @@ import (
 	"github.com/vishvananda/netns"
 )
 
-const SizeofLinkStats = 0x5c
+const (
+	SizeofLinkStats   = 0x5c
+	SizeofLinkStats64 = 0xd8
+	IFLA_STATS64      = 0x17 // syscall pkg does not contain this one
+)
 
 const (
 	TUNTAP_MODE_TUN  TuntapMode = syscall.IFF_TUN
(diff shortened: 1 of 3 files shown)


vishvananda:
This appears to be correct, but I'm thinking it might be better to switch the existing statistics to uint64, and just fill the u64s from the 32 bit values if the kernel doesn't return the 64 bit stats. Thoughts?

aboch:
We could. But it would be an API breakage. Also I think it is good to keep this library in pair with what netlink does.

In fact just verified iproute2 checks first if netlink msg contains the 64 stats otherwise it falls back to the 32bits ones.

aboch:
In the sense, I would let Docker libnetwork, as an example, do the check about the stats and ultimately fill them into the same uint64 based stats object.

So that this library exposes all the information retrieved from kernel (i.e. stat64 supported/not supported)

aboch:
@vishvananda Based on your suggestion, I opened #191
Thanks.

link_linux.go
```diff
@@ -13,7 +13,11 @@ import (
 	"github.com/vishvananda/netns"
 )
 
-const SizeofLinkStats = 0x5c
+const (
+	SizeofLinkStats   = 0x5c
+	SizeofLinkStats64 = 0xd8
+	IFLA_STATS64      = 0x17 // syscall pkg does not contain this one
```
thaJeztah: @aboch should we open an issue or send a patch to golang?

vishvananda:
Yes I like the other one better. I know it is a slight change, but the netlink library is supposed to be a bit higher level. I don't like users having to check multiple fields for null and figuring out which to use.

PR #192, opened 2017-01-16
Add XfrmMonitor

brb:
The implementation in PR subscribes only to `XFRMNLGRP_EXPIRE`.

### DIFF
--- a/xfrm_monitor_linux.go
+++ b/xfrm_monitor_linux.go
@@ -1,0 +1,98 @@
+package netlink
+
+import (
+	"fmt"
+	"syscall"
+
+	"github.com/vishvananda/netns"
+
+	"github.com/vishvananda/netlink/nl"
+)
+
+type XfrmMsg interface {
+	Type() nl.XfrmMsgType
+}
+
+type XfrmMsgExpire struct {
+	XfrmState *XfrmState
+	Hard      bool
+}
+
+func (ue *XfrmMsgExpire) Type() nl.XfrmMsgType {
+	return nl.XFRM_MSG_EXPIRE
+}
+
+func parseXfrmMsgExpire(b []byte) *XfrmMsgExpire {
+	var e XfrmMsgExpire
+
+	msg := nl.DeserializeXfrmUserExpire(b)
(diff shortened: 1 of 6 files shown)


vishvananda:
This looks good. For consistency, can you add a nl/xfrm_monitor_linux_test.go that verifies the serialization in the same way as the other netlink objects?

brb:
Done. PTAL.

xfrm_monitor_linux.go
```diff
@@ -0,0 +1,98 @@
...
+	var e XfrmMsgExpire
+
+	msg := nl.DeserializeXfrmUserExpire(b)
+	e.XfrmState = xfrmStateFromXfrmUsersaInfo(&msg.XfrmUsersaInfo)
+	e.Hard = msg.Hard == 1
+
+	return &e
+}
+
+func XfrmMonitor(ch chan<- XfrmMsg, done <-chan struct{}, errorChan chan<- error,
```
aboch: Given you are adding a new method, I think we should provide it also for the netlnk handle `func (h *Handle) XfrmMonitor(...)` in line with all the other existing xfrm methods.

aboch: Or add another method to allow the user to pass the netns to where to subscribe to `func XfrmMonitorAt(...), like it is done for `routeSubscribeAt` like it is being done for routes and link notifications.

brb: To be honest, I am not very keen on adding a code which might lead to various fringes. It is not possible to safely switch namespaces in the Go runtime. Please see https://groups.google.com/forum/#!msg/golang-nuts/ss1gEOcehjk/7d9UkcIBBAAJ for more details.

aboch: That's true, but consider the namespace switch will happen only for the time of the syscall to create the netlink socket in the netns an under thread lock.

Anyhow, this can be added later and should not block this PR.

brb: > That's true, but consider the namespace switch will happen only for the time of the syscall to create the netlink socket in the netns an under thread lock.

Unfortunately, that does not prevent the runtime from creating another thread which is used for scheduling goroutines.

Issue #198, opened 2017-02-02
Add support for NETLINK_SOCK_DIAG


sebbov:
I have a need to get TCP socket information using netlink and this library seems a good place to build on and add this support to.  Would you consider contributions for that?

PR #199, opened 2017-02-02
Add support for tcp diags.

sebbov:
Adds a single SocketGet function which gets the detail of a socket keyed by its local and remote
addresses.

Organized byte order functionality in a new order.go file.

https://github.com/vishvananda/netlink/issues/198

### DIFF
--- a/link_linux.go
+++ b/link_linux.go
@@ -29,7 +29,6 @@ const (
 	TUNTAP_ONE_QUEUE TuntapFlag = syscall.IFF_ONE_QUEUE
 )
 
-var native = nl.NativeEndian()
 var lookupByDump = false
 
 var macvlanModes = [...]uint32{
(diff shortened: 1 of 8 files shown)


sebbov:
Not sure how to set up the test correctly.  Without the setUpNetlinkTest, it works fine on my local machine.  Any suggestions?

vishvananda:
i don't think you want to add the setUpNetlinkTest to the teardown. Does travis give you an error if you remove that? Also is this something that the netlink c library or iproute cli has commands for? I'm curious how one generally uses this from c, as it doesn't look like iproute2 has any uses of SOCK_DIAG_BY_FAMILY. It is possible that the functionality should be in a separate library.

vishvananda:
i see you had an initial failure. It looks like your issue might be a race between your goroutines. I don't see you waiting for your listen goroutine to get past accept before calling connect. I think if you add some sync there it will pass on travis.

sebbov:
libnl does seem to support NETLINK_INET_DIAG (which is the old name for NETLINK_SOCK_DIAG) iiuc.    Re iproute2, i believe the "ss" util that is part of it uses this API.  I hadn't looked at either or tried to mimic those existing interface in any way, but can take a look at that if you want.  The one example use I had found of this is:  https://github.com/kristrev/inet-diag-example/blob/master/inet_monitor.c

Right, the travis failure occurred before I used setUpNetlinkTest as well.

I don't see the race you mention:  I do start the listener before i connect the client, which should be all that matters (accepting can occur later, right).

Is there anything in the travis environment preventing a test to just listen on a localhost address out of the box?  It seems something is missing before I can do that (in this environment specifically; this works on my workstation).

thanks,
-seb.

vishvananda:
It could be an intermittent failure. I can't rerun the test unless you remove that second commit. It also could have to do with one of the goroutines starting in a separate network namespace, which can happen. You might have some luck using
  defer tearDown()

at the beginning of your test. Although that might mess up your accept goroutine.

sebbov:
Sorry that second commit was intended to call setUpNetlinkTest from my test (not from SetupNetlinkTest itself.  Fixing that didn't do it though.  It does seem like my accepting go routine cannot accept (different "network namespace"?).  However I realized that for the purpose of this test, we never need to accept the client's connection, so I removed that go routine altogether.  With that and the setUpNetlinkTest removed, it now passes.

vishvananda:
Excellent! Would you mind squashing all of those into a single commit?

Vish

sebbov:
Done!

sebbov:
Thanks Vish!

PR #200, opened 2017-02-03
Add RouteReplace

lslatr:
This PR adds RouteReplace which is RouteAdd with the flag NLM_F_REPLACE instead of NLM_F_EXCL for RouteAdd

### DIFF
--- a/route_linux.go
+++ b/route_linux.go
@@ -61,7 +61,22 @@ func RouteAdd(route *Route) error {
 // RouteAdd will add a route to the system.
 // Equivalent to: `ip route add $route`
 func (h *Handle) RouteAdd(route *Route) error {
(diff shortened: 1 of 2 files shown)


lslatr:
This is a simple API add... I think the Makefile is messed up and is causing the automated tests to fail.

vishvananda:
The makefile is fine. You have a formatting issue. Try running go fmt on your file and resubmitting. Also, it is worth adding a simple test for it.  You can copy the TestRouteAddDel and modify it.

lslatr:
Got it... I'm a go newbie.  I'll add a test.

lslatr:
@vishvananda ok added a ReplaceRoute test... think we're good to go.

PR #201, opened 2017-02-06
Properly skip tests which require root.

sebbov:
All these tests currently fail with 'Operation not permitted' when run as
non-root.
With these changes, running tests as non-root passes.


vishvananda:
Looks good. You know we might be able to get these tests to work using user namespaces, but that is a project for another day.

PR #203, opened 2017-02-19
MPLS support

ishidawataru:


route_linux.go
```diff
@@ -60,6 +61,74 @@ func (n *NexthopInfo) ListFlags() []string {
...
 
+type MPLSDestination struct {
+	Labels []int
+}
+
+func (d *MPLSDestination) Family() int {
+	return nl.FAMILY_MPLS
+}
+
+func (d *MPLSDestination) Decode(buf []byte) error {
```
vishvananda: seems odd to use serialize/decode. I think it should be encode/decode or serialize/deserialize here

route.go
```diff
@@ -25,15 +40,36 @@ type Route struct {
 	Type       int
 	Tos        int
 	Flags      int
+	MPLSDst    int
```
vishvananda: I believe 0 is a valid mpls label id, so it might be necessary to use an *int here?

vishvananda:
thanks for this. A couple of comments inline. Also if you could squash into a single commit that would be great.

ishidawataru:
Reflected your review comments, and squashed into one commit.
PTAL

PR #204, opened 2017-02-20
Fix compilation errors for non linux platforms.

aaithal:
#### Summary
The `go get` command and `make` both fail when executed on
non-linux platforms. Modified it so that there are no
compilation errors when developing in such an
environment.

#### Testing Done
Ensured that tests pass on OS X and that builds are successful on Linux.


vishvananda:
Thanks for these changes.

PR #206, opened 2017-03-18
Add capability of specifying TcU32Sel to U32

tfukushima:
With the existing codebase the users can create u32 filters but [only with all match sel key, which is automatically inserted](https://github.com/vishvananda/netlink/blob/fe3b5664d23a11b52ba59bece4ff29c52772a56b/filter_linux.go#L131-L136). This prevents the users from specifying arbitrary sel keys.

This series of patches enables `U32 ` to have a pointer to `nl.TcU32Sel`, and therefore the users can specify any sel keys they want.

The first patch adds the utility functions that convert a 32 bit or a 16 bit value into a network ordered value in `order.go`. These, especially `ToNetworkOrder32` function, are required to set the appropriate value to `Mask` and `Val` in `nl.TcU32Key`, which are supposed to be in the network order. Tests for them are added in a new `order_test.go` file.

The second patch modifies `U32` struct to have a pointer to `nl.TcU32Sel`. This can be `nil` and if it is `nil`, `U32` works as it did. Otherwise the users set it to the pointer to `nl.TcU32Sel` that contains arbitrary `nl.TcU32Key`s. A new test case is added to `filter_test.go` although it has the creation and deletion of a HTB class besides the ones of the u32 filter for using its handle as `ClassID` in `U32`.

### DIFF
--- a/filter.go
+++ b/filter.go
@@ -1,6 +1,10 @@
 package netlink
 
-import "fmt"
+import (
+	"fmt"
+
+	"github.com/vishvananda/netlink/nl"
+)
 
 type Filter interface {
 	Attrs() *FilterAttrs
@@ -180,11 +184,46 @@ func NewMirredAction(redirIndex int) *MirredAction {
 	}
 }
 
+// Constants used in TcU32Sel.Flags.
+const (
+	TC_U32_TERMINAL  = nl.TC_U32_TERMINAL
+	TC_U32_OFFSET    = nl.TC_U32_OFFSET
+	TC_U32_VAROFFSET = nl.TC_U32_VAROFFSET
+	TC_U32_EAT       = nl.TC_U32_EAT
+)
+
+// Sel of the U32 filters that contains multiple TcU32Key. This is the copy
+// and the frontend representation of nl.TcU32Sel. It is serialized into canonical
+// nl.TcU32Sel with the appropriate endianness.
+type TcU32Sel struct {
+	Flags    uint8
+	Offshift uint8
+	Nkeys    uint8
+	Pad      uint8
+	Offmask  uint16
+	Off      uint16
+	Offoff   int16
+	Hoff     int16
+	Hmask    uint32
+	Keys     []TcU32Key
+}
+
+// TcU32Key contained of Sel in the U32 filters. This is the copy and the frontend
+// representation of nl.TcU32Key. It is serialized into chanonical nl.TcU32Sel
+// with the appropriate endianness.
+type TcU32Key struct {
+	Mask    uint32
+	Val     uint32
+	Off     int32
+	OffMask int32
+}
+
 // U32 filters on many packet related properties
 type U32 struct {
 	FilterAttrs
 	ClassId    uint32
 	RedirIndex int
+	Sel        *TcU32Sel
 	Actions    []Action
 }
 
--- a/filter_test.go
+++ b/filter_test.go
@@ -92,6 +92,163 @@ func TestFilterAddDel(t *testing.T) {
 	}
 }
 
+func TestAdvancedFilterAddDel(t *testing.T) {
+	tearDown := setUpNetlinkTest(t)
+	defer tearDown()
+	if err := LinkAdd(&Ifb{LinkAttrs{Name: "baz"}}); err != nil {
+		t.Fatal(err)
+	}
+	link, err := LinkByName("baz")
+	if err != nil {
+		t.Fatal(err)
+	}
+	if err := LinkSetUp(link); err != nil {
+		t.Fatal(err)
+	}
+	index := link.Attrs().Index
+
+	qdiscHandle := MakeHandle(0x1, 0x0)
+	qdiscAttrs := QdiscAttrs{
+		LinkIndex: index,
+		Handle:    qdiscHandle,
+		Parent:    HANDLE_ROOT,
+	}
+
+	qdisc := NewHtb(qdiscAttrs)
+	if err := QdiscAdd(qdisc); err != nil {
+		t.Fatal(err)
+	}
+	qdiscs, err := QdiscList(link)
+	if err != nil {
+		t.Fatal(err)
+	}
+	if len(qdiscs) != 1 {
+		t.Fatal("Failed to add qdisc")
+	}
+	_, ok := qdiscs[0].(*Htb)
+	if !ok {
+		t.Fatal("Qdisc is the wrong type")
+	}
+
+	classId := MakeHandle(0x1, 0x46cb)
+	classAttrs := ClassAttrs{
+		LinkIndex: index,
+		Parent:    qdiscHandle,
+		Handle:    classId,
+	}
+	htbClassAttrs := HtbClassAttrs{
+		Rate:   512 * 1024,
+		Buffer: 32 * 1024,
+	}
+	htbClass := NewHtbClass(classAttrs, htbClassAttrs)
+	if err = ClassReplace(htbClass); err != nil {
+		t.Fatalf("Failed to add a HTB class: %v", err)
+	}
+	classes, err := ClassList(link, qdiscHandle)
+	if err != nil {
+		t.Fatal(err)
+	}
+	if len(classes) != 1 {
+		t.Fatal("Failed to add class")
+	}
+	_, ok = classes[0].(*HtbClass)
+	if !ok {
+		t.Fatal("Class is the wrong type")
+	}
+
+	u32SelKeys := []TcU32Key{
+		TcU32Key{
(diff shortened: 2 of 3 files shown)


tfukushima:
TestXfrmMonitorExpire was timed out in the Travis CI job.

I run tests on my local env and I got failures from tests that are passed in Travis.

```
$ cat /etc/lsb-release 
DISTRIB_ID=Ubuntu
DISTRIB_RELEASE=14.04
DISTRIB_CODENAME=trusty
DISTRIB_DESCRIPTION="Ubuntu 14.04.5 LTS"
$ sudo /usr/local/go/bin/go test 
--- FAIL: TestAddr (0.01s)
        addr_test.go:80: Address flags not set properly, got=0, expected=128
--- FAIL: TestLinkAddDelVxlanGbp (0.08s)
        link_test.go:187: Vxlan.GBP doesn't match
--- FAIL: TestLinkAddDelIPVlanL2 (0.01s)
        link_test.go:28: operation not supported
--- FAIL: TestLinkAddDelIPVlanL3 (0.12s)
        link_test.go:28: operation not supported
2017/03/18 05:37:57 Skipped test because it requires MPLS support.
FAIL
exit status 1
FAIL    github.com/vishvananda/netlink  10.070s
```

My tests passed.
```
$ sudo /usr/local/go/bin/go test -run TestAdvancedFilterAddDel
PASS
ok      github.com/vishvananda/netlink  0.012s
$ sudo /usr/local/go/bin/go test -run TestToNetworkOrder32
PASS
ok      github.com/vishvananda/netlink  0.003s
$ sudo /usr/local/go/bin/go test -run TestToNetworkOrder16
PASS
ok      github.com/vishvananda/netlink  0.003s
```

tfukushima:
I changed the name of the link created in the test from "foo" to "baz" just in case since it seems the tests run in parallel. Now all tests passed.

order.go
```diff
@@ -23,10 +23,40 @@ func htons(val uint16) []byte {
...
 func ntohl(buf []byte) uint32 {
 	return binary.BigEndian.Uint32(buf)
 }
 
 func ntohs(buf []byte) uint16 {
 	return binary.BigEndian.Uint16(buf)
 }
+
+// Convert a 32 bit value to the network byte order
+func ToNetworkOrder32(n uint32) uint32 {
```
vishvananda: just use hotnl, which already exists

order.go
```diff
@@ -23,10 +23,40 @@ func htons(val uint16) []byte {
 	return bytes
 }
 
+func htohl(val uint32) []byte {
```
vishvananda: no need for these new functions

filter_test.go
```diff
@@ -92,6 +94,143 @@ func TestFilterAddDel(t *testing.T) {
...
+		t.Fatal("Failed to add class")
+	}
+	_, ok = classes[0].(*HtbClass)
+	if !ok {
+		t.Fatal("Class is the wrong type")
+	}
+
+	u32SelKeys := []nl.TcU32Key{
+		nl.TcU32Key{
+			Mask:    ToNetworkOrder32(0xff),
```
vishvananda: just use htonl for these

order_test.go
```diff
@@ -0,0 +1,64 @@
+package netlink
```
vishvananda: no need for these tests

vishvananda:
Thank you for the code but the network order additions are not necessary. There is already a hotns and hotnl. You should be able to use those in place of your new functions.

tfukushima:
Hi @vishvananda, thanks for your review.

`ToNetworkOrder32` and `ToNetworkOrder16` were bad naming choices and I renames them `Htonl` and `Htons` looking after `htonl(3)` and `htons(3)` in C since they behave in the same way. I eliminated `htohl` and `htons` as well since I renewed the implementations of `Htonl` and `Htons` with the oldschool shift base ones.

`nl.TcU32Key` requires `uint32` values in the network order for its `Mask` and `Val`. None of the existing functions doesn't conform to this requirement since the existing ones convert a `uint32` or `uint16` value to bytes but not another `uint32` or `uint16`. Plus they, the lower case first character functions are not exposed to the external code. If we want to create a u32 filter with sel keys In another project, we need `Htonl`, especially in the little-endian system.

order_test.go
```diff
@@ -0,0 +1,64 @@
...
+const bitsInByte = 8
+
+func TestHtonl(t *testing.T) {
+	totalBytesInUint32 := 4
+	n := uint32(0x0a0b0c0d)
+	converted := Htonl(n)
+
+	if native == binary.BigEndian {
+		if converted != n {
+			t.Errorf("ToNetworkOrder32(%x) != %x in the bigendian system.", n, converted)
```
vishvananda: error message still has the old name

vishvananda:
It is best if users of the netlink package do not have to worry about byte order. Also they should not have to import the nl library. I think what we want is essentially a copy of the Sel struct in netlink that people can use. It can automatically convert the byte order when serializing and deserializing into the nl form. This also means that we could use the internal ones that go to bytes instead of the new ones you created.

tfukushima:
Hi @vishvananda,

I do agree `nl` package should not be exposed directly to the users. I copied `nl.TcU32Sel` and `nl.TcU32Key` from `nl` package as they are, without comments regarding the endianness though.

Now the endianness of attributes are handled in `FilterAdd` function in `filter_linux.go`. If the native endianness of the host is not the network order, Big-endian, the attributes are converted into Big-endian forms. Then `TcU32Sel` is casted into `nl.TcU32Sel`, which is identical to `TcU32Sel`, its `Keys` attribute is overwritten with the copied array of `nl.TcU32Key` and it is serialized into bytes with the `Serizlize` method of `nl.TcU32Sel`.  So the users don't need to care about the endianness at all. If the endianness of the host is the same as the network order,  `TcU32Sel` is just casted into `nl.TcU32Sel` or the other way around without copying them.

I avoided moving the existing the codebase in `nl` package as possible and just copied the necessary constants and structs. I removed `Ntonl` and `Ntons` functions from `order.go` and I deleted `order_test.go` as well. Please let me know if you had better idea than what I did, then I'll modify the patch.

filter.go
```diff
@@ -180,11 +184,46 @@ func NewMirredAction(redirIndex int) *MirredAction {
...
+	TC_U32_TERMINAL  = nl.TC_U32_TERMINAL
+	TC_U32_OFFSET    = nl.TC_U32_OFFSET
+	TC_U32_VAROFFSET = nl.TC_U32_VAROFFSET
+	TC_U32_EAT       = nl.TC_U32_EAT
+)
+
+// Sel of the U32 filters that contains multiple TcU32Key. This is the copy
+// and the frontend representation of nl.TcU32Sel. It is serialized into canonical
+// nl.TcU32Sel with the appropriate endianness.
+type TcU32Sel struct {
```
vishvananda: since it is an exact copy, can't you just do type TcU32Sel nl.TcU32Sel ?
Then I think you could copy via a cast instead of unsafe pointer copy.

filter.go
```diff
@@ -180,11 +184,46 @@ func NewMirredAction(redirIndex int) *MirredAction {
...
+	Offoff   int16
+	Hoff     int16
+	Hmask    uint32
+	Keys     []TcU32Key
+}
+
+// TcU32Key contained of Sel in the U32 filters. This is the copy and the frontend
+// representation of nl.TcU32Key. It is serialized into chanonical nl.TcU32Sel
+// with the appropriate endianness.
+type TcU32Key struct {
```
vishvananda: Same here

vishvananda:
A couple of minor changes inline. Otherwise it looks great.

tfukushima:
Thanks for your review, @vishvananda.

The problem of `type TcU32Sel nl.TcU32Sel` is `nl.TcU32Sel` has the nested array of `nl.TcU32Key`s as its `Keys` attribute. `[]TcU32Key` and `[]nl.TcU32Key` are regarded as different types since they are named types with different names, and therefore `TcU32Sel` can't be casted into `nl.TcU32Sel` and so for the other way around.

To solve this sort of problem, [tyep alias is being proposed](https://github.com/golang/proposal/blob/master/design/18130-type-alias.md) and in development now. It is expected we have it in Go 1.9 but it will take some time, more than three months.

So I believe it is inevitable to have the duplicated structs although they are pretty ugly and redundant. I thought about moving the structs into the toplevel, not in `nl`, but it leads `nl` package import the toplevel contents, which seemed a bad design. Let me know how do you think.

vishvananda:
Ah that makes sense. Ultimately it might be nice to have a more user friendly api for setting u32 without having to supply the low level offsets, but lets go ahead and merge this for now.

PR #210, opened 2017-03-29
Add support for the Clsact qdisc

dansiemon:


vishvananda:
Looks like you might need to modprobe some modules in travis for the tests to pass? See what they did here: https://github.com/vishvananda/netlink/pull/213

Also would you mind rebasing and/or squashing your changes?

PR #211, opened 2017-03-31
Add AddrReplace

stuart-mclaren:
Add an AddrReplace call (and tests).

Equivalent to:
 $ ip addr replace $addr dev $link


stuart-mclaren:
Thanks for reviewing. (Really useful library btw.)

PR #212, opened 2017-04-03
Enable/disable multicast snooping on bridges

craigfurman:
Hi,

We've added functions to enable/disable multicast snooping on bridges.

We're new to netlink and have a few areas in the PR we are unsure about. Of course there may well be more, please let us know.

Areas of uncertainty:

1. The byte array which translates to a bunch of control chars, followed by ASCII for "bridge", followed by more control chars. We reverse engineered the incantation by running [nltrace](https://github.com/socketpair/nltrace) against `ip link set $bridge type bridge mcast_snooping 0` (only very new versions of `iproute2` support this option). Is there a nicer way to formulate this?
1. We weren't sure if we defined `IFLA_BR_MCAST_SNOOPING` in the right place.

On the plus side, it does seem to work.

Craig & @teddyking


craigfurman:
It appears that the build has failed. It works on our machine (™).

In all seriousness we know that's not enough, is this something you could help us debug please? Perhaps the Travis kernel version is too old to support this operation.

vishvananda:
Yeah you will probably have to skip the test for travis. Other tests do the same. If you have time it might be nice to add a test for an older bridge parameter that will run correctly on travis. That way we can verify that the changes below actually work.

The data block definitely isn't going to cut it. You'll need to figure out how to construct the message properly.

Reading iproute2 it looks like you actually want to create a message that is similar to the LinkAdd message, only without syscall.NLM_F_CREATE|syscall.NLM_F_EXCL, and then append your extra bridge settings to that message. You can probably do that by splitting out LinkAdd into a new method called linkModify that takes flags as a parameter. Take a look at classModify for an example. In terms of adding the new constant that is a fine location, but you should probably add all of the flags from that set instead of just that one. Bonus points if you feel like adding any of the other bridge commands.

craigfurman:
We removed the `BridgeSetMcastSnoop` method, and are just using LinkAdd with `Bridge` objects that the caller can set struct fields on. Our use case actually only needs this field set at creation time, not modified later.

Thanks for the advice about the data block, we managed to construct it without literals.

We are looking into adding a test for an older bridge parameter, to demonstrate this working on Travis CI.

craigfurman:
@vishvananda We've added another bridge parameter (hello_time) in [a WIP branch on our fork](https://github.com/cloudfoundry-incubator/netlink/commit/9b5cd9a5dc7c5e18ff5425123cf21104695cc7ac).

The test passes for us on an Ubuntu 16.04 VM, but not on a 12.04 VM running the 3.13 kernel, like in the Travis build environment.

Looking at the [latest Linux kernel code as of today](https://github.com/torvalds/linux/blob/269c930e6698e6afa6d4606bfd241f4497597e79/include/uapi/linux/if_link.h#L232-L281), you can see all the netlink bridge flags. In the [3.13 version of that file](https://github.com/torvalds/linux/blob/v3.13/include/uapi/linux/if_link.h), it does not have any bridge-specific flags. It looks like all of them have been added since 2012, which does sound unlikely, I admit! Grepping the kernel codebase at tag `v3.13` for `HELLO_TIME` and other bridge flags turns up things related to `brctl`, but not netlink.

`iproute2` version 3.10 (from 2013), doesn't seem to use netlink to set bridge-specific parameters either (at first glance).

We're not too sure where to go from here, we're happy to PR in the bridge hello_time support either as part of this or another PR, but from the look of it, its test would also not be able to run on Travis.

vishvananda:
Ah so the bridge params were added to netlink recently. In that case it seems fine to just skip the test for travis. You can go ahead and add the extra parameter in.

One small change, if you are using a pointer for the snooping flag, I think it makes the most sense to not add the Attr to the message at all if it ithe parameter is nil, instead of defaulting it to true manually. That way we aren't maintaining default values separately from the kernel.

craigfurman:
Good suggestion, latest commit doesn't include the attrs when not set by client. `hello_time` and `mcast_snooping` tests pass on my Ubuntu 16.04 VM, tests pended in Travis due to kernel version.

craigfurman:
Hey @vishvananda , actually please hold off merging this for now. It turns out our use case requires us to be able to change multicast snooping for existing bridges as well, and we'd like to add that today.

craigfurman:
We added that feature back now, what do you think?

vishvananda:
This is looking reasonable. For consistency, hello time should also be a pointer and only added to the message if it is set. It looks like setting it to zero is a valid value to me.

craigfurman:
Ah, I didn't realise hello_time could be 0. We'll change this to a pointer

craigfurman:
Done, please let me know what you think

vishvananda:
sorry it took me a while to respond. This looks good. Could you:

a) add all of the BR flags to nl/link_linux.go using iota so others don't have to go look them up if they want to add other functionality.

b) rebase and squash everything into one commit

Happy to merge once you do that. Thanks.

craigfurman:
Done

PR #213, opened 2017-04-04
Conntrack support

fcrisciani:
These 2 commits are introducing the support for netfilter conntrack into the library.
This PR is a dependency for the fix of the bug: https://github.com/docker/docker/issues/8795

The first commit adds the logic to support the GET/FLUSH/DELETE of conntrack through netlink
The second commit adds a test on top

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,5 +4,10 @@ before_script:
   - sudo sed -i -e 's/^Defaults\tsecure_path.*$//' /etc/sudoers
   # modprobe ip_gre or else the first gre device can't be deleted
   - sudo modprobe ip_gre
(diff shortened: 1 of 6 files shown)


fcrisciani:
Looks like I will have to investigate the failure on Travis. 
Was thinking was missing netcat but with my last commit the test should had been skip

fcrisciani:
The issue was the missing modules in the kernel, added them into Travis file.

vishvananda:
This looks good. I'm thinking we may need a conntrack_unspecified.go with NotImplemented calls so that projects build on non-linux environments?

fcrisciani:
Sure, let me take care of that

PR #217, opened 2017-04-09
link-add: support hardware address

mie00:


vishvananda:
Have you verified that this actually works? The iproute2 code only modifies the hwadress via a set call.

squeed:
I wrote the same code (including tests!) and it worked for me - see https://github.com/vishvananda/netlink/compare/master...squeed:create-set-mac

squeed:
Also, from what I can tell, the current version of iproute2 *does* set the mac address on create; see https://git.kernel.org/pub/scm/linux/kernel/git/shemminger/iproute2.git/tree/ip/iplink.c#n588

Of course, this isn't the easiest code to grok.

vishvananda:
Hmm I see that you are correct, it is called in the create path as well. I will merge this. @mie00 can you rebase?

link_linux.go
```diff
@@ -672,7 +672,7 @@ func LinkAdd(link Link) error {
 // are taken fromt the parameters in the link object.
 // Equivalent to: `ip link add $link`
 func (h *Handle) LinkAdd(link Link) error {
-	// TODO: set mtu and hardware address
+	// TODO: set mtu
```
squeed: I believe MTU is set correctly now too, so this comment can be deleted.

mie00:
I am sorry for my late response, I updated the PR.

squeed:
Can you delete all of line 675? That comment is no longer correct.

mie00:
Updated.

PR #218, opened 2017-04-10
Ensure bond settings are populated correctly

stuart-mclaren:
Currently a LinkByName("bondX") doesn't return the bond specific attributes.

parseBondData needs to update the link that is passed in in order for
the bond's Mode, Miimon etc to be populated correctly.


vishvananda:
Thanks for this, merged

PR #219, opened 2017-04-12
addr: Parse address lifetime information from IFA_CACHEINFO attr

tohojo:
This adds parsing of the preferred and valid lifetime information from the
netlink IFA_CACHEINFO attribute. They are stored as PreferedLft and ValidLft in
the Addr struct if found.

Signed-off-by: Toke Høiland-Jørgensen <toke@toke.dk>

### DIFF
--- a/addr.go
+++ b/addr.go
@@ -10,11 +10,13 @@ import (
 // include a mask, so it stores the address as a net.IPNet.
 type Addr struct {
 	*net.IPNet
-	Label     string
-	Flags     int
-	Scope     int
-	Peer      *net.IPNet
-	Broadcast net.IP
+	Label       string
+	Flags       int
+	Scope       int
+	Peer        *net.IPNet
+	Broadcast   net.IP
+	PreferedLft int
+	ValidLft    int
 }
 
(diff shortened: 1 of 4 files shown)


vishvananda:
Thank you for this addition. I would prefer if the ifa_cacheinfo struct was added into nl with a serialize and deserialize function instead of "magically" pulling two variables out of the struct. The front-end representation in netlink seems fine however.

tohojo:
Vish Ishaya <notifications@github.com> writes:

> Thank you for this addition. I would prefer if the ifa_cacheinfo
> struct was added into nl with a serialize and deserialize function
> instead of "magically" pulling two variables out of the struct.

Can do. Didn't get that far when making sense of the code. Will resubmit...

> The front-end representation in netlink seems fine however.

Cool :)

-Toke

tohojo:
Something like this?

vishvananda:
exactly like that, thanks.

tohojo:
Awesome! Thanks for the prompt response, and for writing the library in the first place :)

PR #220, opened 2017-04-12
AddrUpdate: Include flags, scope and lifetimes

tohojo:
Having the extra data available in the notification for new addresses is useful
to, for instance, filter out temporary addresses or keep track of address
lifetimes.

Signed-off-by: Toke Høiland-Jørgensen <toke@toke.dk>


tohojo:
Actually, I was wondering why the AddrUpdate doesn't just include the Addr struct; but changing that would break the API, so went with this instead :)

tohojo:
Thanks again :)

PR #226, opened 2017-04-29
Get broadcast addr in AddrList

feiskyer:
`ip addr show` command shows broadcast address, but it is not set in `AddrList()`.

This PR sets broadcast addr so it is consistent with `ip addr show`.


feiskyer:
@vishvananda Could you have a look at this?

feiskyer:
@vishvananda Thanks.

PR #234, opened 2017-06-01
nl: Use atomic load/store for fd field

aaronlehmann:
`NetlinkSocket` has a `fd` field that is used by several methods. Since this is accessed without locks or atomics, calling methods that use it from different goroutines would lead to a data race. One example of this is `Receive` called concurrently with `Close`. The check of `s.fd` might work as intended, because `s.fd` can change between the check and the use of the variable. Also, the concurrent read and write to this field would probably trigger the golang race detector.

Use `atomic` to load/store this variable so that concurrent loads and stores are safe.

### DIFF
--- a/nl/nl_linux.go
+++ b/nl/nl_linux.go
@@ -459,7 +459,7 @@ func NewNetlinkRequest(proto, flags int) *NetlinkRequest {
 }
 
 type NetlinkSocket struct {
-	fd  int
+	fd  int32
 	lsa syscall.SockaddrNetlink
 	sync.Mutex
 }
@@ -470,7 +470,7 @@ func getNetlinkSocket(protocol int) (*NetlinkSocket, error) {
 		return nil, err
 	}
 	s := &NetlinkSocket{
-		fd: fd,
+		fd: int32(fd),
 	}
 	s.lsa.Family = syscall.AF_NETLINK
 	if err := syscall.Bind(fd, &s.lsa); err != nil {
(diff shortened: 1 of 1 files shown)


nl/nl_linux.go
```diff
@@ -585,30 +585,33 @@ func SubscribeAt(newNs, curNs netns.NsHandle, protocol int, groups ...uint) (*Ne
 }
 
 func (s *NetlinkSocket) Close() {
-	syscall.Close(s.fd)
-	s.fd = -1
+	fd := int(atomic.LoadInt32(&s.fd))
```
fcrisciani: If you want to avoid the race with the GetFd without using a mutex then you should use: `fd := SwapInt32(&s.fd, -1)`

nl/nl_linux.go
```diff
@@ -585,30 +585,33 @@ func SubscribeAt(newNs, curNs netns.NsHandle, protocol int, groups ...uint) (*Ne
...
+	syscall.Close(fd)
+	atomic.StoreInt32(&s.fd, -1)
```
fcrisciani: this can go away

nl/nl_linux.go
aaronlehmann: Good idea. Switched to using `SwapInt32`.

fcrisciani:
LGTM

fcrisciani:
ping @vishvananda

PR #236, opened 2017-06-06
Add bridge vlan support

chenchun:


vishvananda:
This looks good. Out of curiosity, what is bridge vlan used for? I've only ever used the a vlan device on an interface. Does it allow you to get vlan tagged packets off of a bridge without needing to create an intermediate untagged interface?

chenchun:
Yes, you won't need to create a vlan device to get packets tagged/untagged if using bridge device. 

PVID (Port VID)
- Untagged (and VID 0) packet is assigned this VID
- Per-port configuration
- Default PVID is none (untagged packet is discarded)

Egress policy untagged
- Outgoing packet that matches this policy get untagged
- Per-port per-vlan policy

More details, please refer to http://events.linuxfoundation.org/sites/events/files/slides/LinuxConJapan2014_makita_0.pdf .

PR #249, opened 2017-08-01
Add support to set spoof check attribute for VF

paravmellanox:
Add support to enable/disable spoof check for a VF


paravmellanox:
#249 Travis build failed in building netns. Appears unrelated to my change. I am sending patch for first time so not sure if anything wrong. 
I have tested my changes already.

vishvananda:
The failure is due to a formatting error. Try `go fmt` and then amend your commit. I really need to modify the test to actually print the formatting error instead of dying :)

paravmellanox:
Got it. dropping the patch.  Resending #249 with right formatting.

PR #253, opened 2017-08-08
Corrected library function to have Vf prefix

paravmellanox:
Corrected function signature to have correct name LinkSetVfTrust
instead of LinkSetTrust.
This aligns with code comment and rest of the other VF functions.


paravmellanox:
Please merge the change before it gets in the use with incomplete API name.

PR #257, opened 2017-08-16
Add LinkSetTxQLen

danwinship:
The library currently lets you set txqlen at link creation time but doesn't let you modify it afterward. OpenShift wants to be able to do this so that it can belatedly fix the txqlen of veths created by docker. (Right now it has to exec "ip link set".)


danwinship:
travis failed, but it's not related to this patch. seems to be test flakiness of some sort

vishvananda:
it looks like a few of the tests are failing on the new trusty ci runners. I haven't had time to investigate yet.

PR #261, opened 2017-08-25
Replace redundant copied u32 types with type aliases

tfukushima:
This patch replaces TcU32Sel and TcU32Key that are copied from nl
package with type aliases for the original types. This eliminates the
usages of unsafe packages and redundant copied structs that are
identical to the original ones.

Type aliases are newly introduced in Go 1.9 and it is not backward
compatibile. Therefore this patch shall be merged only if the project is
entirely migrated to Go 1.9.

See #206.

Signed-off-by: Taku Fukushima <taku@soracom.jp>

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,6 @@
 language: go
+go:
+  - 1.9
 before_script:
   # make sure we keep path in tact when we sudo
   - sudo sed -i -e 's/^Defaults\tsecure_path.*$//' /etc/sudoers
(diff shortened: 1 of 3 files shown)


vishvananda:
I don't think we want to require go 1.9 yet.

aboch:
I agree with @vishvananda I would not upgrade to 1.9 for this. Consider this project is imported by several other ones. If they happen to use the `filter` APIs, then they would have to move to 1.9 too in order to compile.

I will close this PR for now. When time comes to move to 1.9, we/you can reopen it.

Thanks!

PR #263, opened 2017-08-29
Add support for GRE Tunneling & static NHRP support via ip neighbour

mcastelino:
Two features included in this PR
1. Add support for creating GRE tunnels

   This is equivalent to
   Point to Point:
    ```
    ip tunnel add tun4 mode gre local 192.0.2.1 remote 203.0.113.6 key 123
    ```

    Point to Multipoint:
    ```
    ip tunnel add tun8 mode gre local 192.0.2.1 key 1234
    ```

2. Add support for specifing an IP Address as a lladdr in the case of ip neighbour

The ip neighbour supports adding of peers statically using commands where the lladdr is an IP address.
```
ip neighbor add 10.0.0.2 lladdr 203.0.113.6 dev tun8
```
This is used in the case of point-to-multipoint GRE to setup the remote end point of the tunnel
Note that link-layer address and neighbor address are both IP addresses

### DIFF
--- a/link.go
+++ b/link.go
@@ -732,6 +732,28 @@ func (iptun *Vti) Type() string {
 	return "vti"
 }
 
+type Gretun struct {
+	LinkAttrs
+	Link     uint32
+	IFlags   uint16
+	OFlags   uint16
+	IKey     uint32
+	OKey     uint32
+	Local    net.IP
+	Remote   net.IP
+	Ttl      uint8
+	Tos      uint8
+	PMtuDisc uint8
+}
+
+func (gretun *Gretun) Attrs() *LinkAttrs {
+	return &gretun.LinkAttrs
+}
+
+func (gretun *Gretun) Type() string {
+	return "gre"
+}
+
 type Vrf struct {
 	LinkAttrs
 	Table uint32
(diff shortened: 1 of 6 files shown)


mcastelino:
@vishvananda the CI seems to be broken.

vishvananda:
@mcastelino this code seems to be broken with kernel 4.15:
```
--- FAIL: TestNeighAddDelLLIPAddr (0.04s)
	neigh_test.go:94: Dump does not contain: {11 0 128 0 0 198.51.100.2  198.51.100.1 0 0}: [{11 2 128 3 0 0.0.0.0  198.51.100.1 0 0}]
	neigh_test.go:101: Failed to NeighDel: no such file or directory
```
It appears that the code for NeighList is no longer getting the IP value when LLIPAddr is set.

mcastelino:
Will take a look to see what has changed in the kernel.

mcastelino:
@vishvananda looks like the kernel behavior changed for ipip tunnels. I have switched the tunnel type to gre and the unit test case passes. This matches what we used in our project.  This ensures we have at least one working test for lladdr to be an IP address.

The PR to fix this test case has been submitted https://github.com/vishvananda/netlink/pull/451

**Note:** I am unsure if the kernel behavior for ipip tunnels is in fact correct. That needs further investigation, The behavior however is the same even with the ip tool. So we match the behavior.

PR #264, opened 2017-09-02
Xdp updates

CtrlZvi:


CtrlZvi:
Looking at the difference in output between this test (failed) and master (succeeding), it looks like the tests that are failing all fail to remove qdisc, which this PR didn't touch. Those same tests fail for me locally when running tests in master. Could it have something to do with the configuration differences between master's tests and this PR's tests in Travis? (For example, master was tested on  Ubuntu 12.04 and the PR on Ubuntu 14.04)

PR #268, opened 2017-09-12
addr: don't log errors happening when receiving notifications

vincentbernat:
A user may have its own log handling. Another solution would be to
make this configurable by allowing a user to provide a logger. A
better solution would be to transmit the errors to some handler to let
the package user knows about them (notably the nature of the receive
error).

This is not done here since this is the only use of `log.Printf`. For
example, when listening for routes, errors are not displayed.

### DIFF
--- a/addr_linux.go
+++ b/addr_linux.go
@@ -2,7 +2,6 @@ package netlink
 
 import (
 	"fmt"
-	"log"
 	"net"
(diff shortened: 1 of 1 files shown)


aboch:
Can you please close this PR, since the changes it carries were already merged as part of #271 ?

PR #270, opened 2017-09-12
handle: get/set receive buffer size

vincentbernat:
When receiving a lot of route changes (10,000 routes are enough), the
default receive buffer size (value of
`/proc/sys/net/core/rmem_default`) is too small and we get a `ENOBUF`
error. A user may want to increase the buffer size up to the value of
`/proc/sys/net/core/rmem_max` (by default, this is the same value). A
`SetSocketReceiveBufferSize()` function is provided to this
destination.

Possible improvements:

 1. automatically increase receive buffer size in higher level
    functions until we hit a maximum (get an error and/or the current
    value is smaller than expected)

 2. accept a "force" argument to use `SO_RCVBUFFORCE` to increase the
    value over `rmem_max` value

### DIFF
--- a/handle_linux.go
+++ b/handle_linux.go
@@ -61,10 +61,14 @@ func (h *Handle) SetSocketTimeout(to time.Duration) error {
 // SetSocketReceiveBufferSize sets the receive buffer size for each
 // socket in the netlink handle. The maximum value is capped by
 // /proc/sys/net/core/rmem_max.
-func (h *Handle) SetSocketReceiveBufferSize(size int) error {
+func (h *Handle) SetSocketReceiveBufferSize(size int, force bool) error {
+	opt := syscall.SO_RCVBUF
+	if force {
+		opt = syscall.SO_RCVBUFFORCE
+	}
 	for _, sh := range h.sockets {
 		fd := sh.Socket.GetFd()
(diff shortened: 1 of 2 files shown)


handle_linux.go
```diff
@@ -61,10 +61,14 @@ func (h *Handle) SetSocketTimeout(to time.Duration) error {
 // SetSocketReceiveBufferSize sets the receive buffer size for each
 // socket in the netlink handle. The maximum value is capped by
 // /proc/sys/net/core/rmem_max.
-func (h *Handle) SetSocketReceiveBufferSize(size int) error {
+func (h *Handle) SetSocketReceiveBufferSize(size int, force bool) error {
```
aboch: This is a breaking change for projects which use this API.
Please make this a variadic function instead:

`func (h *Handle) SetSocketReceiveBufferSize(size int, options... SocketOption) error {`

vincentbernat: This function was introduced in this same PR.

aboch: Ah, I was looking at one commit's diff only. 👍

vincentbernat: This was put in a different commit because it makes the signature of the function a bit unfriendly. So, it's easy to remove this commit if we don't want it.

PR #271, opened 2017-09-12
Ability to grab errors from subscribe functions

vincentbernat:
For link, address, route, add a `WithOptions` variant to the
`*Subscribe()` function to specify a namespace and an error
callback to retrieve fatal errors. Those options can be extended in the future without adding
more functions. For example, it could be possible to subscribe only
for a given family by adding a `Family` member to the appropriate
struct.

As a minor change, the private function is always suffixed by `At`,
since it was the case for route and raw netlink functions (but not for
address and link).

This also modifies the behavior of `AddrSubscribe*()` function to turn
parse errors into fatal errors, as it happens with the other functions.

This supersedes #268.


vincentbernat:
Another "option" would be the socket size. I'll wait for #270 to propose something.

PR #280, opened 2017-09-26
Adding test to verify that close netlink unblock receive

fcrisciani:
Looks like netlink sockets once closed are not waking up the go routines that are waiting on the recv.
This can create easily go routine leaks if there is a go routine that is listening for events and another thread just closes the socket.
At the moment I intentionally added the timeout only on the netlink socket created by the Subscribe method but maybe we can extend the same behavior also to the other netlink socket creation.

Signed-off-by: Flavio Crisciani <flavio.crisciani@docker.com>


fcrisciani:
@vishvananda PTAL

fcrisciani:
maybe it's worth not propagating the timeout error, will make the change on this patch

PR #281, opened 2017-09-27
added encapsulation attributes for Iptun and Sittun to support SIT tunnels

skoef:
The `Iptun` struct didn't support encapsulation attributes like source or destination port yet, while `Gretap` did and the feature appeared for both interface types around the same time. I basically copied the parsing and adding of the attributes from `Gretap` to `Iptun` and it seems to work fine.


skoef:
For this project I'm doing I would like SIT tunnel support in this library as well. Since these links are similar to `Iptun` I could ofcourse copy/paste relevant functions and type of `Iptun` but I'm not sure that is the way to go here

vishvananda:
This looks reasonable. Can you squash it down into a single commit?

Thanks

skoef:
I thought github featured a squash&merge button, nevertheless: it's squashed, no problem!

vishvananda:
it does support squash and merge, but I prefer rebase and merge to keep a linear commit tree. :)

PR #282, opened 2017-09-27
Support LWTUNNEL_ENCAP_SEG6

ebiken:
Added LWTUNNEL_ENCAP_SEG6 support.
With this change, one can do things similar to below iproute2 command.

```
> Add seg6 routes
ip route add 10.0.0.20/32 encap seg6 mode inline segs fc00:00dd::1,fc00:00dd::2 dev enp0s3
ip route add 10.0.0.30/32 nexthop encap seg6 mode encap segs fc00:00bc::1 dev enp0s3 weight 10 \
nexthop encap seg6 mode inline segs fc00:00bc::2 dev enp0s8 weight 20
ip -6 route add fc00:000b::/64 encap seg6 mode encap segs fc00:00bc::1 dev enp0s3

> View seg6 routes
dev2222> ip route
10.0.0.20  encap seg6 mode inline segs 3 [ fc00:dd::1 fc00:dd::2 :: ] dev enp0s3 scope link
10.0.0.30
        nexthop  encap seg6 mode encap segs 1 [ fc00:bc::1 ] dev enp0s3 weight 10
        nexthop  encap seg6 mode inline segs 2 [ fc00:bc::2 :: ] dev enp0s8 weight 20
dev2222>ip -6 route
fc00:b::/64  encap seg6 mode encap segs 1 [ fc00:bc::1 ] dev enp0s3 metric 1024 pref medium
```

### DIFF
--- a/nl/seg6_linux.go
+++ b/nl/seg6_linux.go
@@ -1,0 +1,111 @@
+package nl
+
+import (
+	"errors"
+	"fmt"
+	"net"
+)
+
+type IPv6SrHdr struct {
+	nextHdr      uint8
+	hdrLen       uint8
+	routingType  uint8
+	segmentsLeft uint8
+	firstSegment uint8
+	flags        uint8
+	reserved     uint16
+
+	Segments []net.IP
+}
+
+func (s1 *IPv6SrHdr) Equal(s2 IPv6SrHdr) bool {
+	if len(s1.Segments) != len(s2.Segments) {
+		return false
+	}
+	for i := range s1.Segments {
+		if s1.Segments[i].Equal(s2.Segments[i]) != true {
+			return false
+		}
+	}
+	return s1.nextHdr == s2.nextHdr &&
+		s1.hdrLen == s2.hdrLen &&
+		s1.routingType == s2.routingType &&
+		s1.segmentsLeft == s2.segmentsLeft &&
+		s1.firstSegment == s2.firstSegment &&
+		s1.flags == s2.flags
+	// reserved doesn't need to be identical.
+}
+
+// seg6 encap mode
+const (
+	SEG6_IPTUN_MODE_INLINE = iota
+	SEG6_IPTUN_MODE_ENCAP
+)
+
+// number of nested RTATTR
+// from include/uapi/linux/seg6_iptunnel.h
+const (
+	SEG6_IPTUNNEL_UNSPEC = iota
+	SEG6_IPTUNNEL_SRH
+	__SEG6_IPTUNNEL_MAX
+)
+const (
+	SEG6_IPTUNNEL_MAX = __SEG6_IPTUNNEL_MAX - 1
+)
+
+func EncodeSEG6Encap(mode int, segments []net.IP) ([]byte, error) {
+	nsegs := len(segments) // nsegs: number of segments
+	if nsegs == 0 {
+		return nil, errors.New("EncodeSEG6Encap: No Segment in srh")
+	}
+	b := make([]byte, 12, 12+len(segments)*16)
+	native := NativeEndian()
+	native.PutUint32(b, uint32(mode))
+	b[4] = 0                      // srh.nextHdr (0 when calling netlink)
+	b[5] = uint8(16 * nsegs >> 3) // srh.hdrLen (in 8-octets unit)
+	b[6] = IPV6_SRCRT_TYPE_4      // srh.routingType (assigned by IANA)
+	b[7] = uint8(nsegs - 1)       // srh.segmentsLeft
+	b[8] = uint8(nsegs - 1)       // srh.firstSegment
+	b[9] = 0                      // srh.flags (SR6_FLAG1_HMAC for srh_hmac)
+	// srh.reserved: Defined as "Tag" in draft-ietf-6man-segment-routing-header-07
+	native.PutUint16(b[10:], 0) // srh.reserved
+	for _, netIP := range segments {
+		b = append(b, netIP...) // srh.Segments
+	}
+	return b, nil
+}
+
+func DecodeSEG6Encap(buf []byte) (int, []net.IP, error) {
+	native := NativeEndian()
+	mode := int(native.Uint32(buf))
+	srh := IPv6SrHdr{
+		nextHdr:      buf[4],
+		hdrLen:       buf[5],
+		routingType:  buf[6],
+		segmentsLeft: buf[7],
+		firstSegment: buf[8],
+		flags:        buf[9],
+		reserved:     native.Uint16(buf[10:12]),
+	}
+	buf = buf[12:]
+	if len(buf)%16 != 0 {
+		err := fmt.Errorf("DecodeSEG6Encap: error parsing Segment List (buf len: %d)\n", len(buf))
+		return mode, nil, err
+	}
+	for len(buf) > 0 {
+		srh.Segments = append(srh.Segments, net.IP(buf[:16]))
+		buf = buf[16:]
+	}
+	return mode, srh.Segments, nil
+}
+
+// Helper functions
+func SEG6EncapModeString(mode int) string {
+	switch mode {
+	case SEG6_IPTUN_MODE_INLINE:
+		return "inline"
+	case SEG6_IPTUN_MODE_ENCAP:
+		return "encap"
+	}
+	return "unknown"
+}
--- a/route_linux.go
+++ b/route_linux.go
@@ -125,17 +125,17 @@ func (e *MPLSEncap) Type() int {
 
 func (e *MPLSEncap) Decode(buf []byte) error {
 	if len(buf) < 4 {
-		return fmt.Errorf("Lack of bytes")
+		return fmt.Errorf("lack of bytes")
 	}
 	native := nl.NativeEndian()
 	l := native.Uint16(buf)
 	if len(buf) < int(l) {
-		return fmt.Errorf("Lack of bytes")
+		return fmt.Errorf("lack of bytes")
 	}
 	buf = buf[:l]
 	typ := native.Uint16(buf[2:])
 	if typ != nl.MPLS_IPTUNNEL_DST {
-		return fmt.Errorf("Unknown MPLS Encap Type: %d", typ)
+		return fmt.Errorf("unknown MPLS Encap Type: %d", typ)
 	}
 	e.Labels = nl.DecodeMPLSStack(buf[4:])
 	return nil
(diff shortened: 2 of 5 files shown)


ebiken:
Hi. Let me know if there is any modification I should make for this change. Thanks!

nl/seg6_linux.go
```diff
@@ -0,0 +1,121 @@
...
+	reserved      uint16
+
+	Segments []net.IP
+}
+
+func (s1 *Ipv6SrHdr) Equal(s2 Ipv6SrHdr) bool {
+	if len(s1.Segments) != len(s2.Segments) {
+		return false
+	}
+	if len(s1.Segments) != 0 {
```
aboch: Minor comment: This `if` check is not really needed, if no segments are there, the `for` loop will be a no op.

nl/seg6_linux.go
```diff
@@ -0,0 +1,121 @@
...
+	srh.reserved = 0 // Defined as "Tag" in draft-ietf-6man-segment-routing-header-07
+	b[4] = srh.nexthdr
+	b[5] = srh.hdrlen
+	b[6] = srh.typ
+	b[7] = srh.segments_left
+	b[8] = srh.first_segment
+	b[9] = srh.flags
+	native.PutUint16(b[10:], srh.reserved)
+	for _, netIP := range srh.Segments {
+		b = append(b, netIP.To16()...)
```
aboch: Why `to16()` ? The function is for representing a 4 byte IPv4 address into the 16 bytes IPv4 in IPv6 notation.
Being `srh` of type `Ipv6SrHdr`, I was expecting its segments to be already IPv6 addresses.

nl/seg6_linux.go
```diff
@@ -0,0 +1,121 @@
+package nl
+
+import (
+	"errors"
+	"fmt"
+	"net"
+)
+
+type Ipv6SrHdr struct {
```
aboch: Minor: I think `IP` should be capitalized `IPv6SrHdr`

nl/seg6_linux.go
```diff
@@ -0,0 +1,121 @@
...
+	nexthdr       uint8
+	hdrlen        uint8
+	typ           uint8
```
aboch: Minor: To avoid using the reserved `type` string, can you please rename this field to `routingType` ?
I know it is an internal field, but I am not sure `typ` would directly convey what it represents to somebody reading the code.

route_linux.go
```diff
@@ -185,6 +185,71 @@ func (e *MPLSEncap) Equal(x Encap) bool {
...
+	str := fmt.Sprintf("mode %s segs %d [ %s ]", nl.SEG6EncapModeString(e.Mode),
+		len(e.Srh.Segments), strings.Join(segs, " "))
+	return str
+}
+func (e *SEG6Encap) Equal(x Encap) bool {
+	o, ok := x.(*SEG6Encap)
+	if !ok {
+		return false
+	}
+	if e == nil && o == nil {
```
aboch: Minor: This could be replaced by
```
if e == o {
   return true
}
```

nl/seg6_linux.go
```diff
@@ -0,0 +1,121 @@
+package nl
+
+import (
+	"errors"
+	"fmt"
+	"net"
+)
+
+type Ipv6SrHdr struct {
+	nexthdr       uint8
```
aboch: This project in most parts seems to be following the convention of CamelCase naming.
Please avoid underscore in structure field names and change this one to something like `hdrLen` or `headerLen` what you prefer.

nl/seg6_linux.go
```diff
@@ -0,0 +1,121 @@
...
+	nsegs := len(srh.Segments) // nsegs: number of segments
+	if nsegs == 0 {
+		return nil, errors.New("EncodeSEG6Encap: No Segment in srh")
+	}
+	b := make([]byte, 12, 12+len(srh.Segments)*16)
+	native := NativeEndian()
+	native.PutUint32(b, uint32(mode))
+	srh.nexthdr = 0                     // use 0 when calling netlink
+	srh.hdrlen = uint8(16 * nsegs >> 3) // in 8-octets unit
+	srh.typ = 4                         // Routing Type to be assigned by IANA (suggested value: 4)
```
aboch: Shouldn't we have a constant for this as does the kernel code 
```
#define IPV6_SRCRT_TYPE_4	4	/* Segment Routing with IPv6 */
```

nl/seg6_linux.go
```diff
@@ -0,0 +1,121 @@
...
+		first_segment: buf[8],
+		flags:         buf[9],
+		reserved:      native.Uint16(buf[10:12]),
+	}
+	buf = buf[12:]
+	if len(buf)%16 != 0 {
+		err := fmt.Errorf("DecodeSEG6Encap: Error parsing Segment List (buf len: %d)\n", len(buf))
+		return mode, srh, err
+	}
+	seglist := make([]net.IP, 0, len(buf)/16)
```
aboch: Very minor: We do not need the extra `seglist` slice, we could just replace it with `srh.Segments`.

nl/seg6_linux.go
```diff
@@ -0,0 +1,121 @@
...
+		hdrlen:        buf[5],
+		typ:           buf[6],
+		segments_left: buf[7],
+		first_segment: buf[8],
+		flags:         buf[9],
+		reserved:      native.Uint16(buf[10:12]),
+	}
+	buf = buf[12:]
+	if len(buf)%16 != 0 {
+		err := fmt.Errorf("DecodeSEG6Encap: Error parsing Segment List (buf len: %d)\n", len(buf))
```
aboch: This file and other ones in this project seems to be following the go recommendation where the reported error strings start in lower case. Please adhere to this convention.

route_linux.go
```diff
@@ -185,6 +185,71 @@ func (e *MPLSEncap) Equal(x Encap) bool {
...
+	Mode int
+	Srh  nl.Ipv6SrHdr
+}
+
+func (e *SEG6Encap) Type() int {
+	return nl.LWTUNNEL_ENCAP_SEG6
+}
+func (e *SEG6Encap) Decode(buf []byte) error {
+	if len(buf) < 4 {
+		return fmt.Errorf("Lack of bytes")
```
aboch: This file and other ones in this project seems to be following the go recommendation where the reported error strings start in lower case. Please adhere to this convention here and in the other places.

nl/seg6_linux.go
ebiken: Thanks for the review. I will remove this `if`.

nl/seg6_linux.go
ebiken: Yes. Will change to `b = append(b, netIP...)`.

nl/seg6_linux.go
ebiken: OK. Will change to `IPv6SrHdr`.

nl/seg6_linux.go
ebiken: I followed what's used as temp var in MPLS code.
But yes, `routingType` should be easier to understand as a name of `IPv6SrHdr` member.
I changed ones in `nl/seg6_linux.go` to `routingType`.
I kept ones in `route_linux.go` as is `typ` since they are temp var used very close to where defined by `:=`.

route_linux.go
ebiken: Thank you. Very educational. Will change it.

nl/seg6_linux.go
ebiken: Changed to below.
```
type IPv6SrHdr struct {
    nextHdr       uint8
    hdrLen        uint8
    routingType   uint8
    segmentsLeft  uint8
    firstSegment  uint8
```

nl/seg6_linux.go
ebiken: Defined them in nl/syscall.go
```
// routing header types
const (
    IPV6_SRCRT_STRICT = 0x01 // Deprecated; will be removed
    IPV6_SRCRT_TYPE_0 = 0    // Deprecated; will be removed
    IPV6_SRCRT_TYPE_2 = 2    // IPv6 type 2 Routing Header
    IPV6_SRCRT_TYPE_4 = 4    // Segment Routing with IPv6
)
```
And changed to: `srh.routingType = IPV6_SRCRT_TYPE_4 // Routing Type assigned by IANA`

nl/seg6_linux.go
ebiken: OK. Removed `seglist := make` and modified append line as below:
`srh.Segments = append(srh.Segments, net.IP(buf[:16]))`.

route_linux.go
ebiken: OK. I fixed ones in this PR and from MPLS code which had same error message. (ex: "Lack of bytes")

ebiken:
Applied changes based on feedback. Fixed conflict due to syscall -> unix checkin recently done on master.
Please let me know if I missed anything. Thanks!!

route_linux.go
```diff
@@ -186,6 +186,71 @@ func (e *MPLSEncap) Equal(x Encap) bool {
 	return true
 }
 
+// SEG6 definitions
+type SEG6Encap struct {
+	Mode int
+	Srh  nl.IPv6SrHdr
```
aboch: This is the last thing I am not convinced about.

`Seg6Encap` is `netlink` pkg user interface, lib client will have to provide it when calling `netlink.AddRoute()`.
I am not sure we should have client import `netlink/nl` for setting `Srh`, as it is done now for convenience I guess, where only  its `Segments` field is exported.

I took a look at other `netlink` pkg API, for the most no `netlink/nl` types need to be fed in.

Probably this type could be changed to 
```
type SEG6Encap struct {
	Mode         int
	Segments  []net.IP
}
``` 

and the respective `nl.IPv6SrHdr` could be created internally when interfacing with `netlink/nl` methods.

aboch:
Thank you @ebiken, I have one last comment, otherwise it looks good to me.

route_linux.go
ebiken: I have changed to `Segments  []net.IP` as you suggested.
I removed using `IPv6SrHdr` in `EncodeSEG6Encap` to make code cleaner and short.
I kept using it in `DecodeSEG6Encap` in case one wants to do granular check with `func (s1 *IPv6SrHdr) Equal(s2 IPv6SrHdr) bool` by directly using netlink/nl.
For example, there could be a case that only `segmentsLeft` `firstSegment` are different for two seg6 route entries. However, explicitly configuring those values via iproute2 is not currently supported. Thus, will leave Encap part for future when someone actually wants to do such operation.

ebiken:
Note: I have run the test and confirmed it pass after the change.
```
# go test -run TestSEG6RouteAddDel
PASS
ok      github.com/ebiken/netlink       0.005s
```

aboch:
Thank you @ebiken, looks good to me.

ebiken:
Thank you @aboch for many suggestions.
Do you have write access to the main branch? If so, appreciate if you can merge this.

PR #283, opened 2017-10-10
Allow index specification at link creation time

mcastelino:
Allow the caller to specify the desired link index at link creation.

This is equivalent to
ip link add link eth0 name testmacvtap index 1000 type macvtap
ip link add dummy1 index 1001 type dummy

### DIFF
--- a/link_linux.go
+++ b/link_linux.go
@@ -851,6 +851,10 @@ func (h *Handle) linkModify(link Link, flags int) error {
 		msg.Change |= syscall.IFF_MULTICAST
 		msg.Flags |= syscall.IFF_MULTICAST
 	}
+	if base.Index != 0 {
+		msg.Index = int32(base.Index)
+	}
+
 	req.AddData(msg)
 
 	if base.ParentIndex != 0 {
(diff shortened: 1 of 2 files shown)


link_linux.go
```diff
@@ -851,6 +851,14 @@ func (h *Handle) linkModify(link Link, flags int) error {
 		msg.Change |= syscall.IFF_MULTICAST
 		msg.Flags |= syscall.IFF_MULTICAST
 	}
+	if base.Index != 0 {
+		if flags&syscall.NLM_F_CREATE == 0 {
```
vishvananda: I don't think it should check for NLM_F_CREATE and error here. Specifically this will break BridgeSetMcastSnoop if you attempt to retrieve the link (which sets the index value) and then use the retrieved link to set snoop.

mcastelino: Let me remove that check. I was just being paranoid. I did not realize that there was a case beyond create where it was valid to set the index.

mcastelino:
@vishvananda addressed the review comment. CI has also passed.

PR #285, opened 2017-10-18
added support for Foo-over-UDP netlink calls

skoef:
The Foo-over-UDP patches from Tom Herbert added some netlink calls to the kernel. Here is the support for these calls, similar to `ip fou add` and `ip fou del` as well as a list function based on `FOU_CMD_GET`, to my knowledge currently not (yet) implemented in `iproute2`.

### DIFF
--- a/fou_linux.go
+++ b/fou_linux.go
@@ -1,0 +1,173 @@
+// +build linux
+
+package netlink
+
+import (
+	"encoding/binary"
+	"errors"
+
+	"github.com/vishvananda/netlink/nl"
+	"golang.org/x/sys/unix"
+)
+
+const (
+	FOU_GENL_NAME = "fou"
+)
+
+const (
+	FOU_CMD_UNSPEC uint8 = iota
+	FOU_CMD_ADD
+	FOU_CMD_DEL
+	FOU_CMD_GET
+	FOU_CMD_MAX = FOU_CMD_GET
+)
+
+const (
+	FOU_ATTR_UNSPEC = iota
+	FOU_ATTR_PORT
+	FOU_ATTR_AF
+	FOU_ATTR_IPPROTO
+	FOU_ATTR_TYPE
+	FOU_ATTR_REMCSUM_NOPARTIAL
+	FOU_ATTR_MAX = FOU_ATTR_REMCSUM_NOPARTIAL
+)
+
+const (
+	FOU_ENCAP_UNSPEC = iota
+	FOU_ENCAP_DIRECT
+	FOU_ENCAP_GUE
+	FOU_ENCAP_MAX = FOU_ENCAP_GUE
+)
+
+var fouFamilyId int
+
+func FouFamilyId() (int, error) {
+	if fouFamilyId != 0 {
+		return fouFamilyId, nil
+	}
+
+	fam, err := GenlFamilyGet(FOU_GENL_NAME)
+	if err != nil {
+		return -1, err
+	}
+
+	fouFamilyId = int(fam.ID)
+	return fouFamilyId, nil
+}
+
+func FouAdd(f Fou) error {
+	return pkgHandle.FouAdd(f)
+}
+
+func (h *Handle) FouAdd(f Fou) error {
+	fam_id, err := FouFamilyId()
+	if err != nil {
+		return err
+	}
+
+	// setting ip protocol conflicts with encapsulation type GUE
+	if f.EncapType == FOU_ENCAP_GUE && f.Protocol != 0 {
+		return errors.New("GUE encapsulation doesn't specify an IP protocol")
+	}
+
+	req := h.newNetlinkRequest(fam_id, unix.NLM_F_ACK)
+
+	// int to byte for port
+	bp := make([]byte, 2)
+	binary.BigEndian.PutUint16(bp[0:2], uint16(f.Port))
+
+	attrs := []*nl.RtAttr{
+		nl.NewRtAttr(FOU_ATTR_PORT, bp),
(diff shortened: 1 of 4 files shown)


skoef:
Looking at this error (https://travis-ci.org/vishvananda/netlink/builds/289425912#L541) and some other tests complaining about travis' kernel being too old, I think the same goes for `TestFouAddDel` or `fou` support in general as `fou` was merged in 3.18.

skoef:
@vishvananda how is checking for `TRAVIS_BUILD_DIR` related to kernel 3.13? I'll be happy to add a decent check around `TestFouAddDell` but I'm not sure it makes a lot of sense.

skoef:
I’m not quite happy with this result so far. Am I corect that next to managing Route Attributes there currently is no support for using ‘regular’ netlink attributes (like I’m parsing for FOU in this request)? I would rather implement that and use it for FOU than implement this specifically for FOU.

fou_linux.go
```diff
@@ -0,0 +1,181 @@
...
+	FOU_ENCAP_UNSPEC = iota
+	FOU_ENCAP_DIRECT
+	FOU_ENCAP_GUE
+	FOU_ENCAP_MAX = FOU_ENCAP_GUE
+)
+
+var fouFamilyId int
+
+func FouFamilyId() (int, error) {
+	if fouFamilyId == 0 {
```
aboch: Please make this more readable as
```
func FouFamilyId() (int, error) {
    if fouFamilyId != 0 {
       return fouFamilyId, nil
    }

    fams, err := GenlFamilyList()
    if err != nil {
        return -1, err
    }

    for _, f := range fams {
        if f.Name == FOU_GENL_NAME {
          fouFamilyId = int(f.ID)
          return fouFamilyId, nil
        }
    }

    return -1, fmt.Errorf("could not find genl family %s", FOU_GENL_NAME)
}
```

fou_linux.go
```diff
@@ -0,0 +1,181 @@
...
+)
+
+const (
+	FOU_ENCAP_UNSPEC = iota
+	FOU_ENCAP_DIRECT
+	FOU_ENCAP_GUE
+	FOU_ENCAP_MAX = FOU_ENCAP_GUE
+)
+
+var fouFamilyId int
```
aboch: Wondering if `Handle` would be a good place where to store the generic netlink families ids.

fou_linux.go
```diff
@@ -0,0 +1,181 @@
...
+
+func FouFamilyId() (int, error) {
+	if fouFamilyId == 0 {
+		fams, err := GenlFamilyList()
```
aboch: BTW, any reason for not using `GenlFamilyGet()` ?

fou_linux.go
```diff
@@ -0,0 +1,181 @@
...
+	msgs, err := req.Execute(syscall.NETLINK_GENERIC, 0)
+	if err != nil {
+		return nil, err
+	}
+
+	fous := []Fou{}
+	for _, m := range msgs {
+		if f, err := deserializeFouMsg(m); err != nil {
+			return fous, err
+		} else {
```
aboch: the `else` is not needed

fou_linux.go
```diff
@@ -0,0 +1,181 @@
...
+	}
+
+	req.AddRawData(raw)
+
+	msgs, err := req.Execute(syscall.NETLINK_GENERIC, 0)
+	if err != nil {
+		return nil, err
+	}
+
+	fous := []Fou{}
```
aboch: to avoid reallocation of the slice at each `append`, you can define this as
```
fous := make([]Fou, 0, len(msgs))
```

fou_test.go
```diff
@@ -0,0 +1,103 @@
...
+	msg = []byte{3, 1, 0, 0, 5, 0, 2, 0, 2, 0, 0}
+	if _, err := deserializeFouMsg(msg); err == nil {
+		t.Error("expected attribute body truncated error")
+	} else if err != ErrAttrBodyTruncated {
+		t.Errorf("unexpected error: %s", err.Error())
+	}
+}
+
+func TestFouAddDel(t *testing.T) {
+	tearDown := setUpNetlinkTest(t)
```
aboch: Please add the call to `minKernelRequired(t, M, N)` so that this test only runs on kernel which knows about FOU

aboch:
Please squash the commits into one as they are all about the same feature.

skoef:
@aboch thanks for your suggestions, I've applied them. As you seem invested, what do you think regarding the implementation of parsing regular netlink attributes? It seems foolish to implement it just for decoding FOU attributes while it applies to more non-routing attributes as well.

I'll squash the commits when everything is ok.

fou_linux.go
skoef: Do you mean adding a map of netlink families to the `Handle` struct?

aboch:
I think we can get this PR in as is.
You can come back later to implement generic parsing methods other could use.

Are you sure you can't use `nl.ParseRouteAttr()` ?
I have seen it being used to parse attributes from genl messages.

skoef:
hmm, I think something went wrong rebasing. I merged master in between my development and I see all those commits as well instead of 1 merge commit. Should I squash all these commits into 1?

aboch:
`git fetch <upstream> master` && `git rebase <upstream> master`, fix conflicts, `git rebase --continue`, `git push origin <branch>:<branch> -f`

aboch:
Thanks @skoef . Please `--signoff` your commit.

skoef:
I just did, happy to help!

skoef:
Not sure how this `TestRuleAddDel` test fails since I've signed-off the commit, I didn't touch that code at all.

aboch:
Thanks @skoef , LGTM

PR #286, opened 2017-10-20
replace syscall with golang.org/x/sys/unix

### LINKED ISSUE #49: migrate to golang.org/x/sys
The Go doc for [syscall](https://golang.org/pkg/syscall/) says:

> NOTE: This package is locked down. Code outside the standard Go repository should be migrated to use the corresponding package in the golang.org/x/sys repository. That is also where updates required by new systems or versions should be applied. See https://golang.org/s/go1.4-syscall for more information.

porjo:
Closes #49 

Mostly a search and replace, swapping `syscall` out for `golang.org/x/sys/unix` (where possible)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -3,7 +3,8 @@ DIRS := \
 	nl
 
 DEPS = \
-	github.com/vishvananda/netns
+	github.com/vishvananda/netns \
+	golang.org/x/sys/unix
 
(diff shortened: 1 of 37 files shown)


vishvananda:
Thanks for this. I suppose I should switch over to a real dependency manager like glide at some point.

PR #292, opened 2017-11-07
Allow to skip tests based on min kernel version required

aboch:
- Adds utility to selectively skip tests based on minimum kernel version required.

Example:
```
minRequiredKernel(t, 3, 16)
```
This should be better than what we have now:
```
	if os.Getenv("TRAVIS_BUILD_DIR") != "" {
		t.Skipf("Skipped some tests because travis kernel is to old to support BR_PROXYARP.")
	}
```

- Also, properly use the test `Skip()` method, do not log separately the skip message.  Run the tests with the verbose `-test.v` option which will cause the skip message to be properly printed.

Signed-off-by: Alessandro Boch <aboch@tetrationanalytics.com>

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -19,7 +19,7 @@ $(call goroot,$(DEPS)):
 
 .PHONY: $(call testdirs,$(DIRS))
 $(call testdirs,$(DIRS)):
-	sudo -E go test -test.parallel 4 -timeout 60s -v github.com/vishvananda/netlink/$@
+	go test -test.exec sudo -test.parallel 4 -timeout 60s -test.v github.com/vishvananda/netlink/$@
 
 $(call fmt,$(call testdirs,$(DIRS))):
 	! gofmt -l $(subst fmt-,,$@)/*.go | grep -q .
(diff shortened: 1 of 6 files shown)


aboch:
For some reason, on Travis The SKIP tag and message is not printed next to the test name, while on my 3.10 machine is

```
=== RUN   TestBridgeCreationWithMulticastSnooping
--- SKIP: TestBridgeCreationWithMulticastSnooping (0.00s)
	netlink_test.go:96: Host Kernel (3.10) does not meet test's minimum required version: (4.4)
=== RUN   TestBridgeSetMcastSnoop
--- SKIP: TestBridgeSetMcastSnoop (0.00s)
	netlink_test.go:96: Host Kernel (3.10) does not meet test's minimum required version: (4.4)
=== RUN   TestBridgeCreationWithHelloTime
--- SKIP: TestBridgeCreationWithHelloTime (0.00s)
	netlink_test.go:96: Host Kernel (3.10) does not meet test's minimum required version: (3.18)
```

PR #293, opened 2017-11-09
tuntap: Add multiqueue support

mcastelino:
Add multi queue support to tuntap without breaking legacy users
of tuntap.

### DIFF
--- a/link_linux.go
+++ b/link_linux.go
@@ -21,13 +21,15 @@ const (
 )
 
 const (
-	TUNTAP_MODE_TUN  TuntapMode = unix.IFF_TUN
-	TUNTAP_MODE_TAP  TuntapMode = unix.IFF_TAP
-	TUNTAP_DEFAULTS  TuntapFlag = unix.IFF_TUN_EXCL | unix.IFF_ONE_QUEUE
-	TUNTAP_VNET_HDR  TuntapFlag = unix.IFF_VNET_HDR
-	TUNTAP_TUN_EXCL  TuntapFlag = unix.IFF_TUN_EXCL
-	TUNTAP_NO_PI     TuntapFlag = unix.IFF_NO_PI
-	TUNTAP_ONE_QUEUE TuntapFlag = unix.IFF_ONE_QUEUE
+	TUNTAP_MODE_TUN             TuntapMode = unix.IFF_TUN
+	TUNTAP_MODE_TAP             TuntapMode = unix.IFF_TAP
+	TUNTAP_DEFAULTS             TuntapFlag = unix.IFF_TUN_EXCL | unix.IFF_ONE_QUEUE
+	TUNTAP_VNET_HDR             TuntapFlag = unix.IFF_VNET_HDR
+	TUNTAP_TUN_EXCL             TuntapFlag = unix.IFF_TUN_EXCL
+	TUNTAP_NO_PI                TuntapFlag = unix.IFF_NO_PI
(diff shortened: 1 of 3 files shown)


link_linux.go
```diff
@@ -792,39 +800,76 @@ func (h *Handle) linkModify(link Link, flags int) error {
...
 		// can't set master during create, so set it afterwards
 		if base.MasterIndex != 0 {
 			// TODO: verify MasterIndex is actually a bridge?
-			return h.LinkSetMasterByIndex(link, base.MasterIndex)
+			err := h.LinkSetMasterByIndex(link, base.MasterIndex)
+			if err != nil {
+				_, _, _ = unix.Syscall(unix.SYS_IOCTL, fds[0].Fd(), uintptr(unix.TUNSETPERSIST), 0)
+				cleanupFds(fds)
+			}
+			return err
```
aboch: If we can successfully set the master interface, then we return here, without storing the list of opened fds into `tuntap.Fds`. Is it intentional ?

mcastelino: @aboch that is a bug. I will fix this and update the PR.

link_linux.go
```diff
@@ -792,39 +800,75 @@ func (h *Handle) linkModify(link Link, flags int) error {
...
 
 		// can't set master during create, so set it afterwards
 		if base.MasterIndex != 0 {
 			// TODO: verify MasterIndex is actually a bridge?
-			return h.LinkSetMasterByIndex(link, base.MasterIndex)
+			err := h.LinkSetMasterByIndex(link, base.MasterIndex)
+			if err != nil {
+				_, _, _ = unix.Syscall(unix.SYS_IOCTL, fds[0].Fd(), uintptr(unix.TUNSETPERSIST), 0)
+				cleanupFds(fds)
+			}
```
aboch: Now we will not report this error to the caller. Is this the expected behavior ?

mcastelino: @aboch that should also be fixed now... sorry about that.

link_linux.go
```diff
@@ -792,39 +800,76 @@ func (h *Handle) linkModify(link Link, flags int) error {
...
 		}
+
 		req.Flags |= uint16(tuntap.Mode)
-		copy(req.Name[:15], base.Name)
-		_, _, errno := unix.Syscall(unix.SYS_IOCTL, file.Fd(), uintptr(unix.TUNSETIFF), uintptr(unsafe.Pointer(&req)))
-		if errno != 0 {
-			return fmt.Errorf("Tuntap IOCTL TUNSETIFF failed, errno %v", errno)
+
+		for i := 0; i < queues; i++ {
+			localReq := req
```
aboch: Do we really need this extra variable `localReq` ?

mcastelino: We are sending it down as an unsafe pointer later. So it was just a way to ensure that we started with a clean slate each time around
```
unix.Syscall(unix.SYS_IOCTL, file.Fd(), uintptr(unix.TUNSETIFF), uintptr(unsafe.Pointer(&localReq)))
```

aboch: I thought kernel would just read from that structure, but I see it does actually write to it in case of error
`copy_to_user(argp, &ifr, ifreq_len))` in `__tun_chr_ioctl()`.

Given `ifReq` member types, `localReq := req` will indeed create a deep copy of `req`.

aboch:
LGTM

PR #296, opened 2017-11-21
Allow Tuntap non-persist, allow empty tuntap name

rschmied:
- added conditional build in fou.go, did not build on macOS (cosmetic)
- made tuntap creation to allow non-persistent tuntaps.
- allowed passing an empty interface name for tuntap creation. In this case, the OS will choose an appropriate name and copies it back into the linkattr.Name.

This patch also ensures that the link name is always identical to what the kernel returns. Not only in case of an empty name, but also when using name templates. E.g. when the provided name is `"tap%d"`, the kernel replaces `%d` with the next available number.

Added `Persist` to the Tuntap struct. Now, since this is `false` by default, it *might* be better for backward compatibility to call it `NoPersist` and explicitly set it to `true` if the Tuntap should be created as non-persistent. Thoughts?

### DIFF
--- a/link_linux.go
+++ b/link_linux.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"net"
 	"os"
+	"strings"
 	"syscall"
 	"unsafe"
 
@@ -941,7 +942,7 @@ func LinkAdd(link Link) error {
 }
 
 // LinkAdd adds a new link device. The type and features of the device
-// are taken fromt the parameters in the link object.
+// are taken from the parameters in the link object.
 // Equivalent to: `ip link add $link`
 func (h *Handle) LinkAdd(link Link) error {
 	return h.linkModify(link, unix.NLM_F_CREATE|unix.NLM_F_EXCL|unix.NLM_F_ACK)
@@ -951,14 +952,16 @@ func (h *Handle) linkModify(link Link, flags int) error {
 	// TODO: support extra data for macvlan
 	base := link.Attrs()
 
-	if base.Name == "" {
+	// if tuntap, then the name can be empty, OS will provide a name
+	tuntap, isTuntap := link.(*Tuntap)
+
+	if base.Name == "" && !isTuntap {
 		return fmt.Errorf("LinkAttrs.Name cannot be empty!")
 	}
 
-	if tuntap, ok := link.(*Tuntap); ok {
+	if isTuntap {
 		// TODO: support user
 		// TODO: support group
-		// TODO: support non- persistent
 		if tuntap.Mode < unix.IFF_TUN || tuntap.Mode > unix.IFF_TAP {
 			return fmt.Errorf("Tuntap.Mode %v unknown!", tuntap.Mode)
 		}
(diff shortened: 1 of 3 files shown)


rschmied:
realized there's more required to make this work. closing for the moment and potentially re-opening later.

rschmied:
I'm re-opening this again. I run this for quite a while now without any issues seen. Would this be OK to merge?

jellonek:
This looks great, but it changes defaults from setting tap persistence always, to setting it only when a `Persist` parameter value will be true (default for it is false).
That can break existing software.
Maybe better would be to change it to `NonPersist` bool flag?

rschmied:
ha! I did exactly what you proposed and after pushing to my branch read your comment :D  CI should (did!) not show any error now...

link_linux.go
```diff
@@ -1001,12 +1004,20 @@ func (h *Handle) linkModify(link Link, flags int) error {
 				cleanupFds(fds)
 				return fmt.Errorf("Tuntap IOCTL TUNSETIFF failed [%d], errno %v", i, errno)
 			}
+			// 1) we only care for the name of the first tap in the multi queue set
+			// 2) if the original name was empty, the localReq has now the actual name
+			if i == 0 {
```
fcrisciani: is it worth adding also: `base.Name == ""`?

rschmied: imo, if the creation of the interface is successful, then the localReq struct always has the final say on what the name actually is, regardless what was passed into the syscall.

link_linux.go
```diff
@@ -951,14 +952,16 @@ func (h *Handle) linkModify(link Link, flags int) error {
...
 
-	if base.Name == "" {
+	// if tuntap, then the name can be empty, OS will provide a name
+	_, isTuntap := link.(*Tuntap)
+
+	if base.Name == "" && !isTuntap {
 		return fmt.Errorf("LinkAttrs.Name cannot be empty!")
 	}
 
 	if tuntap, ok := link.(*Tuntap); ok {
```
rschmied: this could be optimized by assigning the tuntap in line 956 already and having the condition in line 958 read 
```
if isTuntap {
}
```

link_linux.go
jellonek: TBH i think that we were doing that up to this change incorrectly because we should always read name returned by kernel, not only in case of empty name, but also in case of using name templates.
I did few years ago similar code in https://github.com/rkt/rkt/blob/master/networking/tuntap/tuntap.go - where you can see `tap%d` as name, where kernel in `%d` puts a next available number.

Without this commit we have simply broken handling not only empty names, but also name templates.
So this commit fixes both things in the same time, even if that was not planned by its autor ;)
(still it's worth to note that in e.g. PR description)

fcrisciani: @rschmied more than in the PR description we should add a note in the method itself. PR description are anyway of no use during development.
Would be great if you can add a comment on the method itself.

link_linux.go
fcrisciani: don't see why not

fcrisciani: you will need to change line 956 to get the object itself

fcrisciani:
@rschmied can you squash everything in a single commit, then I will go ahead an merge

rschmied:
OK, I'm guess my Git foo is not strong enough. I *guess* I need to hard-reset my branch to my earliest commit from November last year and apply my changes in one go, commit to remote and merge in latest changes from upstream remote. Doing `git rebase -i C2A3DE3` where `C2A3DE3` is the commit ID from November 14th does not really produce what I'm expecting.  As the log after squashing looks weird (it creates two change sets where one has added a file that I didn't touch). Appreciate any help with how I should proceed with this.

jellonek:
Create new branch from master, try to cherry pick to it your commits, then `rebase -i origin/master` and select squash for them.

fcrisciani:
@rschmied following the steps without having to change branches etc
```
git rebase -i HEAD~7
then in the interactive mode change all the picks to squash a part from the first one.

example:
$ git rebase -i HEAD~4

pick 01d1124 Adding license
pick 6340aaa Moving license into its own file
pick ebfd367 Jekyll has become self-aware.
pick 30e0ccb Changed the tagline in the binary, too.

has to become:
pick 01d1124 Adding license
squash 6340aaa Moving license into its own file
squash ebfd367 Jekyll has become self-aware.
squash 30e0ccb Changed the tagline in the binary, too.

save and exit the editor and the next screen is to adjust the commit message.
```

After that, check with git log that there is only one commit and that is signed, then do the git push but this time with -f to realign the upstream tree

fcrisciani:
Now looks good!

PR #308, opened 2017-12-15
Eliminate cgo from netlink.

hugelgupf:
I have some usages of this library where I'd like to eliminate cgo altogether.

As far as I can tell, `loadSimpleBpf` is only used in tests; and this was a simple enough conversion. No need for cgo here :)

### DIFF
--- a/bpf_linux.go
+++ b/bpf_linux.go
@@ -1,49 +1,12 @@
 package netlink
 
-/*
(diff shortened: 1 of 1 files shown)


aboch:
LGTM

hugelgupf:
Thanks!

PR #309, opened 2017-12-17
Don't overwrite the XDP file descriptor with flags

CtrlZvi:


aboch:
Thank you for fixing this.
LGTM

PR #311, opened 2017-12-27
Add Matchall filter

sargun:


aboch:
Thanks.
Can you please also add the support where the filters are read from kernel:
```
func (h *Handle) FilterList(link Link, parent uint32) ([]Filter, error) {
```
After that, if possible, please add a test in `filter_test.go` to see whether you can retrieve a match_all filter after you programmed it.

sargun:
@aboch It appears updated.

filter_test.go
```diff
@@ -168,13 +168,13 @@ func TestAdvancedFilterAddDel(t *testing.T) {
 	}
 
 	u32SelKeys := []TcU32Key{
-		TcU32Key{
+		{
```
sargun: These changes were made by gofmt, I can split it out into another PR, or a different commit

filter_test.go
```diff
@@ -584,7 +582,14 @@ func TestFilterClsActBpfAddDel(t *testing.T) {
 	if q, ok := qdiscs[0].(*GenericQdisc); !ok || q.Type() != "clsact" {
 		t.Fatal("qdisc is the wrong type")
 	}
+	return qdiscs[0], link
+}
+
+func TestFilterClsActBpfAddDel(t *testing.T) {
+	tearDown := setUpNetlinkTest(t)
```
aboch: I no longer have a 3.10 kernel VM at hand to check this, but I am assuming this test would fail on older kernel.

Can you please add the call
```
minKernelRequired(t, x, y)
```
at the beginning of this test function, and move the same call at line 572 to the top of the new test function.

Thanks

sargun:
@aboch updated.

filter_test.go
```diff
@@ -546,13 +546,11 @@ func TestFilterU32BpfAddDel(t *testing.T) {
 	}
 }
 
-func TestFilterClsActBpfAddDel(t *testing.T) {
-	tearDown := setUpNetlinkTest(t)
-	defer tearDown()
-	if err := LinkAdd(&Ifb{LinkAttrs{Name: "foo"}}); err != nil {
+func setupLinkForTestWithQdisc(t *testing.T, linkName string) (Qdisc, Link) {
```
aboch: :+1:

aboch:
Thanks @sargun

PR #313, opened 2018-01-02
Fix reference to BPF instructions

### LINKED ISSUE #312: BPF Load Program failing
It looks like on Kernel 4.13, it's failing to load the BPF program with the following error from the verifier:
`jump out of range from insn 1 to 2`

sargun:
Fixes #312

### DIFF
--- a/bpf_linux.go
+++ b/bpf_linux.go
@@ -39,7 +39,7 @@ func loadSimpleBpf(progType BpfProgType, ret uint32) (int, error) {
 	attr := BPFAttr{
 		ProgType: uint32(progType),
 		InsnCnt:  uint32(len(insns)),
-		Insns:    uintptr(unsafe.Pointer(&insns)),
+		Insns:    uintptr(unsafe.Pointer(&insns[0])),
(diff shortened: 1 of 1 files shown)


bpf_linux.go
```diff
@@ -39,7 +39,7 @@ func loadSimpleBpf(progType BpfProgType, ret uint32) (int, error) {
 	attr := BPFAttr{
 		ProgType: uint32(progType),
 		InsnCnt:  uint32(len(insns)),
-		Insns:    uintptr(unsafe.Pointer(&insns)),
+		Insns:    uintptr(unsafe.Pointer(&insns[0])),
```
fcrisciani: I think would be cleaner to remove the second instruction (line 36), the result should be the same

sargun: See the bug.

fcrisciani:
LGTM, fixes: https://github.com/vishvananda/netlink/issues/312

aboch:
Thanks @sargun and @fcrisciani

PR #315, opened 2018-01-17
Send IFA_CACHEINFO when setting up addresses

Natolumin:
This patch implements setting the cacheinfo attributes in addrHandle, to enable setting the valid_lft and preferred_lft values on addresses inserted. Currently, these values are ignored.

There is an interface issue because the default values for these fields in go are valid values for the netlink fields, but they are not very good default values. I have implemented what I think is a reasonable interface there, but I don't believe it is especially great. Any feedback on a better way to handle these parameters would be welcome.

This also removes the IFA_BROADCAST and IFA_LABEL fields from messages for non-ipv4 addresses, since these attributes are ignored for other kinds of addresses.

### DIFF
--- a/addr_linux.go
+++ b/addr_linux.go
@@ -118,6 +118,17 @@ func (h *Handle) addrHandle(link Link, addr *Addr, req *nl.NetlinkRequest) error
 		}
 	}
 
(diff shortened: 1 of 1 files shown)


aboch:
LGTM

PR #318, opened 2018-01-20
Add List option to RouteSubscribeWithOptions, AddrSubscribeWithOptions, and LinkSubscribeWithOptions.

eriknordmark:
If an application wants to track the set of kernel routes (or addresses or links) it needs to both determine the initial content and subscribe to changes. However, there are inherent race conditions in using RouteList + RouteSubscribe. If RouteList is invoked before RouteSubscribe then obviously some changes can be lost in between. But more subtly, in the other other the application can see some adds and deletes from the subscribe before the RouteList results, which makes it hard to track the actual content of the kernel.

This patch adds a List boolean option which makes RouteSubscribeWithOptions (and same for addr and links) request the list and the subscription and feed the initial table as "adds" to the application, which makes it a lot easier to track the kernel table.

### DIFF
--- a/link_linux.go
+++ b/link_linux.go
@@ -1476,13 +1476,13 @@ type LinkUpdate struct {
 // LinkSubscribe takes a chan down which notifications will be sent
 // when links change.  Close the 'done' chan to stop subscription.
 func LinkSubscribe(ch chan<- LinkUpdate, done <-chan struct{}) error {
-	return linkSubscribeAt(netns.None(), netns.None(), ch, done, nil)
+	return linkSubscribeAt(netns.None(), netns.None(), ch, done, nil, false)
 }
 
 // LinkSubscribeAt works like LinkSubscribe plus it allows the caller
 // to choose the network namespace in which to subscribe (ns).
 func LinkSubscribeAt(ns netns.NsHandle, ch chan<- LinkUpdate, done <-chan struct{}) error {
-	return linkSubscribeAt(ns, netns.None(), ch, done, nil)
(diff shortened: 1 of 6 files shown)


link_linux.go
```diff
@@ -1514,6 +1515,16 @@ func linkSubscribeAt(newNs, curNs netns.NsHandle, ch chan<- LinkUpdate, done <-c
 			s.Close()
 		}()
 	}
+	if list {
```
aboch: Ideally, shouldn't the dump request be performed after spawning the receive go routine ?

link_linux.go
```diff
@@ -1476,20 +1476,21 @@ type LinkUpdate struct {
...
-	return linkSubscribeAt(ns, netns.None(), ch, done, nil)
+	return linkSubscribeAt(ns, netns.None(), ch, done, nil, false)
 }
 
 // LinkSubscribeOptions contains a set of options to use with
 // LinkSubscribeWithOptions.
 type LinkSubscribeOptions struct {
 	Namespace     *netns.NsHandle
 	ErrorCallback func(error)
+	List	      bool
```
aboch: Should we call this `ListExisting` ? I think it would be more self-explanatory.

aboch:
Please fix the gofmt issues, so we can get the UT running.

aboch:
Also please add a test function for this.

aboch:
Thanks. Please squash the commits in one.

aboch:
LGTM

PR #321, opened 2018-01-31
Add Fq and Fq_Codel Qdisc support

sargun:


qdisc.go
```diff
@@ -230,3 +230,63 @@ func (qdisc *GenericQdisc) Attrs() *QdiscAttrs {
...
+
+func (qdisc *Fq) Attrs() *QdiscAttrs {
+	return &qdisc.QdiscAttrs
+}
+
+func (qdisc *Fq) Type() string {
+	return "fq"
+}
+
+// Fq is a classless packet scheduler meant to be mostly used for locally generated traffic.
```
fcrisciani: This comment is the same as the above

sargun: @fcrisciani Updated.

sargun:
@fcrisciani updated.

PR #327, opened 2018-02-09
Add IFF_RUNNING flag to be RFC2863 compliant

jjastrze-ovh:
Regarding to:
[https://www.kernel.org/doc/Documentation/networking/operstates.txt](https://www.kernel.org/doc/Documentation/networking/operstates.txt) :

if_flags & IFF_RUNNING:
 Interface is in RFC2863 operational state UP or UNKNOWN. This is for
 backward compatibility, routing daemons, dhcp clients can use this
 flag to determine whether they should use the interface.

In my opinion we should be able to determine operational state using netlink.Link attributes

### DIFF
--- a/link_linux.go
+++ b/link_linux.go
@@ -1838,6 +1838,9 @@ func linkFlags(rawFlags uint32) net.Flags {
 	if rawFlags&unix.IFF_MULTICAST != 0 {
 		f |= net.FlagMulticast
 	}
+	if rawFlags&syscall.IFF_RUNNING != 0 {
+		f |= syscall.IFF_RUNNING
+	}
 	return f
 }
 


aboch:
>In my opinion we should be able to determine operational state using netlink.Link attributes

Agree. And you can already do it now:

```
link.Attrs().RawFlags&unix.IFF_RUNNING == unix.IFF_RUNNING
```

jjastrze-ovh:
If this is okey for @aboch -> can be merged?

aboch:
The fact is that `LinkAttrs.Flags` is of type `net.Flags`.  And this change will add `unix` specific values into it. From the `net` [pkg](https://golang.org/pkg/net/#Flags), only an handful of values are supported. (I believe these are the common ones across OSes go supports?)

jjastrze-ovh:
Closing because of "And this change will add unix specific values into it"

PR #330, opened 2018-02-13
Use IFF_MULTI_QUEUE from x/sys/unix to define TUNTAP_MULTI_QUEUE

tklauser:
Use the constant defined in the golang.org/x/sys/unix package instead of
locally open-coding it.


lzap:
Does not compile on Fedora 30 golang which is go version go1.12.5 linux/amd64 for the record.

tklauser:
@lzap there's probably an outdated version of `golang.org/x/sys/unix` in your `$GOPATH`. Please update it to get the constants.

FWIW, I also sent #456 which adds a `go.mod` file which pins the required version of `golang.org/x/sys/unix`.

PR #332, opened 2018-02-23
Use IFLA_* constants from x/sys/unix

tklauser:
The IFLA_* constants in in x/sys/unix were updated to Linux 4.15 in
golang/sys@88d2dcc51026, so use these instead of locally duplicating
them.


fcrisciani:
LGTM

PR #335, opened 2018-03-16
Add class stats

tfukushima:
This series of patches adds statistics fields that were not parsed previously to classes .

The first patch adds the parsing process of the rtattrs for the stats, `TCA_STATS` and `TCA_STATS2`, adding new structs for the statistics, `ClassStatistics`. `TCA_STATS` and `TCA_STATS2` have different structures internally but they are converted into `ClassStatistics` in the end. Brief sanity checks for the stats fields are added in `class_test.go` as well.

In the second patch, new legitimate tests that actually validate the stats for the real packets are added. It requires github.com/google/gopacket for generating and sending arbitrary packets in the tests.

### DIFF
--- a/class.go
+++ b/class.go
@@ -9,14 +9,63 @@ type Class interface {
 	Type() string
 }
 
+// Generic networking statistics for netlink users.
+// This file contains "gnet_" prefixed structs and relevant functions.
+// See Documentation/networking/getn_stats.txt in Linux source code for more details.
+
+// Ref: struct gnet_stats_basic { ... }
+type GnetStatsBasic struct {
+	Bytes   uint64 // number of seen bytes
+	Packets uint32 // number of seen packets
+}
+
+// Ref: struct gnet_stats_rate_est { ... }
+type GnetStatsRateEst struct {
+	Bps uint32 // current byte rate
+	Pps uint32 // current packet rate
+}
+
+// Ref: struct gnet_stats_rate_est64 { ... }
+type GnetStatsRateEst64 struct {
+	Bps uint64 // current byte rate
+	Pps uint64 // current packet rate
+}
+
+// Ref: struct gnet_stats_queue { ... }
+type GnetStatsQueue struct {
+	Qlen       uint32 // queue length
+	Backlog    uint32 // backlog size of queue
+	Drops      uint32 // number of dropped packets
+	Requeues   uint32 // number of requues
+	Overlimits uint32 // number of enqueues over the limit
+}
+
+// Statistics representaion based on generic networking statisticsfor netlink.
+// See Documentation/networking/gen_stats.txt in Linux source code for more details.
+type ClassStatistics struct {
+	Basic   *GnetStatsBasic
+	Queue   *GnetStatsQueue
+	RateEst *GnetStatsRateEst
+}
+
+// Construct a ClassStatistics struct which fields are all initialized by 0.
+func NewClassStatistics() *ClassStatistics {
+	return &ClassStatistics{
+		Basic:   &GnetStatsBasic{},
+		Queue:   &GnetStatsQueue{},
+		RateEst: &GnetStatsRateEst{},
+	}
+}
+
 // ClassAttrs represents a netlink class. A filter is associated with a link,
 // has a handle and a parent. The root filter of a device should have a
 // parent == HANDLE_ROOT.
 type ClassAttrs struct {
-	LinkIndex int
-	Handle    uint32
-	Parent    uint32
-	Leaf      uint32
+	LinkIndex  int
+	Handle     uint32
+	Parent     uint32
+	Leaf       uint32
+	Statistics *ClassStatistics
 }
 
 func (q ClassAttrs) String() string {
--- a/class_linux.go
+++ b/class_linux.go
@@ -1,12 +1,32 @@
 package netlink
 
 import (
+	"bytes"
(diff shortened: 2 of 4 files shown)


tfukushima:
```
=== RUN   TestFilterAndClassStats
--- FAIL: TestFilterAndClassStats (0.06s)
	filter_test.go:806: Could not send the packet: send: No buffer space available
```

The added test is failing due to "send: No buffer space available" but it is working in my local env:

```
$ sudo /usr/local/go/bin/go test -v -run TestFilterAndClassStats
=== RUN   TestFilterAndClassStats
--- PASS: TestFilterAndClassStats (0.01s)
PASS
ok      github.com/vishvananda/netlink  0.141s
```

I'm taking a look at what is going on. It looks like a TravisCI specific issue for me. Do you have any idea, @vishvananda?

class.go
```diff
@@ -2,21 +2,57 @@ package netlink
 
 import (
 	"fmt"
+
+	"github.com/vishvananda/netlink/nl"
```
aboch: I might be wrong, I think we usually do not import `netlink/nl` in the API go files.
For how annoying it may seem, you may need to redefine those `nl` pkg gnet stats here.

class_linux.go
```diff
@@ -253,3 +269,52 @@ func parseHtbClassData(class Class, data []syscall.NetlinkRouteAttr) (bool, erro
...
+	}
+	native := nl.NativeEndian()
+	stats := NewClassStatistics()
+	for _, datum := range rtAttrs {
+		switch datum.Attr.Type {
+		case nl.TCA_STATS_BASIC:
+			stats.Basic.Bytes = native.Uint64(datum.Value[:8])
+			stats.Basic.Packets = native.Uint32(datum.Value[8:12])
+		case nl.TCA_STATS_QUEUE:
+			stats.Queue.Qlen = native.Uint32(datum.Value[:8])
```
aboch: Why 8 bytes for a Uint32 ? Can you double check.

class_linux.go
```diff
@@ -253,3 +269,52 @@ func parseHtbClassData(class Class, data []syscall.NetlinkRouteAttr) (bool, erro
...
+			stats.Queue.Backlog = native.Uint32(datum.Value[8:16])
+			stats.Queue.Drops = native.Uint32(datum.Value[16:24])
+			stats.Queue.Requeues = native.Uint32(datum.Value[24:32])
+			stats.Queue.Overlimits = native.Uint32(datum.Value[32:40])
+		case nl.TCA_STATS_RATE_EST:
+			stats.RateEst.Bps = native.Uint32(datum.Value[:8])
```
aboch: Same here. Please double check.

vishvananda:
I suspect the container doesn't have the permissions to send packet data using pcap. Might have to skip the test (or at least that part of it) in travis.

class_linux.go
tfukushima: You're totally right. I changed the way of parsing the stats data to use the corresponding structs.

class_linux.go
tfukushima: Ditto.

class.go
```diff
@@ -9,14 +9,79 @@ type Class interface {
 	Type() string
 }
 
+// Internal tc_stats representation in Go struct.
```
aboch: Thanks. But please move this into `class_linux.go`, it belongs there.

aboch:
ping @tfukushima

tfukushima:
Hi @aboch, I am working on something else and I'll be back soon. In the meantime, I push the updated code by which I resolved your comment.

I tried the packet injection with `syscall.Sendto` and `syscall.BindToDevice` but I ran into another issue that they don't seem to reflect the netns created in `setUpNetlinkTest`. I am not sure how I can avoid the issues without removing the test. Besides I would like to ask what is a good way to detect if the tests are running in a container and disable only specific tests.

class_linux.go
```diff
@@ -197,9 +217,10 @@ func (h *Handle) ClassList(link Link, parent uint32) ([]Class, error) {
 		}
 
 		base := ClassAttrs{
-			LinkIndex: int(msg.Ifindex),
-			Handle:    msg.Handle,
-			Parent:    msg.Parent,
+			LinkIndex:  int(msg.Ifindex),
+			Handle:     msg.Handle,
+			Parent:     msg.Parent,
+			Statistics: NewClassStatistics(),
```
aboch: Given `tcstats.Statistics` is a pointer, I do not see a need to initialize an empty stats structure.
Caller will anyway check for non `nil` before accessing a pointer parameter.

later on you can just
```
base.Statistics, err := parseTcStats(attr.Value)
if err != nil {
    return nil, err
}
```

If your idea is to always return a stats structure, maybe empty, then it should not be a pointer.
I prefer current solution as a pointer (nil, in case no stats), up to you.

aboch:
Thanks @tfukushima .
Yes, please drop the non-working test code changes, until you sort it out.
Just push one commit with the code changes and remove the new deps pulled in travis (if not needed).
Got one more comment, then we are good to go.

class_linux.go
tfukushima: That makes sense. Fixed.

tfukushima:
@aboch I updated the first patch and removed the second patch. Now CI looks happy.

aboch:
Thanks

LGTM

PR #337, opened 2018-03-16
Support zeroing counters when listing conntrack table

lookuptable:


lookuptable:
@aboch @fcrisciani 

This PR technically breaks the API (adding a new parameter to an exported function). If API stability is important, I can revise to introduce a new function (e.g., `ConntrackTableListAndReset`) and leave `ConntrackTableList`  untouched.

aboch:
@lookuptable We can get away with it by defining a `ConntrackOption` type which you will use to pass conntrack specific options (hopefully there are more than just -z). Then you modify that API by adding at the end of the parameter list a variadic list of setter functions `func(ConntrackOption)` parameters.

The setter functions will be executed by the API implementation.

aboch:
Thank you @lookuptable .

Also,the library should provide the supported option setters and not expose the option setter prototype, otherwise client can set any parameter.
We should define a type for the option setter, then expose an setter function which construct that option setter:

Something like:
```
type conntrackParameters struct { // note: not exposed
	// Atomically zero counters after reading them.
	zeroCounters bool
}

type ConntrackOption func(*conntrackParameters)   // note: conntrackParameters not exported

// Atomically zero counters after reading them.
func ZeroCounters() ConntrackOption {
   return func(p *conntrackParameters) { p.zeroCounter = true; } // note: zeroCounters not exposed
}

func ConntrackTableList(table ConntrackTableType, family InetFamily, options ...ConntrackOption)) ([]*ConntrackFlow, error) {
[...]
```
Now client will invoke

```
ConntrackTableList(t, f, netlink.ZeroCounters);
```

You can look at libnetwork  [NetworkOption](https://github.com/docker/libnetwork/blob/bump_17.09/network.go#L712) as an example

aboch:
ping @lookuptable

lookuptable:
@aboch sorry for the delay. I'll have the PR updated in the next few days.

vishvananda:
@lookuptable any update on this?

aboch:
Closing for now. Feel free to reopen when you (or any other volunteer) have time to get back to this.

PR #338, opened 2018-03-21
Add support for setting InfininBand Node and Port GUID of a VF

paravmellanox:
Add support for setting InfiniBand Node and Port GUID address
configuration of a VF when InfiniBand HCA are used in SR-IOV mode.

### DIFF
--- a/link_linux.go
+++ b/link_linux.go
@@ -518,6 +518,50 @@ func (h *Handle) LinkSetVfTrust(link Link, vf int, state bool) error {
 	return err
 }
 
+// LinkSetVfNodeGUID sets the node GUID of a vf for the link.
+// Equivalent to: `ip link set dev $link vf $vf node_guid $nodeguid`
+func LinkSetVfNodeGUID(link Link, vf int, nodeguid net.HardwareAddr) error {
+	return pkgHandle.LinkSetVfGUID(link, vf, nodeguid, nl.IFLA_VF_IB_NODE_GUID)
+}
+
+// LinkSetVfPortGUID sets the port GUID of a vf for the link.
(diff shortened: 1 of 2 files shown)


paravmellanox:
Hi @vishvananda, Can you please review this PR? I like to use this APIs in a cni plugin once they are available.

aboch:
@paravmellanox Please add the test code for the new serialization/deserialization methods in https://github.com/vishvananda/netlink/blob/v1.0.0/nl/link_linux_test.go

aboch:
Also, please add a simple test code where you add/get/delete this link type in https://github.com/vishvananda/netlink/blob/v1.0.0/link_test.go which runs only on kernel versions that support infiniband (use `minKernelRequired`).

Thanks

paravmellanox:
@aboch, infiniband links are created by the ib_ipoib.ko driver when there is underlying InfiniBand hardware present. There is no software implementation available at this point.
I am not sure if I can add test for add/del command.
Get command can be possibly done but Get command is no different than other Get commands.

Also testing VF commands requires Supported hardware. Current link_test.go doesn't have VF related test case infrastructure in place.
How about having those tests added in separate PR?
I am happy to write those VF tests, just don't know yet how to do it. Enabling VFs is not through netlink commands. (done through writing in sysfs entry). So currently I have my test programs to validate this commands in our lab where such hardware is provisioned.

aboch:
@paravmellanox Thanks for the explanation.

> How about having those tests added in separate PR?

That sounds good to me. Please be aware that for the netlink/nl pkg serialize/deserialize test code it is just a basic ut.

LGTM

PR #341, opened 2018-04-18
Enhance u32 filters

tfukushima:
This series of patches introduces `Divisor` and `Hash` attributes to `U32` struct.

* With `Divisor` specified, a u32 hash table can be created
* A filter can be created and associated with the filter that has the hash table ID specified with `Hash` attribute

Although these changes enable users to construct the more flexible filter configurations, I'd like to note a hash table cannot be listed or deleted with the current interfaces, `FilterList` and `FilterDel`. We can create a hash table but we *cannot* list and delete one either.

I just added a capability to create a hash table with this PR. If we want to make sure a hash table can be treated as the legitimate entity like it currently is in `tc`, the overhaul for the filter interfaces will be required in the following patches.

See the following links to know the details of the u32 hash table.
- https://git.kernel.org/pub/scm/network/iproute2/iproute2.git/tree/README.iproute2+tc#n92
- https://www.systutorials.com/docs/linux/man/8-tc-u32/

### DIFF
--- a/filter.go
+++ b/filter.go
@@ -212,6 +212,7 @@ type TcU32Key struct {
 type U32 struct {
 	FilterAttrs
 	ClassId    uint32
+	Divisor    uint32 // Divisor MUST be power of 2.
 	RedirIndex int
 	Sel        *TcU32Sel
 	Actions    []Action
--- a/filter.go
+++ b/filter.go
@@ -213,6 +213,7 @@ type U32 struct {
 	FilterAttrs
 	ClassId    uint32
 	Divisor    uint32 // Divisor MUST be power of 2.
+	Hash       uint32
 	RedirIndex int
 	Sel        *TcU32Sel
 	Actions    []Action
--- a/filter_linux.go
+++ b/filter_linux.go
@@ -178,6 +178,9 @@ func (h *Handle) FilterAdd(filter Filter) error {
 			}
 			nl.NewRtAttrChild(options, nl.TCA_U32_DIVISOR, nl.Uint32Attr(filter.Divisor))
 		}
+		if filter.Hash != 0 {
+			nl.NewRtAttrChild(options, nl.TCA_U32_HASH, nl.Uint32Attr(filter.Hash))
+		}
 		actionsAttr := nl.NewRtAttrChild(options, nl.TCA_U32_ACT, nil)
 		// backwards compatibility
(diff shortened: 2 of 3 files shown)


filter.go
```diff
@@ -212,6 +212,7 @@ type TcU32Key struct {
 type U32 struct {
 	FilterAttrs
 	ClassId    uint32
+	Divisor    uint32 // Divisor MUT be a power of 2.
```
aboch: `// Must be power of 2`

filter_linux.go
```diff
@@ -172,6 +172,14 @@ func (h *Handle) FilterAdd(filter Filter) error {
 		if filter.ClassId != 0 {
 			nl.NewRtAttrChild(options, nl.TCA_U32_CLASSID, nl.Uint32Attr(filter.ClassId))
 		}
+		if filter.Divisor != 0 {
+			// Validate if the divisor is a power of two.
```
aboch: Pls remove this comment, the error message is explains it all

filter_linux.go
```diff
@@ -172,6 +172,14 @@ func (h *Handle) FilterAdd(filter Filter) error {
...
+			if (filter.Divisor-1)&filter.Divisor != 0 {
+				return fmt.Errorf("Illegal divisor, which must be a power of two.")
```
aboch: Please follow the convention of error string starting with lower case
`"illegal divisor: %d. Must be a power of 2.", filter.Divisor`

filter_linux.go
```diff
@@ -172,6 +172,14 @@ func (h *Handle) FilterAdd(filter Filter) error {
...
+			} else {
```
aboch: Please drop this `else`

aboch:
Thanks @tfukushima only few minor comments.

filter.go
tfukushima: Fixed.

filter_linux.go
tfukushima: Fixed.

filter_linux.go
tfukushima: Fixed.

filter_linux.go
tfukushima: Fixed.

aboch:
Thanks @tfukushima LGTM

PR #343, opened 2018-04-23
Add RDMA netlink socket for RDMA device information

paravmellanox:
This patch adds very basic support for getting information about RDMA
networking device; starting with device index, name, firmware version,
node GUID and system image GUID.
This is done through RDMA netlink socket.

RDMA devices are some what similar to Ethernet devices.
However there are few major differences between them.
RDMA devices usually have one or two ports, unlike Ethernet devices.
Each port has its own attributes, state and network addresses which are
different than Ethernet devices (Link and LinkAttrs). They almost don't
overlap with Link and LinkAttrs.

Therefore it doesn't derive Link and LinkAttrs structure; instead they
are represented using RdmaLink and RdmaLinkAttrs.

RdmaLink represents a RDMA device containing its attributes.
All Rdma device communication occurs through rdma subsystem's netlink
socket.

Signed-off-by: Parav Pandit parav@mellanox.com


rdma_link_linux.go
```diff
@@ -0,0 +1,109 @@
...
+			return nil, err
+		}
+		if link.Attrs.Name == name {
+			return link, nil
+		}
+	}
+	return nil, fmt.Errorf("Rdma device %v not found", name)
+}
+
+// LinkByName finds a link by name and returns a pointer to the object.
```
aboch: Given this is exported, pls fix the function name in the comment. Or just remove the comment, I don't think it is actually needed.

rdma_link_test.go
```diff
@@ -0,0 +1,38 @@
...
+	"strings"
+	"testing"
+)
+
+func setupRdmaKModule(t *testing.T, name string) {
+	file, err := ioutil.ReadFile("/proc/modules")
+	if err != nil {
+		t.Fatal("Failed to open /proc/modules", err)
+	}
+	found := false
```
aboch: You could drop this variable:
```
[...]
       if n == name {
              t.SkipUnlessRoot();
              return;
        }
   }
   t.Skipf("Test requires kmodule %q.", name)
}
```

aboch:
Thanks @paravmellanox . Just a minor comment.

rdma_link_linux.go
paravmellanox: @aboch, I will fix the comment.

rdma_link_test.go
paravmellanox: @aboch, sure. I will simplify this.

rdma_link_linux.go
```diff
@@ -0,0 +1,109 @@
...
+func RdmaLinkByName(name string) (*RdmaLink, error) {
+	return pkgHandle.RdmaLinkByName(name)
+}
+
+func (h *Handle) RdmaLinkByName(name string) (*RdmaLink, error) {
```
fcrisciani: missing comment for exported method

paravmellanox: @fcrisciani right; adding it.

aboch:
Looks like we do not run golint in this project, otherwise CI would have pointed out when exported function comments do not start with the function name.

If we ever decide to enable it in future, at that time we can take care of this issue across all files at once.

LGTM

paravmellanox:
I added the comment for exported function, addressed comments and pushed the changes again.

PR #344, opened 2018-04-26
Parse GRE IP addresses right

skoef:
I had some issue with some gretun devices after upgrading the netlink library in my project. I traced it down to 5a988e882da75e9451a5a6b02eecd36df51606ed where IPv6 support for these interfaces was introduced. My problem is that the IP addresses are always considered IPv6 addresses, even when it is supposed to be IPv4.

I added additional testing to `Gretun` and `Gretap` interfaces and it turned out that it probably isn't working properly for anyone. I also added a proposed fix for it. The only thing left to discuss is the fact that `PMtuDisc` isn't properly parsed for IPv6 `Gretap`s, but I'm wondering if this is even relevant; isn't MTU discovery dealt with differently for IPv6 in the first place?

### DIFF
--- a/link_linux.go
+++ b/link_linux.go
@@ -1974,9 +1974,9 @@ func parseGretapData(link Link, data []syscall.NetlinkRouteAttr) {
 		case nl.IFLA_GRE_IKEY:
 			gre.OKey = ntohl(datum.Value[0:4])
 		case nl.IFLA_GRE_LOCAL:
-			gre.Local = net.IP(datum.Value[0:16])
+			gre.Local = net.IP(datum.Value)
 		case nl.IFLA_GRE_REMOTE:
-			gre.Remote = net.IP(datum.Value[0:16])
+			gre.Remote = net.IP(datum.Value)
 		case nl.IFLA_GRE_ENCAP_SPORT:
 			gre.EncapSport = ntohs(datum.Value[0:2])
 		case nl.IFLA_GRE_ENCAP_DPORT:
@@ -1985,7 +1985,6 @@ func parseGretapData(link Link, data []syscall.NetlinkRouteAttr) {
 			gre.IFlags = ntohs(datum.Value[0:2])
 		case nl.IFLA_GRE_OFLAGS:
 			gre.OFlags = ntohs(datum.Value[0:2])
-
 		case nl.IFLA_GRE_TTL:
 			gre.Ttl = uint8(datum.Value[0])
 		case nl.IFLA_GRE_TOS:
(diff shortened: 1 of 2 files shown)


vishvananda:
This looks great. thank you  for the test. A couple things:

1) It appears that the new tests are not passing in CI. One of the errors, for sexample is: "link_test.go:308: Gretun.Remote doesn't match"
2) can you please squash your changes into a single commit?

skoef:
As you can see from a3ccbfb, the parsing of `IKey` and `OKey` didn't work properly either and I disabled setting `PMtuDisc` for the IPv6 check on `Gretap`. I'm still figuring out why the flow based check doesn't work

skoef:
This is the interface I end up with running `TestLinkAddDelGretapFlowBased`:
```
$ ip -d link show dev foo
5: foo@NONE: <BROADCAST,MULTICAST> mtu 1434 qdisc noop state DOWN mode DEFAULT group default
    link/ether 26:4d:12:64:00:99 brd ff:ff:ff:ff:ff:ff promiscuity 0
    ip6gretap remote any local any hoplimit inherit encaplimit 0 tclass 0x00 flowlabel 0x00000 addrgenmode eui64 numtxqueues 1 numrxqueues 1 gso_max_size 65536 gso_max_segs 65535
```

How can I even check if flow based/collect metadata is even enabled? Fow now I'll just remove the check from `compareGretun` and `compareGretap`

skoef:
Once again I made a complete mess of my git repo, any pointers how to squash the commits would be awesome!

vishvananda:
squashing is a useful tool to have into your aresenal. There are various ways to do it. Here is a flow that might work for you assuming your upstream repo is called vishvananda and your repo is called origin:
```
git fetch vishvananda # get the newest code
git merge vishvananda/master # merge the master branch into your branch
git branch -m greTunIPv6old # rename your branch
git checkout -b greTunIPv6 vishvananda/master # create a new branch from base code
git diff -R greTunIPv6old | git apply # apply the old changes to the new branch
git commit -a # commit the changes
git push --force origin greTunIPv6:greTunIPv6 # push the new changes
```

skoef:
That is the most comprehensive way of squashing I ever read 👍 Although I'm not sure it actually qualifies as squasing, it worked!

link_test.go
```diff
@@ -325,7 +453,6 @@ func TestLinkAddDelGretap(t *testing.T) {
 		LinkAttrs: LinkAttrs{Name: "foo6"},
 		IKey:      0x101,
 		OKey:      0x101,
-		PMtuDisc:  1,
```
vishvananda: did you remove this line on purpose? Was it causing a failure?

link_linux.go
```diff
@@ -2051,14 +2059,22 @@ func parseGretunData(link Link, data []syscall.NetlinkRouteAttr) {
 	gre := link.(*Gretun)
 	for _, datum := range data {
 		switch datum.Attr.Type {
-		case nl.IFLA_GRE_OKEY:
-			gre.IKey = ntohl(datum.Value[0:4])
```
aboch: :+1: Nice catch.

link_linux.go
```diff
@@ -1974,9 +1974,17 @@ func parseGretapData(link Link, data []syscall.NetlinkRouteAttr) {
 		case nl.IFLA_GRE_IKEY:
 			gre.OKey = ntohl(datum.Value[0:4])
 		case nl.IFLA_GRE_LOCAL:
-			gre.Local = net.IP(datum.Value[0:16])
+			if len(datum.Value) == 4 {
```
aboch: This logic is basically saying `use the whole Value buffer as input`. Can you please try replacing it with
```
gre.Local = net.IP(datum.Value)
```
At the end

link_test.go
skoef: It was on purpose: setting it isn't reflected in the resulting `ip6gretap` device so indeed checking for a match resulted in a failing check. The `IFLA_GRE_PMTUDISC` flag is missing from `net/ipv6/ip_gre.c` in the Linux kernel and I think that is because IPv6 uses different approaches to MTU discovery.

link_linux.go
vishvananda: I think @aboch was suggesting that in both cases we want to use the whole buffer, so we can replace the whole `if` statement with just
```
gre.Local = net.IP(datum.Value)
```
There doesn't seem to be any need to slice it at all.

vishvananda:
lgtm thanks

PR #345, opened 2018-04-26
Add info about VFs on link

jellonek:
We had already most required things (data deserialization) prepared, we had possibility to change VFs parameters on particular link - but there was no method to read them, what this PR changes.

Example use: https://gist.github.com/jellonek/1518bfe98ec4b46e07a430a06440f382 - output is ~ the same as returned by `ip link show ...` for particular link.

### DIFF
--- a/link.go
+++ b/link.go
@@ -41,6 +41,18 @@ type LinkAttrs struct {
 	NetNsID      int
 	NumTxQueues  int
 	NumRxQueues  int
+	Vfs          []VfInfo // virtual functions available on link
+}
+
+// VfInfo represents configuration of virtual function
+type VfInfo struct {
+	ID        int
+	Mac       net.HardwareAddr
+	Vlan      int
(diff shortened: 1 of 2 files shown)


jellonek:
@vishvananda, @aboch, @fcrisciani can you take a look?

link.go
```diff
@@ -41,6 +41,18 @@ type LinkAttrs struct {
 	NetNsID      int
 	NumTxQueues  int
 	NumRxQueues  int
+	Vfs          []VfInfo // virtual functions available on link
```
aboch: @jellonek Is LinkAttr the only place or is it the proper place where we can pass these virtual functions ? Can you think of another way to expose this setting ?
I am asking because this would break the existing API.

aboch: True, this is addition, we should be fine. Thanks.

aboch:
LGTM

PR #346, opened 2018-04-29
Run both Inline/Encap mode in TestSEG6RouteAddDel

ebiken:
TestSEG6RouteAddDel was skipped due to checking unnecessary Kernel build option and hiding a bug in SEG6_INLINE mode testing.
Changes in this PR will fix both of it. (Details will follow)

### DIFF
--- a/netlink_test.go
+++ b/netlink_test.go
@@ -109,12 +109,6 @@ func setUpSEG6NetlinkTest(t *testing.T) tearDownNetlinkTest {
 		log.Printf(msg)
 		t.Skip(msg)
 	}
-	key = string("CONFIG_IPV6_SEG6_INLINE=y")
-	if _, err := grepKey(key, filename); err != nil {
-		msg := "Skipped test because it requires SEG6_INLINE support."
-		log.Printf(msg)
-		t.Skip(msg)
-	}
 	// Add CONFIG_IPV6_SEG6_HMAC to support seg6_hamc
 	// key := string("CONFIG_IPV6_SEG6_HMAC=y")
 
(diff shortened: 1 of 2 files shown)


ebiken:
Kernel built without CONFIG_IPV6_SEG6_INLINE also accepts setting seg6 mode inline routes.
Thus you can run test even if you couldn't find "CONFIG_IPV6_SEG6_INLINE=y" in Kernel build options.

```
root@coreswitch:~/tmp/ebiken-netlink# git diff
diff --git a/netlink_test.go b/netlink_test.go
index e332ed2..5c0ad38 100644
--- a/netlink_test.go
+++ b/netlink_test.go
@@ -109,12 +109,6 @@ func setUpSEG6NetlinkTest(t *testing.T) tearDownNetlinkTest {
                log.Printf(msg)
                t.Skip(msg)
        }
-       key = string("CONFIG_IPV6_SEG6_INLINE=y")
-       if _, err := grepKey(key, filename); err != nil {
-               msg := "Skipped test because it requires SEG6_INLINE support."
-               log.Printf(msg)
-               t.Skip(msg)
-       }
        // Add CONFIG_IPV6_SEG6_HMAC to support seg6_hamc
        // key := string("CONFIG_IPV6_SEG6_HMAC=y")
```

You will get the error below when running TestSEG6RouteAddDel.
This is because inline mode requires IPv6 route. (SRH could not be added to IPv4)

```
root@coreswitch:~/tmp/ebiken-netlink# go test -run TestSEG6RouteAddDel
--- FAIL: TestSEG6RouteAddDel (0.00s)
        route_test.go:958: invalid argument
FAIL
exit status 1
FAIL    _/home/ebiken/tmp/ebiken-netlink        0.006s
```

Now apply the fix in this PR and run the test again.

```
root@coreswitch:~/tmp/ebiken-netlink# go test -run TestSEG6RouteAddDel
PASS
ok      _/home/ebiken/tmp/ebiken-netlink        0.006s
```

PR #356, opened 2018-05-23
[WIP]Add Neighbor Subscription

benagricola:


benagricola:
I've noticed an issue with this code (and potentially an issue with the Neighbor code in general) in that when a link is deleted (`ip link set eth0 down`) the `RTM_DELNEIGH` event seems to always be triggered *after* the `RTM_DELLINK` event.

This is a problem because `DeserializeNeigh` necessarily attempts to look up the related link to identify encapsulation type using `LinkByIndex` - on a link that no longer exists.

This throws an error and effectively causes the event to be discarded.

In this situation, we could still return `Neigh` but as we're unable to look up the encapsulation type on a deleted link then the values of `LLIPAddr` or `IP` in tunnel cases would not be accurate (or set).

Any input or opinions on how to resolve this would be very helpful.

neigh_linux.go
```diff
@@ -235,7 +238,7 @@ func (h *Handle) neighList(linkIndex, family, flags int) ([]Neigh, error) {
 	return res, nil
 }
 
-func NeighDeserialize(m []byte) (*Neigh, error) {
+func DeserializeNeigh(m []byte) (*Neigh, error) {
```
aboch: This has been an exported netlink package method.
We should not rename it for backward compatibility of existing code which may use it.

I like it too for the methods names to be consistent.
If you want to stick to it, you could 
```
func NeighDeserialize(m []byte) (*Neigh, error) {
+        return DeserializeNeigh(m)
+}
+func DeserializeNeigh(m []byte) (*Neigh, error) {
```

benagricola: Yep good point, I'll add the renamed methods to keep compatibility.

addr_linux.go
```diff
@@ -183,7 +183,7 @@ func (h *Handle) AddrList(link Link, family int) ([]Addr, error) {
 	return res, nil
 }
 
-func parseAddr(m []byte) (addr Addr, family, index int, err error) {
+func DeserializeAddr(m []byte) (addr Addr, family, index int, err error) {
```
aboch: If not required, we should not export methods. I get it `LinkDeserialize` and `NeighDeserialize` made it out, although I am not sure if for good reasons.

benagricola: My thinking behind this was that `nl.SubscribeAt` is already exported, so it's possible for library users to write functionality based on that, which doesn't  actually use the individual `*Subscribe` methods.

This is how I started out subscribing to neighbor events prior to modifying the library itself (this PR), and I still use it in this fashion listening to routes + neighbors + links + addresses from a single `nl.SubscribeAt` call.

In those sorts of implementations, it's nice to have consistent library functions that are able to decode the relevant netlink messages, and as it already existed, I exported it when renaming 😃 

Point taken though that this increases the public API surface area complexity and this may not be the best way forward...

PR #359, opened 2018-06-14
Implement "vti6" device support

semicomplete:
Updates VTI code to support "vti6"devices in addition to "vti".
Also fixes a bug in point-to-point address assignment/parsing. VTI devices functionality depends on the fix.

### DIFF
--- a/link.go
+++ b/link.go
@@ -784,9 +784,8 @@ func (vti *Vti) Attrs() *LinkAttrs {
 func (vti *Vti) Type() string {
 	if vti.Local.To4() == nil {
 		return "vti6"
-	} else {
-		return "vti"
 	}
+	return "vti"
 }
 
 type Gretun struct {
--- a/link.go
+++ b/link.go
@@ -781,8 +781,12 @@ func (vti *Vti) Attrs() *LinkAttrs {
 	return &vti.LinkAttrs
 }
 
-func (iptun *Vti) Type() string {
-	return "vti"
+func (vti *Vti) Type() string {
+	if vti.Local.To4() == nil {
+		return "vti6"
+	} else {
+		return "vti"
+	}
 }
 
 type Gretun struct {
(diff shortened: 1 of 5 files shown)


link.go
```diff
@@ -781,8 +781,12 @@ func (vti *Vti) Attrs() *LinkAttrs {
 	return &vti.LinkAttrs
 }
 
-func (iptun *Vti) Type() string {
-	return "vti"
+func (vti *Vti) Type() string {
+	if vti.Local.To4() == nil {
+		return "vti6"
+	} else {
```
aboch: Please remove the `else`

link_linux.go
```diff
@@ -2304,12 +2304,27 @@ func parseSittunData(link Link, data []syscall.NetlinkRouteAttr) {
 func addVtiAttrs(vti *Vti, linkInfo *nl.RtAttr) {
 	data := nl.NewRtAttrChild(linkInfo, nl.IFLA_INFO_DATA, nil)
 
-	ip := vti.Local.To4()
```
aboch: I understand it is just personal preference, but we could avoid the new `family` variable (and most of all for a future reader to have to look into the function if the variable is used anywhere else) if we feel the following is clear enough:
```
ip := vti.Local
if ip.To4() != nil { 
    ip = ip.To4()
}
```
same for `vti.Remote` below

semicomplete: The reason is that `family` determines both addresses family. I.e. it doesn't make sense to test family of `Remote`, as both should belong to the same family.

Would you still prefer calling `.To4()` twice?

aboch: Yes, either ways you look at it it has its pros and cons. Just personal preference. I am fine to leave it as is too.

aboch:
@semicomplete I know @vishvananda prefers single commit per PR, when possible.
Can you please then open a new PR for the prefixlen PtP fix and one have this one for the vti6 support only. Thanks.

semicomplete:
Added two separate pull-requests #360 and #361
Closing this one.

PR #363, opened 2018-06-26
Fix: Proper parsing of mark

sibicramesh:
--> This PR fixes the bug that breaks the parsing

`udp	17 src=127.0.0.1 dst=127.0.0.10 sport=2003 dport=3000 packets=0 bytes=0	src=127.0.0.10 dst=127.0.0.1 sport=3000 dport=2003 packets=0 bytes=0 mark=0`

### DIFF
--- a/conntrack_linux.go
+++ b/conntrack_linux.go
@@ -242,59 +242,59 @@ func parseByteAndPacketCounters(r *bytes.Reader) (bytes, packets uint64) {
 }
 
 func parseRawData(data []byte) *ConntrackFlow {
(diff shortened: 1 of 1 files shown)


fcrisciani:
@sibicramesh can you keep the original indentation so to minimize the diff?

fcrisciani:
can you also add a test, so that we avoid further regressions?

aboch:
@sibicramesh Please also squash your commits in one after you fix the formatting
`gofmt *.go`

aboch:
Closing for inactivity.

PR #366, opened 2018-07-04
[RFC] protinfo: Check if object is nil

jodh-intel:
Avoid segfaults in certain environments by checking if the Stringer
interface pointer receiver function is `nil`.

Signed-off-by: James O. D. Hunt <james.o.hunt@intel.com>

### DIFF
--- a/protinfo.go
+++ b/protinfo.go
@@ -18,6 +18,10 @@ type Protinfo struct {
 
 // String returns a list of enabled flags
 func (prot *Protinfo) String() string {
+	if prot == nil {
+		return "<nil>"
+	}
+
 	var boolStrings []string
 	if prot.Hairpin {
 		boolStrings = append(boolStrings, "Hairpin")


jodh-intel:
As shown by the summary, this is an RFC. I'm seeing a segfault at runtime without this check (see https://github.com/kata-containers/agent/issues/281) but I haven't yet pinned down why this is happening as it only happens when our agent process runs as PID 1.

protinfo.go
```diff
@@ -18,6 +18,10 @@ type Protinfo struct {
 
 // String returns a list of enabled flags
 func (prot *Protinfo) String() string {
+	if prot == nil {
+		return ""
```
fcrisciani: I guess it will be better to return something like "<nil>" in order to distinguish from an object that is actually properly pointed but has no flag enabled

jellonek: In stdlib in such places we have more often used "<nil>".

jodh-intel: Thanks for reviewing! Branch updated to make the function return `"<nil>"`.

PR #368, opened 2018-07-06
Ingress qdisc add/del Test case

paravmellanox:


qdisc_test.go
```diff
@@ -456,3 +456,37 @@ func TestFqCodelAddChangeDel(t *testing.T) {
 		t.Fatal("Failed to remove qdisc")
 	}
 }
+
+func TestIngressAddDel(t *testing.T) {
```
jellonek: This test func is missing:
```
	defer tearDown()
	if err := LinkAdd(&Ifb{LinkAttrs{Name: "foo"}}); err != nil {
		t.Fatal(err)
	}
```
at the beginning.
If it's working without that - we have a bigger issue ;)

jellonek: So will that work with `foo`? I'm not fully clear if it has to be prepared somehow to be working in any testenv.

paravmellanox: @jellonek yes, it works with foo. Test in this PR is no different than any other tests in qdisc_test.go file.
Preparation test only does in different net namespace. So PR is updated like other tests for pre-test setup. I do have tested with foo as well.

aboch:
LGTM

PR #370, opened 2018-07-11
Add network namespace ID management.

squeed:
Adds the ability to set and retrieve network namespace IDs. This is useful, for example, for determining the "other side" of a veth pair.

### DIFF
--- a/netns_linux.go
+++ b/netns_linux.go
@@ -1,0 +1,141 @@
+package netlink
+
+// Network namespace ID functions
+//
+// The kernel has a weird concept called the network namespace ID.
+// This is different from the file reference in proc (and any bind-mounted
+// namespaces, etc.)
+//
+// Instead, namespaces can be assigned a numeric ID at any time. Once set,
+// the ID is fixed. The ID can either be set manually by the user, or
+// automatically, triggered by certain kernel actions. The most common kernel
+// action that triggers namespace ID creation is moving one end of a veth pair
+// in to that namespace.
+
+import (
+	"fmt"
+
+	"github.com/vishvananda/netlink/nl"
+	"golang.org/x/sys/unix"
+)
+
+// These can be replaced by the values from sys/unix when it is next released.
+const (
+	_ = iota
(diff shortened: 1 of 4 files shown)


netns_unspecified.go
```diff
@@ -0,0 +1,19 @@
+// +build !linux
+
+package netlink
+
+func GetNetNsIdByPid(pid int) (int, error) {
+	return -1, ErrNotImplemented
```
jellonek: IMO it's more common to return a zero value instead of -1 in case of error.
Setting there -1 would be more like double error indication with false information.

netns_unspecified.go
```diff
@@ -0,0 +1,19 @@
...
+}
+
+func SetNetNsIdByPid(pid, nsid int) error {
+	return ErrNotImplemented
+}
+
+func GetNetNsIdByFd(fd int) (int, error) {
+	return -1, ErrNotImplemented
```
jellonek: Same as above.

netns_linux.go
```diff
@@ -0,0 +1,128 @@
...
+			case NETNSA_NSID:
+				return int(int32(native.Uint32(attr.Value))), nil
+			}
+		}
+	}
+
+	return 0, fmt.Errorf("unexpected empty result")
+}
+
+// getNetNsId sets the netnsid for a given type-val pair
```
jellonek: setNetNsId

netns_unspecified.go
squeed: The challenge is that 0 is a valid namespace ID.

aboch: I think @jellonek comment is more about current diff overloads the `int` return param having it signal an error. Which is unnecessary and confusing for the reader of the code.

Given the function (correctly in go world) returns an error param, caller must honor it and not even look or use the other return params in case err != nil. The double error marking is unnecessary and confusing.

squeed:
OK, updated. I added some tests and changed the error value.

Issue #373, opened 2018-07-25
Conntrack nat filters incompatibility


bboreham:
I see in `github.com/vishvananda/netlink/conntrack_linux.go` the following comments, with code to match:

```
// Conntrack parameters and options:                                                                                                              
//   -n, --src-nat ip                      source NAT ip                                                                                          
//   -g, --dst-nat ip                      destination NAT ip                                                                                     
//   -j, --any-nat ip                      source or destination NAT ip                                                                           
```

This does not match the `conntrack` command, e.g. from its [man page](https://manpages.debian.org/testing/conntrack/conntrack.8.en.html)

```
-n, --src-nat
    Filter source NAT connections.
-g, --dst-nat
    Filter destination NAT connections.
-j, --any-nat
    Filter any NAT connections.
```

There is no IP address on these options - they simply filter the conntrack table for entries that involve NAT.

Of course, it's fine for `vishvananda/netlink` to document and implement something different from standard Linux conntrack.  But I did find it odd so thought I would raise this to check.

vishvananda:
I believe that this is only an error in the comment and the functionality is the same as the conntrack command. Do you see an actual difference in the implementation?

bboreham:
I'm looking at [this](https://github.com/vishvananda/netlink/blob/master/conntrack_linux.go#L374):

	match = match && elem.Equal(flow.Forward.DstIP)

where it is checking an IP, and there is no IP to check in the conntrack command.  So yes I think it is different.

vishvananda:
hmm looks like you may be right. @fcrisciani I believe you authored that code. Any comments?

fcrisciani:
@bboreham @vishvananda will have to check the code of the conntrack command, don't recall on top of my mind.
As far as I can see the filter aims to cover these:
```
      -s, --orig-src IP_ADDRESS
              Match only entries whose source address in the original direction equals the one specified as argument.

       -d, --orig-dst IP_ADDRESS
              Match only entries whose destination address in the original direction equals the one specified as argument.

       -r, --reply-src IP_ADDRESS
              Match only entries whose source address in the reply direction equals the one specified as argument.

       -q, --reply-dst IP_ADDRESS
              Match only entries whose destination address in the reply direction equals the one specified as argument.
```
That's why there is the match there.
The one that needs a review I guess is `ConntrackNatAnyIP         // -any-nat ip    Source or destination NAT ip` for which I don't remember the details, and what really the man means when says:
```
-j, --any-nat
[...]

PR #374, opened 2018-07-26
Adjust conntrack filter doc

fcrisciani:
Today the filter implementation implements
only ip matching for src,dst,reply src,reply dst.
Updating the comments on the filter to reflect that
more clearly

Fixes: #373 

Signed-off-by: Flavio Crisciani <flavio.crisciani@docker.com>

### DIFF
--- a/conntrack_linux.go
+++ b/conntrack_linux.go
@@ -309,7 +309,7 @@ func parseRawData(data []byte) *ConntrackFlow {
 // Common parameters and options:
 //   -s, --src, --orig-src ip              Source address from original direction
 //   -d, --dst, --orig-dst ip              Destination address from original direction
-//   -r, --reply-src ip            Source addres from reply direction
+//   -r, --reply-src ip            Source address from reply direction
 //   -q, --reply-dst ip            Destination address from reply direction
 //   -p, --protonum proto          Layer 4 Protocol, eg. 'tcp'
 //   -f, --family proto            Layer 3 Protocol, eg. 'ipv6'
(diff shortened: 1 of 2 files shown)


fcrisciani:
cc @bboreham

conntrack_linux.go
```diff
@@ -326,11 +326,11 @@ func parseRawData(data []byte) *ConntrackFlow {
...
-	ConntrackOrigSrcIP = iota // -orig-src ip   Source address from original direction
-	ConntrackOrigDstIP        // -orig-dst ip   Destination address from original direction
-	ConntrackNatSrcIP         // -src-nat ip    Source NAT ip
-	ConntrackNatDstIP         // -dst-nat ip    Destination NAT ip
-	ConntrackNatAnyIP         // -any-nat ip    Source or destination NAT ip
+	ConntrackOrigSrcIP = iota // -orig-src ip    Source address from original direction
+	ConntrackOrigDstIP        // -orig-dst ip    Destination address from original direction
+	ConntrackNatSrcIP         // --reply-src ip  Reply Source IP
+	ConntrackNatDstIP         // --reply-dst ip  Reply Destination IP
+	ConntrackNatAnyIP         // Match source or destination reply IP
```
jellonek: This one still seems to be confusing in this form. It's also not consistent with comments in `MatchConntrackFlow` at the bottom of file.

fcrisciani: Sure will update the MatchConntrack flow comment, what else is it confusing?

jellonek: I was confused by this "reply" addition but that was my fault in understanding the flow. It's perfectly fine as is.

conntrack_linux.go
```diff
@@ -384,7 +384,7 @@ func (f *ConntrackFilter) MatchConntrackFlow(flow *ConntrackFlow) bool {
 		match = match && elem.Equal(flow.Reverse.DstIP)
 	}
 
-	// -any-nat ip    Source or destination NAT ip
+	// Match source or destination reply IP
```
jellonek: As this comment is updated to match comment in consts also both two above comments could be also updated, but it's nothing blocking imo.

aboch:
@fcrisciani The const you are changing are actually exported and were there before we cut `1.0.0`.
You may add the new ones and mark the existing ones as deprecated in their comment.

conntrack_linux.go
```diff
@@ -326,11 +326,11 @@ func parseRawData(data []byte) *ConntrackFlow {
 type ConntrackFilterType uint8
 
 const (
-	ConntrackOrigSrcIP = iota // -orig-src ip   Source address from original direction
```
aboch: @fcrisciani The const you are changing are actually exported and were there before we cut `1.0.0`.
You may add the new ones and mark the existing ones as deprecated in their comment.

fcrisciani:
@aboch rebased and deprecated the 3 names in favor of 3 new ones

aboch:
LGTM

PR #377, opened 2018-08-13
Add support for action and ifindex in XFRM policy

ctelfer:
The purpose of this PR is to allow the user to create "block" action IPSec policies and/or IPSec policies that apply only to packets traversing a specific interface.

The action and ifindex fields aren't represented in the netlink.XfrmPolicy struct although they exist in the the linux equivalent data structures and are, of course, represented in the serialized versions of those datatypes.  So this patch simply exposes those fields to the user-consumable side of the API.  

This patch makes the policy's action a specific type in the same style as the Dir field in XfrmPolicy.

Signed-off-by: Chris Telfer <ctelfer@docker.com>

### DIFF
--- a/xfrm_policy.go
+++ b/xfrm_policy.go
@@ -35,6 +35,25 @@ func (d Dir) String() string {
 	return fmt.Sprintf("socket %d", d-XFRM_SOCKET_IN)
 }
 
+// PolicyAction is an enum representing an ipsec policy action.
+type PolicyAction uint8
+
+const (
+	XFRM_POLICY_ALLOW PolicyAction = 0
+	XFRM_POLICY_BLOCK PolicyAction = 1
+)
+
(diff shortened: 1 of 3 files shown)


aboch:
Thanks. Please extend the existing UT verifying policy returned by kernel has ifIndex == 0 (If that is the default when iface was not specified during config)  and policy is allow for the existing UT.
Maybe also add a test where the xfrm rule has block policy.

ctelfer:
Updated and added unit tests as requested.   Interface index 0 is indeed the default and according to RFC 3493 is basically the only "invalid" or "unspecified" value (since the index is unsigned).  Thanks for the prompt response!

aboch:
LGTM

As per this repository convention, please squash the two commits in one as they refer to the same functionality. (GH merge-squash is disabled here)

PR #381, opened 2018-08-25
Service curves fix + expansion

fbegyn:
Added a function that simply sets a rate on a curve, so the behaviour oc tc class add .. sc/ls/ul rate is mimmicked.

There was something of with the math for the ServiceCurves. The set value was 8 times to large and the returned one was 8 times to small. This is now fixed in class_linux.go so the user can just set the rate in bit and it will match the tc show output

I know this is kinda breaking the previous service curves, but this makes more sense. The end user shouldn't have to calculate the need values, the unerlaying layers should do that.

### DIFF
--- a/class.go
+++ b/class.go
@@ -154,33 +154,48 @@ type HfscClass struct {
 
 // SetUsc sets the Usc curve
 func (hfsc *HfscClass) SetUsc(m1 uint32, d uint32, m2 uint32) {
-	hfsc.Usc = ServiceCurve{m1: m1 / 8, d: d, m2: m2 / 8}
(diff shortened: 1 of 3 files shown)


class.go
```diff
@@ -154,33 +154,48 @@ type HfscClass struct {
 
 // SetUsc sets the Usc curve
 func (hfsc *HfscClass) SetUsc(m1 uint32, d uint32, m2 uint32) {
-	hfsc.Usc = ServiceCurve{m1: m1 / 8, d: d, m2: m2 / 8}
```
aboch: This will change the exported method semantic

PR #382, opened 2018-08-29
Avoid serializing empty TCA_OPTIONS in qdisc messages

joestringer:
This was causing a QdiscReplace() call for a generic qdisc
(QdiscType="clsact", Parent=HANDLE_CLSACT) to fail with the error
"invalid argument" after the qdisc was first created.


joestringer:
@aboch @fcrisciani @vishvananda  did you get a chance to look over this proposal?

Having a 0-sized options TLV messes up the kernel side for a few of the qdics types that lack `sch->ops->change` implementations, and this solves an issue for us.

aboch:
LGTM

PR #383, opened 2018-09-05
fix: BRIDGE_FLAGS_* constants off-by-one

willkelleher:
These flags are off-by-one because iota starts at 0.

See https://github.com/torvalds/linux/blob/master/include/uapi/linux/if_bridge.h#L104 for expected values.

The result is that it becomes impossible to modify VLANs on the bridge interface itself, where the `self=true` flag is required.

### DIFF
--- a/nl/bridge_linux.go
+++ b/nl/bridge_linux.go
@@ -11,8 +11,8 @@ const (
 
 /* Bridge Flags */
 const (
-	BRIDGE_FLAGS_MASTER = iota /* Bridge command to/from master */
-	BRIDGE_FLAGS_SELF          /* Bridge command to/from lowerdev */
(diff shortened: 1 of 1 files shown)


nl/bridge_linux.go
```diff
@@ -11,8 +11,8 @@ const (
 
 /* Bridge Flags */
 const (
-	BRIDGE_FLAGS_MASTER = iota /* Bridge command to/from master */
-	BRIDGE_FLAGS_SELF          /* Bridge command to/from lowerdev */
+	BRIDGE_FLAGS_MASTER = iota + 1 /* Bridge command to/from master */
```
aboch: :+1:

aboch:
LGTM

PR #384, opened 2018-09-05
Add support for hoplimit metric in routes

mageshgv:
Using Hoplimit in routes is a much lesser used functionality in linux. The default hoplimit is 0 which means use system defaults: net.ipv4.ip_default_ttl for ipv4 and net.ipv6.conf.all.hop_limit for ipv6

Equivalent in iproute2: ip route add 192.168.0.0/24 via 192.168.10.1 hoplimit 2

### DIFF
--- a/route.go
+++ b/route.go
@@ -47,6 +47,7 @@ type Route struct {
 	Encap      Encap
 	MTU        int
 	AdvMSS     int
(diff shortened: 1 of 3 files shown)


aboch:
LGTM

PR #390, opened 2018-09-30
Export NeighListFiltered

corny:
I need to get a list of all reachable neighbors.

Does anyone know a better name for this method?

### DIFF
--- a/neigh_linux.go
+++ b/neigh_linux.go
@@ -45,6 +45,7 @@ const (
 	NTF_ROUTER = 0x80
 )
 
+// Ndmsg is for adding, removing or receiving information about a neighbor table entry
 type Ndmsg struct {
 	Family uint8
 	Index  uint32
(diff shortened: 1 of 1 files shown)


neigh_linux.go
```diff
@@ -243,7 +243,7 @@ func (h *Handle) NeighListFiltered(linkIndex, family, flags, state int) ([]Neigh
 	return res, nil
 }
 
-func NeighDeserialize(m []byte) (*Neigh, error) {
+func neighDeserialize(m []byte) (*Neigh, error) {
```
aboch: I agree with this, but unfortunately we release `1.0.0` which had this method exported (this was there for 4 years).
Unexporting it would be a breaking change. We'd prefer avoiding a major release change for this.

neigh_linux.go
```diff
@@ -174,40 +175,47 @@ func neighHandle(neigh *Neigh, req *nl.NetlinkRequest) error {
...
 // Equivalent to: `ip neighbor show proxy`.
 // The list can be filtered by link, ip family.
 func (h *Handle) NeighProxyList(linkIndex, family int) ([]Neigh, error) {
-	return h.neighList(linkIndex, family, NTF_PROXY)
+	return h.NeighListFiltered(linkIndex, family, NTF_PROXY, 0)
 }
 
-func (h *Handle) neighList(linkIndex, family, flags int) ([]Neigh, error) {
+// NeighListFiltered returns a list of neighbour entries filtered by link, ip family, flag and state.
+func (h *Handle) NeighListFiltered(linkIndex, family, flags, state int) ([]Neigh, error) {
```
aboch: should the `state` field be `uint16` or do we have a convention in this project where exported method simply take a `int` parameter and caller is supposed to know the value range ?

aboch:
Thanks @corny 
Please remove the unexport change commit from this PR.

aboch:
>Does anyone know a better name for this method?

`NeighListByState` ? What you have `NeighListFiltered` seems fine too.

@fcrisciani @vishvananda for their input

neigh_linux.go
fcrisciani: I like more the uint16 convention

corny:
Thanks for you feedback. I've updated the pull request.

neigh_linux.go
```diff
@@ -176,40 +177,47 @@ func neighHandle(neigh *Neigh, req *nl.NetlinkRequest) error {
...
+// NeighProxyList returns a list of neighbor proxies in the system.
 // Equivalent to: `ip neighbor show proxy`.
 // The list can be filtered by link and ip family.
 func NeighProxyList(linkIndex, family int) ([]Neigh, error) {
 	return pkgHandle.NeighProxyList(linkIndex, family)
 }
 
-// NeighList gets a list of IP-MAC mappings in the system (ARP table).
+// NeighListFiltered returns a list of neighbour entries filtered by link, ip family, flag and state.
+func NeighListFiltered(linkIndex, family, flags int, state uint16) ([]Neigh, error) {
```
aboch: I think we should have a structure which describe the query filter as parameter, instead of a list of (actual) supported filter params. 

Also, at least in the comment we should specify what are the bare minimum params that need to be specified.
Or would all entries be returned if no param is specified ?

aboch: Something like
```
func NeighListFiltered(NeighFilter filter) ([]Neigh, error) {
```
we should also investigate if such a filtered query exist for other areas like routes and see if we can come out with a generic filter structure which works for all. Or at least a base iproute2 filter which other filters will embed.

corny: In that case we should simply use `Ndmsg`. I added another commit.

corny:
One test produces instable results:
```
=== RUN   TestNeighSubscribeWithOptions
--- FAIL: TestNeighSubscribeWithOptions (0.22s)
	neigh_test.go:371: Add update not received as expected: want 28 got 0
	neigh_test.go:337: Fatal error received during subscription: Link not found
```
Sometimes it fails with Go 1.10, sometimes with Go 1.11.

neigh_linux.go
```diff
@@ -191,34 +191,30 @@ func NeighProxyList(linkIndex, family int) ([]Neigh, error) {
...
+		Family: uint8(family),
+		Index:  uint32(linkIndex),
+		Flags:  NTF_PROXY,
+	})
 }
 
-// NeighListFiltered returns a list of neighbour entries filtered by link, ip family, flag and state.
-func (h *Handle) NeighListFiltered(linkIndex, family, flags int, state uint16) ([]Neigh, error) {
+// NeighListExecute returns a list of neighbour entries filtered by link, ip family, flag and state.
+func (h *Handle) NeighListExecute(msg Ndmsg) ([]Neigh, error) {
```
aboch: Please also add a package function for this `func NeighListExecute(msg Ndmsg) ([]Neigh, error) {`

corny: done

corny:
Okay, ready to be merged?

fcrisciani:
@corny there is a couple of test failures:
```
--- FAIL: TestNeighSubscribe (0.38s)
	neigh_test.go:317: Add update not received as expected: want 28 got 0
[...]
--- FAIL: TestNeighSubscribeAt (0.21s)
	neigh_test.go:422: Add update not received as expected: want 28 got 0
```

aboch:
Pls rebase

corny:
Rebase done.

aboch:
LGTM

PR #391, opened 2018-09-30
Refactoring

corny:
This might break external code using `nl.NewRtAttrChild()`.

### DIFF
--- a/link_linux.go
+++ b/link_linux.go
@@ -1524,7 +1524,8 @@ func LinkDeserialize(hdr *unix.NlMsghdr, m []byte) (Link, error) {
 				if err != nil {
 					return nil, err
 				}
-				base.Protinfo = parseProtinfo(attrs)
+				protinfo := parseProtinfo(attrs)
+				base.Protinfo = &protinfo
 			}
 		case unix.IFLA_OPERSTATE:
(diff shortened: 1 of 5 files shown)


aboch:
>This might break external code using nl.NewRtAttrChild().

Yes this is a problem as we started versioning this library.

I suggest to retain the `NewRtAttrChild()` definition along with your newly added `func (a *RtAttr) AddRtAttr()`

Please also note in this project one commit per PR is encouraged.
The NewRtAttr() change deserve its own PR.

Thanks.

corny:
Thanks for your feedback. I've split up my refactoring into three pull requests.

link_linux.go
```diff
@@ -1524,7 +1524,8 @@ func LinkDeserialize(hdr *unix.NlMsghdr, m []byte) (Link, error) {
 				if err != nil {
 					return nil, err
 				}
-				base.Protinfo = parseProtinfo(attrs)
+				protinfo := parseProtinfo(attrs)
+				base.Protinfo = &protinfo
```
aboch: Why this change ?

aboch:
LGTM

link_linux.go
corny: Yes .. it does not return a pointer any more.

PR #393, opened 2018-10-04
Avoid function calls to NativeEndian

corny:


aboch:
May you add some more information about what this change is trying to achieve. Thanks.

nl/nl_unspecified.go
```diff
@@ -6,6 +6,6 @@ import "encoding/binary"
 
 var SupportedNlFamilies = []int{}
 
-func NativeEndian() binary.ByteOrder {
+func native binary.ByteOrder {
```
dmke: This seems like a mistake. Search and replace too greedy?

dmke:
@aboch: This PR consists of 3 changes:

1. Rename `native` to `nativeOrder`. (I guess to better communicate intent)
2. Move endian detection from `NativeEndian()` to program startup and degrade `NativeEndian()` to a simple getter.
3. Replace internal `NativeEndian()` calls with the cached `nativeOrder` value. This removes one function call from the stack.

vishvananda:
I'm ok with this approach in general if you want to rebase it, although I'm not sure the savings are significant enough to make a large difference.

corny:
@vishvananda I rebased my changes on top of the current master. Please review them.

nl/nl_linux.go
```diff
@@ -46,32 +46,30 @@ func GetIPFamily(ip net.IP) int {
 	return FAMILY_V6
 }
 
-var nativeEndian binary.ByteOrder
+var nativeEndian = func() binary.ByteOrder {
```
dmke, line 49: I'd add a comment, stating something along the lines of "native byte order as determined at program startup" and "assuming it doesn't change at runtime".

dmke (changes requested):
LGTM, but I'd like to see a bit of documentation to *clearly* state the intended purpose (see comment below).

PR #395, opened 2018-10-23
Add an API to rename rdma device name

paravmellanox:
Add an API RdmaLinkSetName() and test case to rename a
rdma device name.


aboch:
LGTM

PR #400, opened 2018-10-29
Run TravisCI with Go 1.10 and 1.11

corny:


aboch:
Should we at least keep a run for 1.9.x ?
It was released 1 year ago only.

corny:
Good question. By default only the most recent major versions of Go are supported by the Go maintainers: https://github.com/golang/go/wiki/MinorReleases

PR #402, opened 2018-11-12
Make the RoutingTable a uint32. It's defined that way in the kernel.

cedi:
The Kernel defines the routing table as an uint32_t, so should we do in
our go implementation.
Not sure if i catched everything here, but a grep looks promising.
Unfortunately the tests are failing on my machine. Looks like some
kernel options aren't supported by my kernel. Let's see what the CI says
about this commit :)

### DIFF
--- a/route.go
+++ b/route.go
@@ -38,7 +38,7 @@ type Route struct {
 	MultiPath  []*NexthopInfo
 	Protocol   int
 	Priority   int
-	Table      int
+	Table      uint32
 	Type       int
 	Tos        int
 	Flags      int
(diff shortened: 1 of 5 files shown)


aboch:
This came up before in the past. If I remember correctly, @vishvananda chose this approach for most exported structure and method parameters in this library, as he wanted them to use the basic `int` type, not to expose the exact sized int used by the correspondent kernel API.

vishvananda:
closing this because as mentioned the choice of int for the high-level interface was intentional

PR #404, opened 2018-11-18
Add devlink commands for devlink device information

paravmellanox:
This patch adds very basic support for getting information about
devlink devices which are typically PCI devices which exposes Networking
switch or legacy devices.

This information includes bus name, device name and eswitch modes.
This is done through devlink family of commands via generic netlink
sockets provided by Linux kernel.

DevlinkDevice represents a devlink device which is identified by bus
name and device name (unlike interface index for netdevices).
It contains the DevlinkDevAttrs device attributes.
Currently only eswitch attributes are queried. In future more attributes
such as port, shared buffer, traffic class will be added.

### DIFF
--- a/devlink_linux.go
+++ b/devlink_linux.go
@@ -1,0 +1,167 @@
+package netlink
+
+import (
+	"syscall"
+
+	"github.com/vishvananda/netlink/nl"
+	"golang.org/x/sys/unix"
+)
+
+// DevlinkDevEswitchAttr represents device's eswitch attributes
+type DevlinkDevEswitchAttr struct {
+	Mode       string
+	InlineMode string
+	EncapMode  string
+}
+
+// DevlinkDevAttrs represents device attributes
+type DevlinkDevAttrs struct {
+	Eswitch DevlinkDevEswitchAttr
+}
+
+// DevlinkDevice represents device and its attributes
+type DevlinkDevice struct {
+	BusName    string
+	DeviceName string
+	Attrs      DevlinkDevAttrs
+}
+
+func parseDevLinkDeviceList(msgs [][]byte) ([]*DevlinkDevice, error) {
+	devices := make([]*DevlinkDevice, 0, len(msgs))
+	for _, m := range msgs {
+		attrs, err := nl.ParseRouteAttr(m[nl.SizeofGenlmsg:])
+		if err != nil {
+			return nil, err
+		}
+		dev := &DevlinkDevice{}
+		if err = dev.parseAttributes(attrs); err != nil {
+			return nil, err
+		}
+		devices = append(devices, dev)
+	}
+	return devices, nil
+}
+
+func parseEswitchMode(mode uint16) string {
+	var eswitchMode = map[uint16]string{
+		nl.DEVLINK_ESWITCH_MODE_LEGACY:    "legacy",
+		nl.DEVLINK_ESWITCH_MODE_SWITCHDEV: "switchdev",
+	}
+	if eswitchMode[mode] == "" {
+		return "unknown"
+	} else {
+		return eswitchMode[mode]
+	}
+}
+
+func parseEswitchInlineMode(inlinemode uint8) string {
+	var eswitchInlineMode = map[uint8]string{
+		nl.DEVLINK_ESWITCH_INLINE_MODE_NONE:      "none",
+		nl.DEVLINK_ESWITCH_INLINE_MODE_LINK:      "link",
+		nl.DEVLINK_ESWITCH_INLINE_MODE_NETWORK:   "network",
+		nl.DEVLINK_ESWITCH_INLINE_MODE_TRANSPORT: "transport",
+	}
+	if eswitchInlineMode[inlinemode] == "" {
+		return "unknown"
+	} else {
+		return eswitchInlineMode[inlinemode]
+	}
+}
+
+func parseEswitchEncapMode(encapmode uint8) string {
+	var eswitchEncapMode = map[uint8]string{
+		nl.DEVLINK_ESWITCH_ENCAP_MODE_NONE:  "disable",
+		nl.DEVLINK_ESWITCH_ENCAP_MODE_BASIC: "enable",
+	}
+	if eswitchEncapMode[encapmode] == "" {
+		return "unknown"
+	} else {
+		return eswitchEncapMode[encapmode]
+	}
+}
+
+func (d *DevlinkDevice) parseAttributes(attrs []syscall.NetlinkRouteAttr) error {
+	for _, a := range attrs {
+		switch a.Attr.Type {
+		case nl.DEVLINK_ATTR_BUS_NAME:
+			d.BusName = string(a.Value)
+		case nl.DEVLINK_ATTR_DEV_NAME:
+			d.DeviceName = string(a.Value)
+		case nl.DEVLINK_ATTR_ESWITCH_MODE:
+			d.Attrs.Eswitch.Mode = parseEswitchMode(native.Uint16(a.Value))
+		case nl.DEVLINK_ATTR_ESWITCH_INLINE_MODE:
+			d.Attrs.Eswitch.InlineMode = parseEswitchInlineMode(uint8(a.Value[0]))
+		case nl.DEVLINK_ATTR_ESWITCH_ENCAP_MODE:
+			d.Attrs.Eswitch.EncapMode = parseEswitchEncapMode(uint8(a.Value[0]))
+		}
+	}
+	return nil
+}
+
+func (dev *DevlinkDevice) parseEswitchAttrs(msgs [][]byte) {
+	m := msgs[0]
+	attrs, err := nl.ParseRouteAttr(m[nl.SizeofGenlmsg:])
+	if err != nil {
+		return
+	}
+	dev.parseAttributes(attrs)
+}
+
+func (h *Handle) getEswitchAttrs(family *GenlFamily, dev *DevlinkDevice) {
+	msg := &nl.Genlmsg{
+		Command: nl.DEVLINK_CMD_ESWITCH_GET,
+		Version: nl.GENL_DEVLINK_VERSION,
+	}
+	req := h.newNetlinkRequest(int(family.ID), unix.NLM_F_REQUEST|unix.NLM_F_ACK)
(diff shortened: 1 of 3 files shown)


devlink_linux.go
```diff
@@ -0,0 +1,167 @@
...
+			d.Attrs.Eswitch.InlineMode = parseEswitchInlineMode(uint8(a.Value[0]))
+		case nl.DEVLINK_ATTR_ESWITCH_ENCAP_MODE:
+			d.Attrs.Eswitch.EncapMode = parseEswitchEncapMode(uint8(a.Value[0]))
+		}
+	}
+	return nil
+}
+
+func parseEswitchAttrs(msgs [][]byte, dev *DevlinkDevice) {
+
```
callmehollow: unnecessary empty line

devlink_linux.go
```diff
@@ -0,0 +1,167 @@
...
+		case nl.DEVLINK_ATTR_ESWITCH_INLINE_MODE:
+			d.Attrs.Eswitch.InlineMode = parseEswitchInlineMode(uint8(a.Value[0]))
+		case nl.DEVLINK_ATTR_ESWITCH_ENCAP_MODE:
+			d.Attrs.Eswitch.EncapMode = parseEswitchEncapMode(uint8(a.Value[0]))
+		}
+	}
+	return nil
+}
+
+func parseEswitchAttrs(msgs [][]byte, dev *DevlinkDevice) {
```
callmehollow: May be make that func a method of *DevlinkDevice like parseAttributes?
And may be return an error? Or it should be silent?

devlink_linux.go
```diff
@@ -0,0 +1,167 @@
...
+		return nil, err
+	}
+	devices, err := parseDevLinkDeviceList(msgs)
+	if err != nil {
+		return nil, err
+	}
+	for _, d := range devices {
+		h.getEswitchAttrs(f, d)
+	}
+	return devices, err
```
callmehollow: I think return should look like this
`return devices, nil`

devlink_linux.go
```diff
@@ -0,0 +1,167 @@
...
+// DevlinkDevice represents device and its attributes
+type DevlinkDevice struct {
+	BusName    string
+	DeviceName string
+	Attrs      DevlinkDevAttrs
+}
+
+func parseDevLinkDeviceList(msgs [][]byte) ([]*DevlinkDevice, error) {
+	devices := make([]*DevlinkDevice, 0, len(msgs))
+	for _, m := range msgs {
```
callmehollow: err var inits 2 times every loop

devlink_linux.go
```diff
@@ -0,0 +1,167 @@
...
+
+	m := msgs[0]
+	attrs, err := nl.ParseRouteAttr(m[nl.SizeofGenlmsg:])
+	if err != nil {
+		return
+	}
+	dev.parseAttributes(attrs)
+}
+
+func (h *Handle) getEswitchAttrs(family *GenlFamily, dev *DevlinkDevice) {
```
callmehollow: should getEswitchAttrs be silent?

devlink_linux.go
paravmellanox: @callmehollow removed.

devlink_linux.go
paravmellanox: @callmehollow yes, because devices which doesn't support will return error unsupported. Typically when sriov is disabled they return unsupported, once its enabled they operate in switchdev or legacy mode. so in such case no eswitch attributes are initialized. So getEswitchAttrs() doesn't return an error code. May be if error code other than "unsupported" is returned, than this function can return error if needed.

devlink_linux.go
paravmellanox: @callmehollow yes, make sense. changing it.

devlink_linux.go
paravmellanox: @callmehollow now made it a method of *DevlinkDevice.

devlink_linux.go
paravmellanox: @callmehollow removed second init now.

devlink_linux.go
```diff
@@ -0,0 +1,166 @@
...
+	for _, a := range attrs {
+		switch a.Attr.Type {
+		case nl.DEVLINK_ATTR_BUS_NAME:
+			d.BusName = string(a.Value)
+		case nl.DEVLINK_ATTR_DEV_NAME:
+			d.DeviceName = string(a.Value)
+		case nl.DEVLINK_ATTR_ESWITCH_MODE:
+			d.Attrs.Eswitch.Mode = parseEswitchMode(native.Uint16(a.Value))
+		case nl.DEVLINK_ATTR_ESWITCH_INLINE_MODE:
+			d.Attrs.Eswitch.InlineMode = parseEswitchInlineMode(uint8(a.Value[0]))
```
fcrisciani: do you need checks on a.Value length?

nl/devlink_linux.go
```diff
@@ -0,0 +1,36 @@
+package nl
+
+const (
```
fcrisciani: can you put a reference of the linux kernel from where these values are defined?

devlink_linux.go
```diff
@@ -0,0 +1,166 @@
...
+		dev := &DevlinkDevice{}
+		if err = dev.parseAttributes(attrs); err != nil {
+			return nil, err
+		}
+		devices = append(devices, dev)
+	}
+	return devices, nil
+}
+
+func parseEswitchMode(mode uint16) string {
```
fcrisciani: Would be great, if possible, to change this into a map that capture all the possible modes:
```
var L4ProtoMap = map[uint16]string{
	nl.DEVLINK_ESWITCH_MODE_LEGACY:  "legacy",
	nl.DEVLINK_ESWITCH_MODE_SWITCHDEV: "switchdev",
        [...]
```

devlink_linux.go
```diff
@@ -0,0 +1,166 @@
...
+	case nl.DEVLINK_ESWITCH_MODE_LEGACY:
+		return "legacy"
+	case nl.DEVLINK_ESWITCH_MODE_SWITCHDEV:
+		return "switchdev"
+	default:
+		return "unknown"
+	}
+}
+
+func parseEswitchInlineMode(inlinemode uint8) string {
```
fcrisciani: same for this

devlink_linux.go
fcrisciani: If it is a get maybe a better function prototype would be:
`func (h *Handle) getEswitchAttrs(family *GenlFamily) (*DevlinkDevice, error)`

devlink_test.go
```diff
@@ -0,0 +1,34 @@
...
+
+package netlink
+
+import (
+	"io/ioutil"
+	"strings"
+	"testing"
+)
+
+func setupDevlinkKModule(t *testing.T, name string) {
```
fcrisciani: use this method: https://github.com/vishvananda/netlink/blob/master/netlink_test.go#L118

paravmellanox: @fcrisciani done

devlink_linux.go
paravmellanox: @fcrisciani the current function signature of getEswitchAttrs() is this way, because DevlinkDevice consist of few members and additional eswitch attributes and in future more attributes will be added.
So scope of geteswitchAttrs() is to just update those fields. I am not sure alternatively returning/copying only EswitchAttr is worth instead of passing DevlinkDevice pointer.

devlink_linux.go
paravmellanox: @fcrisciani I am not sure. This is coming from kernel. Should we stop at just the length, or also check that it contains just the number? I was thinking we can trust the kernel here to return sane values of sane size? What do you think?

nl/devlink_linux.go
paravmellanox: @fcrisciani sure, reference in code or in commit log? for example nl/tc_linux.go doesn't have any reference for value definition for TCA_POLICE_TBF etc.
But that may be old. I can add the reference to kernel uapi header file. Let me know.

devlink_linux.go
paravmellanox: @fcrisciani , yes, this is possible, if we dont' find it in map, will continue to return unknown in else condition, so that when tomorrow if kernel adds new option, we can still return unknown.

devlink_linux.go
fcrisciani: I was more concerned if there is some parsing issue will crash here

nl/devlink_linux.go
fcrisciani: look at https://github.com/vishvananda/netlink/blob/master/nl/conntrack_linux.go, I guess there is value in having something like that so that there is a clear reference on where the values are coming from

devlink_linux.go
fcrisciani: sure

nl/devlink_linux.go
paravmellanox: @fcrisciani sure. I will add it.

paravmellanox:
@fcrisciani let me know if comments I fixed looks ok. I like to push more APIs post this pull request sometime next week.

paravmellanox:
@fcrisciani @vishvananda 
ping.
Also not sure, why there are CI failures in addr_test.go. This PR is not touching addr_test.go.

aboch:
Ping @fcrisciani

fcrisciani:
@paravmellanox 
the failure that I see is:
```
--- FAIL: TestNeighSubscribe (0.49s)
	neigh_test.go:317: Add update not received as expected: want 28 got 0
```

aboch:
Please rebase to pick https://github.com/vishvananda/netlink/commit/78a3099b7080a21527eb1e172546da32bee38f2a

it should make that test failure go away

fcrisciani:
@aboch if the test failure is not related I'm good with merging this one

aboch:
>Pull requests that have a failing status can’t be merged on a phone.

:)

paravmellanox:
> 
> 
> > Pull requests that have a failing status can’t be merged on a phone.
> 
> :)

@aboch @fcrisciani I will rebase in few hours and push again the same code.

paravmellanox:
@aboch @fcrisciani, ok. travis-ci looks good now.

PR #405, opened 2018-11-20
Make test suite more deterministic

corny:
expectNeighUpdate might fail on unexpected updates:
```
=== RUN   TestNeighSubscribeWithOptions
--- FAIL: TestNeighSubscribeWithOptions (0.22s)
	neigh_test.go:371: Add update not received as expected: want 28 got 0
	neigh_test.go:337: Fatal error received during subscription: Link not found
```

### DIFF
--- a/neigh_test.go
+++ b/neigh_test.go
@@ -3,7 +3,6 @@
 package netlink
 
 import (
-	"fmt"
 	"net"
 	"testing"
 	"time"
@@ -257,26 +256,17 @@ func TestNeighAddDelProxy(t *testing.T) {
 	}
 }
 
-func expectNeighUpdate(ch <-chan NeighUpdate, t uint16, state int, ip net.IP) error {
+// expectNeighUpdate returns whether the expected updated is received within one second.
+func expectNeighUpdate(ch <-chan NeighUpdate, t uint16, state int, ip net.IP) bool {
 	for {
 		timeout := time.After(time.Second)
 		select {
 		case update := <-ch:
-			if update.Type != t {
-				return fmt.Errorf("want %d got %d", t, update.Type)
+			if update.Type == t && update.Neigh.State == state && update.Neigh.IP != nil && update.Neigh.IP.Equal(ip) {
+				return true
 			}
-			if update.Neigh.State != state {
-				return fmt.Errorf("want %d got %d", state, update.State)
-			}
-			if update.Neigh.IP == nil {
-				return fmt.Errorf("Neigh.IP is nil")
-			}
-			if !update.Neigh.IP.Equal(ip) {
-				return fmt.Errorf("Neigh.IP: want %s got %s", ip, update.Neigh.IP)
-			}
-			return nil
 		case <-timeout:
-			return fmt.Errorf("timeout")
+			return false
 		}
 	}
 }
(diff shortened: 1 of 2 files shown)


neigh_test.go
```diff
@@ -257,26 +257,30 @@ func TestNeighAddDelProxy(t *testing.T) {
...
 			}
-			return nil
+			err = nil
+			return
 		case <-timeout:
-			return fmt.Errorf("timeout")
+			if err == nil {
+				err = fmt.Errorf("timeout")
+			}
+			return
```
aboch: What does using a naked return give us here ?
Besides the error msg improvement, I'd prefer the direct `return fmt.Errorf()` on error.

neigh_test.go
```diff
@@ -257,26 +257,30 @@ func TestNeighAddDelProxy(t *testing.T) {
...
 			if !update.Neigh.IP.Equal(ip) {
-				return fmt.Errorf("Neigh.IP: want %s got %s", ip, update.Neigh.IP)
+				err = fmt.Errorf("Neigh.IP: want %s got %s", ip, update.Neigh.IP)
 			}
-			return nil
+			err = nil
+			return
 		case <-timeout:
-			return fmt.Errorf("timeout")
+			if err == nil {
```
aboch: Can we even get here with a `err != nil` ?

neigh_test.go
```diff
@@ -257,26 +257,30 @@ func TestNeighAddDelProxy(t *testing.T) {
...
 			if update.Neigh.IP == nil {
-				return fmt.Errorf("Neigh.IP is nil")
+				err = fmt.Errorf("Neigh.IP is nil")
 			}
 			if !update.Neigh.IP.Equal(ip) {
-				return fmt.Errorf("Neigh.IP: want %s got %s", ip, update.Neigh.IP)
+				err = fmt.Errorf("Neigh.IP: want %s got %s", ip, update.Neigh.IP)
 			}
-			return nil
+			err = nil
```
fcrisciani: what's the point to set err if then is put to nil?

corny:
I forgot to add `else` keywords. Please check out my changes.

neigh_test.go
corny: I fixed the if/else structure.

corny:
Instead of returning an error we can simply return a boolean the the expected updates is not receiceived in time. But this could make debugging more difficult.

aboch:
I think I get it now: This function was broken from the beginning because it was assuming that the only update it would receive is the one the caller is looking for. While in reality while waiting for a link/addr/route/... programming feedback, other events are likely to come (entry purged,...).

Written this way, IMO this function is no use, because it will always fail when the first event is not the one we are looking for.

I believe @corny 's change is fixing this logic. But I'd instead prefer to rewrite the code to something like:
```
for {
    select {
        case update := <-ch:
            if update.Type == t && update.Neigh.State == state &&  ip.Equal(update.Neigh.IP) {
                return nil
            }
        case <-time.After(time.Second):
            return fmt.Errorf("timeout")
    }
}
```

corny:
In that case the function better returns a boolean exactly like `expectAddrUpdate` and `expectRouteUpdate` do.

aboch:
> In that case the function better returns a boolean exactly like expectAddrUpdate and expectRouteUpdate do.

Agree

aboch:
@corny are you still pursuing this?

corny:
I've updated my pull request.

aboch:
LGTM

corny:
Thanks!

Issue #243, opened 2017-07-05
Build fails on Mac because of undefined variables


kklin:
```
$ go build .
# github.com/vishvananda/netlink
./filter.go:189: undefined: nl.TC_U32_TERMINAL
./filter.go:190: undefined: nl.TC_U32_OFFSET
./filter.go:191: undefined: nl.TC_U32_VAROFFSET
./filter.go:192: undefined: nl.TC_U32_EAT
```

It looks like the errors are because the constants are defined in a `_linux` file: https://github.com/vishvananda/netlink/blob/master/nl/tc_linux.go#L471.

serngawy:
I have the same issue, does anyone had a fix for it

supernomad:
I have the same issue, but when attempting to compile on FreeBSD.

trevrosen:
I moved the offending lines from `filter.go` into `filter_linux.go` and all was well. I'm going to PR.

PR #406, opened 2018-11-30
Replace redundant copied u32 types with type aliases

tfukushima:
This patch replaces TcU32Sel and TcU32Key that are copied from nl
package with type aliases for the original types. This eliminates the
usages of unsafe packages and redundant copied structs that are
identical to the original ones.

Type aliases are newly introduced in Go 1.9 and it is not backward
compatibile. Therefore this patch shall be merged only if the project is
entirely migrated to Go 1.9. Currently we are building this project
with Go 1.10 and Go 1.11, so it is reasonable to require Go version
greater than 1.9.

See #206 and #261, which is the previous life of this patch.

This patch also moves TcU32Sel, TcU32Key and U32 structs from filter.go
to filter_linux.go to make it possible to build on macOS. See #243 for
more details.

Signed-off-by: Taku Fukushima <taku@soracom.jp>

### DIFF
--- a/filter.go
+++ b/filter.go
@@ -182,51 +182,6 @@ func NewMirredAction(redirIndex int) *MirredAction {
 	}
 }
 
-// Sel of the U32 filters that contains multiple TcU32Key. This is the copy
-// and the frontend representation of nl.TcU32Sel. It is serialized into canonical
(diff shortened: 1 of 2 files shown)


aboch:
LGTM

PR #410, opened 2018-12-18
neigh_linux: Fix failure on deleted link neighs updates

### LINKED ISSUE #409: NeighSubscribe: Error on link deletion
When subscribing to neigh updates, on link deletion, an error is raised, and channel gets closed.

Error:
```golang
(netlink.LinkNotFoundError) {
 error: (*errors.errorString)(0xc000010030)({
  s: (string) (len=14) "Link not found"
 })
}
```
I suspect this line:
https://github.com/vishvananda/netlink/blob/093e80f9fa21ff7b6d400d4cc54933558fd446a4/neigh_linux.go#L267

As the link no longer exists

diconico07:
The kernel sends a RTM_DELNEIGH for every neighbours on link
deletion by the time the message is deserialized, the interface
no longer exists so we cannot call LinkByIndex on it.

This call to LinkByIndex is only used to get the encapType to be
able to set either IP or HardwareAddr correctly. The attrLen
attribute can be used here as only ipv4 are used with a size of 4,
and only ipv6 and FireWire HWaddr have a size of 16.

As such this change decrease the number of calls to LinkByIndex,
so it is called only when needed to choose between ipv6 or
FireWire Hwaddr, it also fallback to HWaddr in case of error with
LinkByIndex.

Fix: 921f7441f1ad68ebc6bbebe00664dfad83a7dbc6
Fix #409

Signed-off-by: Nicolas Belouin <nicolas.belouin@gandi.net>

### DIFF
--- a/neigh_linux.go
+++ b/neigh_linux.go
@@ -253,14 +253,6 @@ func NeighDeserialize(m []byte) (*Neigh, error) {
 		return nil, err
 	}
 
-	// This should be cached for perfomance
-	// once per table dump
-	link, err := LinkByIndex(neigh.LinkIndex)
-	if err != nil {
-		return nil, err
-	}
-	encapType := link.Attrs().EncapType
-
 	for _, attr := range attrs {
 		switch attr.Attr.Type {
 		case NDA_DST:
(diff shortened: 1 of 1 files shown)


vishvananda:
Nice find. LGTM

PR #411, opened 2018-12-20
WIP: Add support for group_fwd_mask

cloudtoad:
group_fwd_mask is a bitmask that can enable a bridge to forward link-local frames like LLDP.  We are building virtual network topologies using containers/images of network switches (like Juniper's vMX or vQFX).  LLDP is a useful troubleshooting tool and is useful for automation related tasks.

# TODO

- [ ] Need to write unit test... will use MulticastSnooping as a template.

### DIFF
--- a/link_linux.go
+++ b/link_linux.go
@@ -229,6 +229,21 @@ func (h *Handle) macvlanMACAddrChange(link Link, addrs []net.HardwareAddr, mode 
 	return err
 }
 
+
+// LinkSetHardwareAddr sets the hardware address of the link device.
+// Equivalent to: `ip link set $link address $hwaddr`
+func BridgeSetGroupFwdMask(link Link, mask uint16) error {
+	return pkgHandle.BridgeSetGroupFwdMask(link, mask)
+}
+
+// LinkSetHardwareAddr sets the hardware address of the link device.
+// Equivalent to: `ip link set $link address $hwaddr`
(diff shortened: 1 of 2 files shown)


link_linux.go
```diff
@@ -229,6 +229,21 @@ func (h *Handle) macvlanMACAddrChange(link Link, addrs []net.HardwareAddr, mode
...
+// LinkSetHardwareAddr sets the hardware address of the link device.
+// Equivalent to: `ip link set $link address $hwaddr`
+func BridgeSetGroupFwdMask(link Link, mask uint16) error {
+	return pkgHandle.BridgeSetGroupFwdMask(link, mask)
+}
+
+// LinkSetHardwareAddr sets the hardware address of the link device.
+// Equivalent to: `ip link set $link address $hwaddr`
+func (h *Handle) BridgeSetGroupFwdMask(link Link, mask uint16) error {
+	bridge := link.(*Bridge)
```
aboch: what if it is not of type Bridge ?

cloudtoad: That's a good question.  Note that I just copied/pasted the MulticastSnooping functions.  So those don't have a check for that either.  If I add one here, should it also be added to that function?

vishvananda:
This has not been updated in a while. If you wish to update it with the mentioned unit test, please reopen.

PR #412, opened 2018-12-21
Add bridge dump support

diconico07:
This PR add support for bridge neigh entries dump on subscribe and also add a bridge related attribute to neigh structure

### DIFF
--- a/neigh.go
+++ b/neigh.go
@@ -17,6 +17,7 @@ type Neigh struct {
 	LLIPAddr     net.IP //Used in the case of NHRP
 	Vlan         int
 	VNI          int
+	MasterIndex  int
 }
 
 // String returns $ip/$hwaddr $label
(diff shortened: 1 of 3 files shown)


diconico07:
Added support of MasterIndex in neighHandle() as I forgot it at first.

vishvananda:
The code here looks correct. Is it possible to create a test to validate that it is working properly?

diconico07:
I had to change a bit more than I thought for the tests. The only disappointing thing is the need for a vxlan interface in the test as the FDB entries for vxlan interfaces are really close to usual neigh entries and can thus be checked using the same function in the tests.

diconico07:
Anything else needed from me ?

aboch:
Please squash your commits into one

aboch:
LGTM

PR #418, opened 2019-01-03
Fix Race Condition in TestXfrmMonitorExpire

seglberg:
This fixes the TestXfrmMonitorExpire test, which previously exhibited a race-condition. 

Since netlink messages are asynchronous and order is not guaranteed by the Kernel, the order of the expire messages could not be assumed.

---
This change is [Reviewable](https://reviewable.io/reviews/vishvananda/netlink/418)


aboch:
Thanks.

LGTM

PR #425, opened 2019-01-22
Add bond slave information

lebauce:


lebauce:
@vishvananda Can you please take a look ? Thanks !

vishvananda:
This code looks correct, but it would be great to have a test for it. Is it possible to add a bond slave already? If so, a test should be straightforward. If not, then that is probably functionality we need as well.

aboch:
ping @lebauce

aboch:
Sorry, no feedback for the past 5 months. Will go with #487

lebauce:
@aboch Sorry, missed the notifications. Thanks for finishing my work :-)

PR #444, opened 2019-05-16
tuntap: Return TunTapLink instead of GenericLink

amshinde:
For tuntap interfaces, return a TunTap Interface instead of
a Generic link when retrieving the interface.
Use netlink extended attributes to populate the Link attributes
for the tuntap link.
In case of older tun driver which does not provide these
attributes, use sysfs to retrieve these attributes.

This commit also adds Owner and Group attributes for the TunTap
Link.

Signed-off-by: Archana Shinde <archana.m.shinde@intel.com>


link_linux.go
```diff
@@ -1580,9 +1586,57 @@ func LinkDeserialize(hdr *unix.NlMsghdr, m []byte) (Link, error) {
...
+
+				if flags&unix.IFF_PERSIST == 0 {
+					tuntap.NonPersist = true
+				} else {
+					tuntap.NonPersist = false
+				}
+
+			}
+			if owner, err := readSysProp(ifname, "owner"); err == nil && owner > 0 {
+				tuntap.Owner = uint32(owner)
```
jellonek: Hmmm... Are we loosing there info that the device have no owner, instead passing an info that the owner is root? The same goes for `group`.

link_linux.go
```diff
@@ -1580,9 +1586,57 @@ func LinkDeserialize(hdr *unix.NlMsghdr, m []byte) (Link, error) {
...
+			if group, err := readSysProp(ifname, "group"); err == nil && group > 0 {
+				tuntap.Group = uint32(group)
+			}
+		}
+	}
+
 	return link, nil
 }
 
+func readSysProp(ifname, prop string) (int64, error) {
```
jellonek: Please rename this func to `readSysPropAsInt64`.

link_linux.go
amshinde: If the owner is root, I noticed that the sysfs interface returns -1 instead of 0 here. That's why the workaround. The sysfs interface probably needs to be modified to return 0 for root instead of -1.

link_linux.go
amshinde: Will do.

link_linux.go
jellonek: ACK.

link_linux.go
amshinde: Fixed.

amshinde:
@jellonek I have added a test to check for tuntap attributes.
For older kernels, since netlink does not provide the tun attributes, we rely on sysfs. 
But the sysfs needs to be remounted after the network namespace is created. I think this should be the responsibility of the caller.
The unit tests handle this by mounting and unmounting sysfs.

amshinde:
@jellonek I am seeing Travis failures that are completele unrelated to this PR.
Is this expected or something I should be looking at?

jellonek:
@vishvananda already fixed some of these errors (maybe even all of them) but i don't have a rights to restart test job.

btw. Iv'e found a small issue with your testing procedure - imo unmounting sys should be added as a defered call just after mount, before calling the inner test.

vishvananda:
@jellonek I fixed the errors and merged yesterday in #447. Can you try rebasing on master and repushing? The sys unmount seems reasonable.

amshinde:
@jellonek I have fixed the sys umount now. PTAL. Rebased as well.

jellonek:
@amshinde LGTM
@vishvananda i was only mentioning that YOU have fixed these bugs ;) (sry for my english ;) )

link_linux.go
```diff
@@ -1580,9 +1586,57 @@ func LinkDeserialize(hdr *unix.NlMsghdr, m []byte) (Link, error) {
...
 	*link.Attrs() = base
 
+	// If the tuntap attributes are not updated by netlink due to
+	// an older driver, use sysfs
+	if link != nil && linkType == "tun" {
+		tuntap := link.(*Tuntap)
+
+		if tuntap.Mode == 0 {
+			ifname := tuntap.Attrs().Name
+			if flags, err := readSysPropAsInt64(ifname, "tun_flags"); err == nil {
```
aboch: Why aren't we handling the error ?

link_linux.go
```diff
@@ -1580,9 +1586,57 @@ func LinkDeserialize(hdr *unix.NlMsghdr, m []byte) (Link, error) {
...
+
+				if (flags & unix.IFF_TUN) != 0 {
+					tuntap.Mode = unix.IFF_TUN
+				} else if (flags & unix.IFF_TAP) != 0 {
+					tuntap.Mode = unix.IFF_TAP
+				}
+
+				if flags&unix.IFF_PERSIST == 0 {
```
aboch: please `go fmt` this.

aboch: Why here `if A & B == 0` and above `if (A&B) != 0` ?
Please be consistent, if parenthesis are not needed, do not add them

link_linux.go
aboch: Then please add this explanation as a comment.

link_linux.go
```diff
@@ -1580,9 +1586,57 @@ func LinkDeserialize(hdr *unix.NlMsghdr, m []byte) (Link, error) {
...
+					tuntap.NonPersist = true
+				} else {
+					tuntap.NonPersist = false
+				}
+
+			}
+			if owner, err := readSysPropAsInt64(ifname, "owner"); err == nil && owner > 0 {
+				tuntap.Owner = uint32(owner)
+			}
+			if group, err := readSysPropAsInt64(ifname, "group"); err == nil && group > 0 {
```
aboch: We should not discard the error.

link_linux.go
```diff
@@ -1580,9 +1586,57 @@ func LinkDeserialize(hdr *unix.NlMsghdr, m []byte) (Link, error) {
...
+				tuntap.Group = uint32(group)
+			}
+		}
+	}
+
 	return link, nil
 }
 
+func readSysPropAsInt64(ifname, prop string) (int64, error) {
+	var num int64
```
aboch: Please remove this definition `strconv.ParseInt` returns an `int64`
`num, err := strconv.ParseInt(...`

link_linux.go
```diff
@@ -1580,9 +1586,57 @@ func LinkDeserialize(hdr *unix.NlMsghdr, m []byte) (Link, error) {
...
+
+	fname := fmt.Sprintf("/sys/class/net/%s/%s", ifname, prop)
+
+	contents, err := ioutil.ReadFile(fname)
+	if err != nil {
+		return num, err
```
aboch: on same line to keep code lean, after taking care of previous comment, it is perfectly fine to`return 0, err` here.

link_linux.go
```diff
@@ -1580,9 +1586,57 @@ func LinkDeserialize(hdr *unix.NlMsghdr, m []byte) (Link, error) {
...
+				if (flags & unix.IFF_TUN) != 0 {
+					tuntap.Mode = unix.IFF_TUN
+				} else if (flags & unix.IFF_TAP) != 0 {
+					tuntap.Mode = unix.IFF_TAP
+				}
+
+				if flags&unix.IFF_PERSIST == 0 {
+					tuntap.NonPersist = true
+				} else {
+					tuntap.NonPersist = false
```
aboch: I feel the following form is better readable
```
tuntap.NonPersist = false
if flags&unix.IFF_PERSIST == 0 {
    tuntap.NonPersist = true
}
```

link_linux.go
amshinde: @aboch I chose to ignore this error, in case sysfs has not been mounted for the network namespace, resulting in interfaces not found on /sys.
I took the approach of best-effort to get the interface attributes. If you prefer erroring out here, 
I can follow that approach.

aboch: >I took the approach of best-effort to get the interface attributes.

Makes sense. Thanks.

amshinde: Thanks @aboch. I have addressed your other comments. PTAL.

link_linux.go
aboch: best effort logic as explained by @amshinde

aboch:
LGTM

PR #446, opened 2019-05-29
Revert "Add support for TC_ACT_CONNMARK"

vishvananda:
This reverts commit 00009fb8606a3b58605059db3a56262f763299d4.

This is a test revert to see if the connmark patch broke CI, or if it is an unrelated faliure due to the ci image changing


vishvananda:
Ok so this is a testing error. It looks like ClassAdd with the HtbClass is now failing. Perhaps the add now creates multiple classes due to a kernel change

PR #447, opened 2019-05-29
Fix tests by removing duplicate classes

vishvananda:
In some kernels, creating a class and then listing will actually return multiple copies of the added class. Fix the broken tests to work around this by using a class list function that removes duplicates.

### DIFF
--- a/class_test.go
+++ b/class_test.go
@@ -102,7 +102,7 @@ func TestClassAddDel(t *testing.T) {
 	if err := ClassAdd(class); err != nil {
 		t.Fatal(err)
 	}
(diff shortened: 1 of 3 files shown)


vishvananda:
@aboch @fcrisciani This gets the Travis tests working again, although i had to disable a few tests that were broken. I've created two issues to track fixing the disabled tests. If either of you has time to take a look at them feel free. I need to go create a vm with the proper kernel version to try to reproduce. You can see the failure messages in travis. I suggest we go ahead and merge this to get things unblocked.

aboch:
LGTM

fcrisciani:
LGTM

PR #450, opened 2019-05-31
Mirred and connmark clobber their ActionAttrs

kmjohansen:
Encountered this in a local test.  It turns out that in parseActions
mirred has a bug where it parses the action attributes but then on the
very next line overwrites this hard work by assigning an empty
ActionAttrs struct on top.  I copy pasta'd this into connmark.  Fix both
instances and amend the unit tests to catch this going forward.

Signed-off-by: Krister Johansen <krister.johansen@oracle.com>


kmjohansen:
@vishvananda @aboch @fcrisciani This one is a quick fix for a bug in the mirred action attribute parsing that I accidentally duplicated into the connmark action parsing.  This fixes both cases and updates the unit tests.  Apologies for missing this one on the first go round.

kmjohansen:
Thanks, gents

PR #451, opened 2019-06-01
Fix unit test failure: TestNeighAddDelLLIPAddr

mcastelino:
TestNeighAddDelLLIPAddr was failing due to the Neighbour table
not getting properly populated when using a ipip tunnel.
This matches the behaviour in the latest kernel when using
the ip command.

Switch the tunnel type to a gre point to multi-point tunnel.
The neighbour table gets properly populated in this case.


mcastelino:
The travis test failure for go 1.10.x seems to be unrelated to this PR.

aboch:
LGTM

PR #458, opened 2019-06-10
Allow replacing filters

lmb:
Add a function FilterReplace, which mirrors the behaviour of
QdiscReplace, etc. This makes it possible to swap out filters
with a single netlink message.

### DIFF
--- a/filter_linux.go
+++ b/filter_linux.go
@@ -151,8 +151,24 @@ func FilterAdd(filter Filter) error {
 // FilterAdd will add a filter to the system.
 // Equivalent to: `tc filter add $filter`
 func (h *Handle) FilterAdd(filter Filter) error {
+	return h.filterModify(filter, unix.NLM_F_CREATE|unix.NLM_F_EXCL)
+}
+
+// FilterReplace will replace a filter.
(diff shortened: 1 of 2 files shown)


filter_linux.go
```diff
@@ -151,8 +151,24 @@ func FilterAdd(filter Filter) error {
...
+// FilterReplace will replace a filter.
+// Equivalent to: `tc filter replace $filter`
+func FilterReplace(filter Filter) error {
+	return pkgHandle.FilterReplace(filter)
+}
+
+// FilterReplace will replace a filter.
+// Equivalent to: `tc filter replace $filter`
+func (h *Handle) FilterReplace(filter Filter) error {
+	return h.filterModify(filter, unix.NLM_F_CREATE)
```
aboch: Are we sure we should not use `unix.NLM_F_REPLACE` ?

lmb:
Thanks for taking a look!


This matches what iproute2 does: 
https://github.com/shemminger/iproute2/blob/afa588490b7e87c5adfb05d5163074e20b6ff14a/tc/tc_filter.c#L766

aboch:
LGTM

PR #460, opened 2019-06-17
veth: Set peer hardware addr when creating

brb:
This PR extends the `LinkAdd()` function for `Veth` by allowing to specify peer hardware address.


aboch:
LGTM

PR #461, opened 2019-06-17
Read connection marking information from CT flow TLV

eriknordmark:
The current conntrack code tries to read connection marking information by reading at a fixed offset (different for UDP and TCP) of CT module TLV response. This offset based read does not seem to be correct. Similar to how original/reply flows are being read by parsing the TLV response, we look for TLV type CTA_MARK and read the corresponding value as connection marking.


aboch:
@eriknordmark I know it was not there already, but could you please add/enhance existing tests to test the mark setting.
Thnaks

aboch:
LGTM

PR #466, opened 2019-07-09
Add support for setting VLAN QoS for VF links

przemeklal:
This change adds support for setting VLAN QoS (priority) field for the
SR-IOV Virtual Function links.


przemeklal:
Hey @aboch I'd love to, however it's impossible to write tests for the functions that depend on the hardware (SR-IOV VF in this case). Other SR-IOV related functions don't have any tests implemented due to this fact. Note that this function is basically a copy of the `LinkSetVfVlan` with the only change being `Qos` field added in the `VfVlan` struct.

PR #468, opened 2019-07-15
netlink: enforce similar pid checks as in iproute2

borkmann:
iproute2's own netlink library asserts that the sockaddr sender pid
has to be the one of the kernel [0]. It also doesn't bail out on pid
mismatch but only skips the message instead. We've seen cases where
the latter had a pid 0; in such case we should skip to the next nl
message instead of hard bail out.

  [0] https://git.kernel.org/pub/scm/network/iproute2/iproute2.git/tree/lib/libnetlink.c
      rtnl_dump_filter_l() and __rtnl_talk_iov()

Signed-off-by: Daniel Borkmann <daniel@iogearbox.net>

### DIFF
--- a/nl/nl_linux.go
+++ b/nl/nl_linux.go
@@ -26,6 +26,8 @@ const (
 	// from kernel more verbose messages e.g. for statistics,
 	// tc rules or filters, or other more memory requiring data.
 	RECEIVE_BUFFER_SIZE = 65536
+	// Kernel netlink pid
+	PidKernel uint32 = 0
 )
 
 // SupportedNlFamilies contains the list of netlink families this netlink package supports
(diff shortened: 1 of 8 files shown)


nl/nl_linux.go
```diff
@@ -617,22 +622,28 @@ func (s *NetlinkSocket) Send(request *NetlinkRequest) error {
...
 	}
 	if nr < unix.NLMSG_HDRLEN {
-		return nil, fmt.Errorf("Got short response from netlink")
+		return nil, nil, fmt.Errorf("Got short response from netlink")
 	}
 	rb2 := make([]byte, nr)
 	copy(rb2, rb[:nr])
-	return syscall.ParseNetlinkMessage(rb2)
+	nl, err := syscall.ParseNetlinkMessage(rb2)
+	return nl, fromAddr, err
```
jellonek: It's a good practice to return zero values when `err` is non nil, so please add an `if` after this syscall invocation as in line 633.

borkmann: Makes sense, done, thanks!

PR #469, opened 2019-07-16
Add support for IFLA_VF_RATE

adrianchiris:
Today netlink package supports Get/Set of a VF's max TX rate
via IFLA_VF_TX_RATE netlink attribute.

This patch adds support to Get/Set of a VFs min and max TX rate
via IFLA_VF_RATE netlink attribute.

- Add support to set min/max tx rate for VF via IFLA_VF_RATE
- Added IFLA_VF_RATE min/max tx rate attributes to netlink.VfInfo
  including parsing support in netlink.parseVfInfo()

NOTE: According to [1] IFLA_VF_RATE takes precedence over
      IFLA_VF_TX_RATE. Dealing with the coexistence of these
      netlink attributes is left for the user to handle.

[1]https://lists.openwall.net/netdev/2014/05/22/42

### DIFF
--- a/link.go
+++ b/link.go
@@ -53,9 +53,11 @@ type VfInfo struct {
 	Mac       net.HardwareAddr
 	Vlan      int
 	Qos       int
-	TxRate    int
+	TxRate    int // IFLA_VF_TX_RATE  Max TxRate
 	Spoofchk  bool
 	LinkState uint32
+	MaxTxRate uint32 // IFLA_VF_RATE Max TxRate
+	MinTxRate uint32 // IFLA_VF_RATE Min TxRate
 }
 
 // LinkOperState represents the values of the IFLA_OPERSTATE link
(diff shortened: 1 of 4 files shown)


link.go
```diff
@@ -53,9 +53,11 @@ type VfInfo struct {
 	Mac       net.HardwareAddr
 	Vlan      int
 	Qos       int
-	TxRate    int
+	TxRate    int // IFLA_VF_TX_RATE  Max TxRate
```
jellonek: This looks like some remaining after refactor, as max tx rate is defined also below and later in code only `MaxTxRate` is used, without no further mention of plain `TxRate`.
Am I correct there?

adrianchiris: The comment is here to clarify that TxRate maps to VF_TX_RATE netlink attribute.
This commit added MinTxRate and MaxTxRate which map to VF_RATE netlink attribute.

jellonek: ACK. Now it's clear for me.

aboch:
LGTM

Issue #329, opened 2018-02-12
AddrAdd shouldn't add broadcast address if not directly asked to


jjastrze-ovh:
Let's consider following code:
```go
	link := &netlink.Dummy{
		LinkAttrs: netlink.LinkAttrs{
			Flags: net.FlagUp,
			Name:  "ntest",
		},
	}

	if err := netlink.LinkAdd(link); err != nil {
		fmt.Println("add", err)
		return
	}

	addr := &netlink.Addr{
		IPNet: &net.IPNet{
			IP:   net.ParseIP("192.168.22.19"),
			Mask: net.CIDRMask(24, 32),
		},
	}

	if err := netlink.AddrAdd(link, addr); err != nil {
		fmt.Println("add addr", err)
		return
	}
```
The code should add **192.168.22.19** address to device **ntest**, but it also adds broadcast address, which is some scenarios breaks eg. adding routes with error=invalid argument.

```bash
[root@machine] ip -4 a s ntest
6488: ntest: <BROADCAST,NOARP,UP,LOWER_UP> mtu 1500 qdisc noqueue state UNKNOWN group default 
    inet 192.168.22.19/24 brd 192.168.22.255 scope global ntest
       valid_lft forever preferred_lft forever
```

For me presented code should be equivalent to what following ip command does:
```bash
[root@root@machine] ip l a ntest type dummy
[root@root@machine] ip a a 192.168.22.19/24 dev ntest
[root@root@machine] ip -4 a s ntest
6489: ntest: <BROADCAST,NOARP> mtu 1500 qdisc noop state DOWN group default 
    inet 192.168.22.19/24 scope global ntest
       valid_lft forever preferred_lft forever
```

**Not** to this:
```bash
[root@root@machine] ip a a 192.168.22.19/24 brd 192.168.22.255 dev ntest
[root@root@machine] ip -4 a s ntest
6489: ntest: <BROADCAST,NOARP,UP,LOWER_UP> mtu 1500 qdisc noqueue state UNKNOWN group default 
    inet 192.168.22.19/24 brd 192.168.22.255 scope global ntest
       valid_lft forever preferred_lft forever
```
iptool makes end user responsibility to specify broadcast address if it is needed.
In my opinion this library should behave in the same way

The proposal for this issue is:
[...]

PR #472, opened 2019-07-22
Add AddrAddWithoutCalculatedBroadcast

### LINKED ISSUE #471: Broadcast address added to device while not explicitly requested
When I call AddrAdd(link, addr), the library currently automatically adds a broadcast address:
https://github.com/vishvananda/netlink/blob/master/addr_linux.go#L110

This can become problematic when we are using a /31 for p2p links. For example in our use case we have a GRE tunnel. If we want to assign e.g. 10.128.0.2/31 to one side, and then try to use 10.128.0.3 as gateway to route to remote tunnel, then it will report invalid gateway.

Can we have an option to not by default add brd address?

jellonek:
Added as a separate func to not introduce an API breaking change.
Closes #471

### DIFF
--- a/addr_linux.go
+++ b/addr_linux.go
@@ -15,16 +15,31 @@ import (
 const IFA_FLAGS = 0x8
 
 // AddrAdd will add an IP address to a link device.
-// Equivalent to: `ip addr add $addr dev $link`
+// Equivalent to: `ip addr add $addr dev $link` with a note:
+// If broadcast address is undefined, calculated one will be added.
 func AddrAdd(link Link, addr *Addr) error {
 	return pkgHandle.AddrAdd(link, addr)
 }
 
 // AddrAdd will add an IP address to a link device.
-// Equivalent to: `ip addr add $addr dev $link`
+// Equivalent to: `ip addr add $addr dev $link` with a note:
+// If broadcast address is undefined, calculated one will be added.
 func (h *Handle) AddrAdd(link Link, addr *Addr) error {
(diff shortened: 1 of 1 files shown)


jellonek:
@vishvananda @aboch @fcrisciani PTAL
I still thing that this should be done as a breaking API change, that broadcast calculated address should be added only when requested as an option, but that would require a new release of whole library.
WDYT?

jerryz920:
Thanks for the PR! Do you have other concerns to merge and release this one @vishvananda ?

aboch:
IMO, I believe it is worth investigating other approaches, trying hard to avoid adding such a specific method.

jerryz920:
Without adding new method then it will definitely cause API semantic change. Will it make sense to not automatically add brd address on /31 or /30?

jellonek:
IMO it makes sense, but that would be breaking change. That's why I'm asking there also @vishvananda

jellonek:
@vishvananda ping

brandt:
@aboch @jellonek I just submitted PR #496 with a different approach mirroring @dannyk81's suggestion in #329 to no longer automatically set the broadcast address on subnets too small to logically have one (/32 and /31).

As I mention in #496, another alternative – one that already works today – is to set the broadcast address to `net.IP{0, 0, 0, 0}`. However, I don't presently know if that's part of netlink's public API or if it's a quirk that could break in the future.

dannyk81:
thanks @brandt, at the time I was meaning to do a similar PR and... life got the best of me :sweat_smile:

vishvananda:
I went ahead and merged the non-breaking change so going to close this for now. Please rebase and reopen if you think that we should also have this method.

PR #476, opened 2019-08-12
Add Enums of Timestamp, Timeout to netlink/conntrack

naiming-zededa:
Signed-off-by: Naiming Shen <naiming@zededa.com>

This patch is to add Some Enums to support of Conntrack Timestamp, Timeout, and make the query flow more efficient when walk through items in the conntrack returned objects.


aboch:
LGTM

PR #491, opened 2019-09-30
Change LinkSetMaster's master argument type. (*Bridge -> Link)

hiepon:
LinkSetMaster also works with Bond device.
So this PR changes the type of master argument to Link.


aboch:
LGTM

PR #496, opened 2019-10-30
Don't auto-set broadcast unless subnet larger than /31

brandt:
Since [#248](https://github.com/vishvananda/netlink/pull/248), adding an address automatically sets the broadcast if the broadcast address was not specified. This is undesirable when adding an IP with a prefixlen of /31 or /32. (Additional details in the issues linked below.)

This PR changes the behavior so that the broadcast is only automatically set if the prefixlen is /30 or larger.

Issue reported in:

- https://github.com/vishvananda/netlink/issues/329
- https://github.com/vishvananda/netlink/issues/471

See also:

- [RFC 3021](http://tools.ietf.org/html/rfc3021)

Alternatives to this PR:

A. https://github.com/vishvananda/netlink/issues/472 - Adds `AddrAddWithoutCalculatedBroadcast`.
B. https://github.com/jjastrze-ovh/netlink/commit/9a85a619d2f74114626ca04b83b56f5a84f711df - Breaking change to make auto-setting the broadcast address an opt-in feature.
C. already works - Suppress setting the broadcast when addr's broadcast address is set to `0.0.0.0`. (This works today, but I'm not sure the behavior can be relied upon as a public API.)

### DIFF
--- a/addr_linux.go
+++ b/addr_linux.go
@@ -15,39 +15,62 @@ import (
 const IFA_FLAGS = 0x8
 
 // AddrAdd will add an IP address to a link device.
+//
 // Equivalent to: `ip addr add $addr dev $link`
+//
+// If `addr` is an IPv4 address and the broadcast address is not given, it
+// will be automatically computed based on the IP mask if /30 or larger.
 func AddrAdd(link Link, addr *Addr) error {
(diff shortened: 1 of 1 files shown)


dannyk81:
I think the approach of this PR is the optimal one.

PR #501, opened 2019-12-03
Fix LinkSetVfGUID to not reverse the GUID

Mmduh-483:
when call netlink.LinkSetVfGUID for port or node, with guid like 00:11:22:33:44:55:66:77 the actual GUID will be 77:66:55:44:33:22:11:00.

This is due to using LittleEndian instead of BigEndian

### DIFF
--- a/link_linux.go
+++ b/link_linux.go
@@ -681,7 +681,7 @@ func (h *Handle) LinkSetVfGUID(link Link, vf int, vfGuid net.HardwareAddr, guidT
 	var guid uint64
 
 	buf := bytes.NewBuffer(vfGuid)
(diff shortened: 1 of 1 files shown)


Mmduh-483:
@vishvananda @aboch , Any comments ?

Mmduh-483:
@jellonek it is not possible to add test case for this change, please check the original patch #338

adrianchiris:
seems like there was a discussion about the dpendency on kernel for testing this
https://github.com/vishvananda/netlink/pull/338#issuecomment-376373984

vishvananda:
It is still possible to write a test case and disable it in the ci system which doesn't have the required hardware, but I'll go ahead and merge this as is for now.

PR #511, opened 2020-01-17
rule: add support for dport/sport

chendotjs:
This PR adds ability equivalent to `ip rule add dport xxx-xxx sport xxx table main`

### DIFF
--- a/rule.go
+++ b/rule.go
@@ -23,6 +23,8 @@ type Rule struct {
 	SuppressIfgroup   int
 	SuppressPrefixlen int
 	Invert            bool
+	Dport             *RulePortRange
+	Sport             *RulePortRange
 }
 
 func (r Rule) String() string {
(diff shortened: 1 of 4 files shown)


rule.go
```diff
@@ -40,3 +42,14 @@ func NewRule() *Rule {
 		Flow:              -1,
 	}
 }
+
+// RulePortRange creates rule sport/dport range.
+func NewRulePortRange(start, end uint16) *RulePortRange {
```
aboch: Why not making this a method of `Rule` structure ?
```
(r *Rule) AddPortRange(start, end uint16)
```

chendotjs: Thanks for review @aboch 
`(r *Rule) AddPortRange(start, end uint16)` may need another parameter to specify it's a Src or a Dst port range, which might add more complexity.
IMO, current fields of `Rule` structure is assigned by users manually, so `NewRulePortRange` is enough.

aboch:
@chendotjs pls rebase

chendotjs:
@aboch done

aboch:
LGTM

PR #513, opened 2020-01-18
Service curves fix

fbegyn:
There was something off with the math for the ServiceCurves. The set value was 8 times too large and the returned one was 8 times too small. This is now fixed in `class.go` and `class_linux.go` so the user can just set the rate in bit and it will match the tc show output.

In light of this change I've also expanded the comments of the functions to offer more information as to how these work. Before the function gave the misleading impression that they behaved exactly like the `tc` command-line tool, which was not the case. This commit fixes this and correct the behaviour as intended.

### DIFF
--- a/class.go
+++ b/class.go
@@ -132,7 +132,10 @@ func (class *GenericClass) Type() string {
 	return class.ClassType
 }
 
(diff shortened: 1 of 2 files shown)


fcrisciani:
LGTM

PR #514, opened 2020-01-21
Use IfaCacheinfo and IFA_* consts from golang.org/x/sys/unix

tklauser:
Use the IfaCacheinfo type and the IFA_* consts from
golang.org/x/sys/unix instead of locally duplicating them.


fcrisciani:
LGTM

PR #515, opened 2020-01-23
Save XDP attach mode info into LinkXdp.

belyalov:
Currently there is no way to check how XDP program is attached (SKB, Native or HW mode)


fcrisciani:
LGTM

PR #520, opened 2020-02-05
check payload of the NLMSG_DONE nlmsg

Jianlin-lv:
check the payload of the NLMSG_DONE type netlink
message for possible error

Signed-off-by: Jianlin Lv <Jianlin.Lv@arm.com>

### DIFF
--- a/nl/nl_linux.go
+++ b/nl/nl_linux.go
@@ -439,10 +439,7 @@ done:
 			if m.Header.Pid != pid {
 				continue
 			}
-			if m.Header.Type == unix.NLMSG_DONE {
-				break done
-			}
-			if m.Header.Type == unix.NLMSG_ERROR {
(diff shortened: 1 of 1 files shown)


nl/nl_linux.go
```diff
@@ -440,6 +440,11 @@ done:
 				continue
 			}
 			if m.Header.Type == unix.NLMSG_DONE {
+				native := NativeEndian()
+				error := int32(native.Uint32(m.Data[0:4]))
+				if error < 0 {
+					return nil, syscall.Errno(-error)
+				}
```
joestringer: Maybe this if condition and the next case (`unix.NLMSG_ERROR`) can completely share the implementation by just expanding the if condition below on line 450 to include the `unix.NLMSG_DONE` case?

aboch:
Thanks @Jianlin-lv and @joestringer 

LGTM

PR #522, opened 2020-02-10
Introduce RouteGetWithOptions that allows to modify the RouteGet

amitn:
Add support for VrfName as the RouteGetOptions

Signed-off-by: Amit Nishry <amit.nishry@gmail.com>

### DIFF
--- a/route_linux.go
+++ b/route_linux.go
@@ -938,15 +938,27 @@ func deserializeRoute(m []byte) (Route, error) {
 	return route, nil
 }
 
+// RouteGetOptions contains a set of options to use with
+// RouteGetWithOptions
+type RouteGetOptions struct {
+	VrfName string
+}
+
+// RouteGetWithOptions gets a route to a specific destination from the host system.
+// Equivalent to: 'ip route get <> vrf <VrfName>'.
+func RouteGetWithOptions(destination net.IP, options *RouteGetOptions) ([]Route, error) {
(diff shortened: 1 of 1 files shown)


route_linux.go
```diff
@@ -938,15 +938,27 @@ func deserializeRoute(m []byte) (Route, error) {
...
 func RouteGet(destination net.IP) ([]Route, error) {
 	return pkgHandle.RouteGet(destination)
 }
 
-// RouteGet gets a route to a specific destination from the host system.
-// Equivalent to: 'ip route get'.
-func (h *Handle) RouteGet(destination net.IP) ([]Route, error) {
+// RouteGetWithOptions gets a route to a specific destination from the host system.
+// Equivalent to: 'ip route get <> vrf <VrfName>'.
+func (h *Handle) RouteGetWithOptions(destination net.IP, options *RouteGetOptions) ([]Route, error) {
```
aboch: I am not against this approach, was just thinking that if we add variadic options to existing function instead:

```
func (h *Handle) RouteGet(destination net.IP, options...RouteOption) ([]Route, error)
```

then we can avoid adding a new API.

amitn: totally forgot that it is an option, too much C

aboch:
Having an explicit method for querying a route with options is not that bad either.

LGTM

PR #523, opened 2020-02-10
Correct value of SizeofXfrmUsersaFlush

tklauser:
`struct xfrm_usersa_flush` contains a single `u8`, thus `sizeof(struct
xfrm_usersa_flush) == 1` as can be verified by running the following
code through `go tool cgo -godefs`:

    package xfrm_test

    // #include <linux/xfrm.h>
    import "C"

    const SizeofXfrmUsersaFlush = C.sizeof_struct_xfrm_usersa_flush

which results in

    // Code generated by cmd/cgo -godefs; DO NOT EDIT.
    // cgo -godefs foo.go

    package xfrm_test

    const SizeofXfrmUsersaFlush = 0x1


aboch:
LGTM

PR #525, opened 2020-02-18
make tuntap pollable

rschmied:
this patch allows tuntaps to be used with go1.13. It worked w/ 1.12 and before. Then failed with 1.13 resulting in `read /dev/net/tun: not pollable`.

The comment in the code has additional details. Thanks to @breakintheweb for the help to fix this.

See <https://github.com/golang/go/issues/30426> which got exposed in go 1.13 by the fix to <https://github.com/golang/go/issues/30624> .

### DIFF
--- a/link_linux.go
+++ b/link_linux.go
@@ -1089,21 +1089,52 @@ func (h *Handle) linkModify(link Link, flags int) error {
 		}
 
 		req.Flags |= uint16(tuntap.Mode)
-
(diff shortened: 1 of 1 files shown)


rschmied:
@vishvananda is there anything that I can help with to expedite this PR? Or any questions around this?

rschmied:
@vishvananda ping... thanks for considering

aboch:
LGTM

PR #534, opened 2020-04-13
Add SocketDiagTCPInfo with some constants and structs to get tcp_info…

RealFatCat:
… via sock_diag.

Probably, needs some restructurisation.


randomizedcoder:
G'day @RealFatCat , This is a really great PR.

- How did you actually test this?
- What are your thoughts in terms of keeping github.com/vishvananda/netlink in line with the kernel struct?  Some kind of kernel version detection, to map to the correct struct? e.g. Latest iproute2 has even more good stuff in the struct:
https://github.com/shemminger/iproute2/blob/afa588490b7e87c5adfb05d5163074e20b6ff14a/misc/ss.c#L807

tcp_linux.go
```diff
@@ -0,0 +1,124 @@
...
+	t.Tcpi_rwnd_limited = native.Uint64(rb.Next(8))
+	t.Tcpi_sndbuf_limited = native.Uint64(rb.Next(8))
+	t.Tcpi_delivered = native.Uint32(rb.Next(4))
+	t.Tcpi_delivered_ce = native.Uint32(rb.Next(4))
+	t.Tcpi_bytes_sent = native.Uint64(rb.Next(8))
+	t.Tcpi_bytes_retrans = native.Uint64(rb.Next(8))
+	t.Tcpi_dsack_dups = native.Uint32(rb.Next(4))
+	t.Tcpi_reord_seen = native.Uint32(rb.Next(4))
+	t.Tcpi_rcv_ooopack = native.Uint32(rb.Next(4))
+	t.Tcpi_snd_wnd = native.Uint32(rb.Next(4))
```
khasanovbi, line 122:
```suggestion
	err := binary.Read(bytes.NewReader(b), native, t)
	if err != nil {
		return fmt.Errorf("tcpinfo data short read (%d); want %d: %w", len(b), sizeofTCPInfo, err)
	}
```

khasanovbi: Maybe we can simplify code?

tcp.go
```diff
@@ -0,0 +1,22 @@
+package netlink
+
+const (
+	sizeofTCPInfo = 0x68
```
khasanovbi, line 4:
```suggestion
	sizeofTCPInfo = 0xe8
```

khasanovbi:
```
	z := &bytes.Buffer{}
	err := binary.Write(z, native, TCPInfo{})
	if err != nil {
		panic(err)
	}
	fmt.Println(z.Len())
```

This code show that your struct size 232 = 0xe8

RealFatCat:
Yeah, found the problem that in different kernels we have different size of tcp_info struct.
Moreover, some fields can be added in the middle of structure.
For example, in 4.19 there is `tcpi_delivery_rate_app_limited` right after scales https://elixir.bootlin.com/linux/v4.19/source/include/uapi/linux/tcp.h#L184
But in v4.4 we don't have it: https://elixir.bootlin.com/linux/v4.4/source/include/uapi/linux/tcp.h#L159
I didn't count on such "inserts" in the kernel.

Any advises how to handle this?
Keep different structures for different kernels and determine current version on deserialization?

khasanovbi:
Yeap, understand problem now.
I'm think, we could determine version or use size of b.
Maybe we could write some private structs for different versions (looks like blame should help https://github.com/torvalds/linux/blame/29d9f30d4ce6c7a38745a54a8cddface10013490/include/uapi/linux/tcp.h) and converters to our unified struct.
We can use struct embedding to don't duplicate same fields between different linux versions.

RealFatCat:
Probably, there is no issue with `tcpi_delivery_rate_app_limited`.
I'll update this PR after some tests.

RealFatCat:
This is probably not the most beautiful solution, but should work on every kernel.

socket_linux.go
```diff
@@ -160,3 +171,68 @@ func SocketGet(local, remote net.Addr) (*Socket, error) {
...
+		return nil, err
+	}
+	defer s.Close()
+
+	req := nl.NewNetlinkRequest(nl.SOCK_DIAG_BY_FAMILY, unix.NLM_F_DUMP)
+	req.AddData(&socketRequest{
+		Family:   family,
+		Protocol: unix.IPPROTO_TCP,
+		Ext:      INET_DIAG_INFO,
+		States:   uint32(0xfff), // All TCP states
```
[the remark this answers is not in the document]
RealFatCat: No, it's not enough. `0xf` is only 4 bits, but we have 11 states, and, as far I can understand, the first bit is ignored. Actually, there are 12 states, if `TCP_NEW_SYN_RECV` counts. So, this mask probably should be `0x1fff`, but `ss` do not requests this state, so I used `0xfff`.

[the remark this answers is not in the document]
aboch: :facepalm: That's right.

aboch:
Is it feasible to add a UT for this ?

tcp_linux.go
```diff
@@ -0,0 +1,393 @@
...
+	}
+	t.Snd_wscale = scales >> 4  // first 4 bits
+	t.Rcv_wscale = scales & 0xf // last 4 bits
+
+	rateLimAndFastOpen, err := rb.ReadByte()
+	if err != nil {
+		return checkDeserErr(err)
+	}
+	t.Delivery_rate_app_limited = rateLimAndFastOpen >> 7   // get first bit
+	t.Fastopen_client_fail = rateLimAndFastOpen >> 5 & 0b11 // get next two bits
```
aboch, line 119: image

RealFatCat:
> 
> 
> Is it feasible to add a UT for this ?

I really don't know how to do it properly. It seems like we need to mock kernel answers somehow, to compare them with expected stats data. But how.

vishvananda:
This looks good? How about a unit test to make sure that the request to the kernel for stats returns data without actually checking the values to see if they are valid? That would at least validate that the call works.

RealFatCat:
Will do.

PR #535, opened 2020-04-14
Allow placing veth peer into a namespace

const86:
Creating a veth pair across namespaces now is a multistep process.
Doing it in one shot with this change is clearer as current
namespace never sees peer IF. Also, moving peer into a namespace
may be rather slow, so better avoided.


const86:
@vishvananda, anything I could do to push this through?

aboch:
Thanks.
Please add a UT for this as well.

const86:
@aboch, added tests and rebased onto fresh master branch.

PR #540, opened 2020-05-04
Configure numrxqueues/numtxqueues on veth peer (fixes #536)

### LINKED ISSUE #536: NumRxQueues/NumTxQueues is not configured on veth peer
It does not appear possible to correctly configure the number of rx/tx queues on a veth pair.

When creating a veth pair like so:

```go
link := nl.Veth{
	LinkAttrs: nl.LinkAttrs{
		Name:  "outer0",
		NumRxQueues: 16,
		NumTxQueues: 16,
	},
	PeerName: "inner0",
}
```

I would expect the outcome to be equivalent to the following `ip` command:

```sh
ip link add outer0 numrxqueues 16 numtxqueues 16 type veth peer inner0 numrxqueues 16 numtxqueues 16
```

However, this library actually performs the equivalent of the following `ip` command:

```sh
ip link add outer0 numrxqueues 16 numtxqueues 16 type veth peer inner0
```

The problem seems to be here, where the NumRxQueues/NumTxQueues attributes are not applied to the veth peer interface:

https://github.com/vishvananda/netlink/blob/5f2fc868c2d0841565778edda0e0ea2843a49431/link_linux.go#L1277-L1290

alexforster:
See #536

### DIFF
--- a/link_linux.go
+++ b/link_linux.go
@@ -1282,6 +1282,12 @@ func (h *Handle) linkModify(link Link, flags int) error {
 		if base.TxQLen >= 0 {
 			peer.AddRtAttr(unix.IFLA_TXQLEN, nl.Uint32Attr(uint32(base.TxQLen)))
 		}
+		if base.NumTxQueues > 0 {
+			peer.AddRtAttr(unix.IFLA_NUM_TX_QUEUES, nl.Uint32Attr(uint32(base.NumTxQueues)))
+		}
+		if base.NumRxQueues > 0 {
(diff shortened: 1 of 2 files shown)


vishvananda:
this looks correct. Could you modify the test to check for this as well, around here? https://github.com/vishvananda/netlink/blob/c79a4b7b40668c3f7867bf256b80b6b2dc65e58e/link_test.go#L93

alexforster:
Added.

aboch:
Thanks for adding the UT.

LGTM

PR #541, opened 2020-05-15
Add LinkModify support

gollub:
Signed-off-by: Daniel Gollub <dgollub@att.com>


aboch:
Please also add a UT for this.

gollub:
Added a UT for LinkModify which creates a Dummy interface and updates/modifies the MTU.

aboch:
LGTM

PR #543, opened 2020-06-02
Add LinkIndex to Addr struct

brb:
Currently, it's cumbersome to get a link by an IP addr - one needs to list all links and then call `AddrList()` for each of them.

Considering that ifindex is already available to to the `parseAddr()` helper function, we can expose it to a user via the newly added `Addr.LinkIndex` field. This makes the retrieving link by IP addr much more simple.

### DIFF
--- a/addr.go
+++ b/addr.go
@@ -17,6 +17,7 @@ type Addr struct {
 	Broadcast   net.IP
 	PreferedLft int
 	ValidLft    int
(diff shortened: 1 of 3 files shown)


brb:
Ping @aboch @jellonek

brb:
@aboch Thanks!

PR #559, opened 2020-08-18
rule_linux: convert IIFNAME and OIFNAME to null terminated string

Cordius:
Strings in GO is not null-terminated while linux is written by
C and strings in C is null-terminated. Request will fail if we
perform rule request with not null-terminated iifname or ofiname,
with error message "no such file or directory".

It has been tested on CentOS 7.x while in newer kernel version such
as 4.19 this issue doesn't exist

Signed-off-by: Wu Zongyong <wuzongyong@linux.alibaba.com>

### DIFF
--- a/rule_linux.go
+++ b/rule_linux.go
@@ -142,10 +142,10 @@ func ruleHandle(rule *Rule, req *nl.NetlinkRequest) error {
 		}
 	}
 	if rule.IifName != "" {
-		req.AddData(nl.NewRtAttr(nl.FRA_IIFNAME, []byte(rule.IifName)))
+		req.AddData(nl.NewRtAttr(nl.FRA_IIFNAME, []byte(rule.IifName+"\x00")))
 	}
 	if rule.OifName != "" {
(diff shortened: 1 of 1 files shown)


Cordius:
Test Code
```go
package main

import (
        "fmt"
        "github.com/pkg/errors"
)

func main() {
        rule := netlink.NewRule()
        rule.Priority = 1000
        rule.IifName = "test"
        rule.Family = netlink.FAMILY_V4
        rule.Table = 100

        if err := netlink.RuleAdd(rule); err != nil {
                fmt.Println(errors.Wrapf(err, "add rule error"))
                return
        }

        rules, err := netlink.RuleList(netlink.FAMILY_V4)
        if err != nil {
                fmt.Println(errors.Wrapf(err, "list rules error"))
                return
        }

        for _, rule := range rules {
                if rule.Table == 100 {
                        if err := netlink.RuleDel(&rule); err != nil {
                                fmt.Println(errors.Wrapf(err, "delete rules error"))
                                return
                        }
                        fmt.Println("delete ok")
                }
        }

}
```

Issue #560, opened 2020-08-24
NewHtbClass calculates incorrect Buffer/Cbuffer


chenchun:
Golang netlink lib calculates buffer=15824 and cbuffer=15824 for 1000Mbit rate/ceil.
![image](https://user-images.githubusercontent.com/985861/90998911-c0a5de00-e5f7-11ea-9fa6-007a8c98b614.png)

C tc calculates buffer=187 and cbuffer=187 for 1000Mbit rate/ceil.
![90998702-52f9b200-e5f7-11ea-9bfc-0de9ebbf6ef8](https://user-images.githubusercontent.com/985861/91147345-b57faa80-e6ea-11ea-8224-fe0ac03cf4ef.png)

It seems this is due to the hz is incorrect in golang.

chenchun:
Fixed by https://github.com/vishvananda/netlink/pull/561

PR #561, opened 2020-08-24
Fix NewHtbClass

chenchun:
- Prio and Quantum should be set equal to the input params
- Fix buffer and cbuffer due to incorrect hz


chenchun:
Fixes https://github.com/vishvananda/netlink/issues/560

chenchun:
Add a second commit to fix Xmittime which is also not equal the behavior of c library. C library converts time to uint32 to drop precision before multiply tick_in_usec. ref https://git.kernel.org/pub/scm/network/iproute2/iproute2.git/tree/tc/tc_core.c#n62

vishvananda:
this lgtm, but can you squash it into commit please?

chenchun:
done

PR #567, opened 2020-09-24
Add Geneve link support

shassard:
Heavily based on the existing Gretap support


shassard:
See comments in PR related to testing of previous PR: https://github.com/vishvananda/netlink/pull/565

link.go
```diff
@@ -834,6 +834,30 @@ func (b *BondSlave) SlaveType() string {
...
+// https://github.com/torvalds/linux/blob/47ec5303d73ea344e84f46660fff693c57641386/drivers/net/geneve.c#L1209-L1223
+type Geneve struct {
+	LinkAttrs
+	ID              uint32 // vni
+	Remote          net.IP
+	Ttl             uint8
+	Tos             uint8
+	Dport           uint16
+	UdpCsum         uint8
+	UdpZeroCsum6_Tx uint8
```
vishvananda, line 847: It is generally against go style to use underscores in Attributes. Could you remove them from this and the following attr?

vishvananda (changes requested):
Very minor change here. Otherwise the patch looks good.

link.go
shassard: I've adjusted the attribute names as requested.

ddyzhang:
Sorry for the ping @vishvananda, but does this PR look good to you?

Looking forward to begin consuming these changes.

ddyzhang:
Also, @shassard I've been testing this locally and I think there might be a small bug in the marshaling and unmarshling of the VNI. I can't get Geneve interface creation to work properly when the VNI is sufficiently large.

For example, replacing the `0x1000` test case with something larger like `0x8eecf6` causes the test to fail. Digging into this, I managed to fix it by copying the VNI marshaling/unmarshaling logic from the VXLAN implementation, which also happens to be a 24-bit identifier.

With my changes, [L2444 of link_linux.go](https://github.com/shassard/netlink/commit/30a7e7e40bb8d4a97dee7f08b2a54a88915bca64#diff-350920f2216bdd28aa0ddd41f71bcab0R2444) becomes:
```
data.AddRtAttr(nl.IFLA_GENEVE_ID, nl.Uint32Attr(geneve.ID))
```

Similarly, [L2465 of link_linux.go](https://github.com/shassard/netlink/commit/30a7e7e40bb8d4a97dee7f08b2a54a88915bca64#diff-350920f2216bdd28aa0ddd41f71bcab0R2465) becomes:
```
geneve.ID = native.Uint32(datum.Value[0:4])
```

Another thing I've noticed is that after fixing these issues, the Geneve tunnel that's generated doesn't seem to have the remote target IP address set properly. After writing a small program to call these new Geneve operations, I found that the resulting interface has no remote configured (verified via `ip -d link show <geneve-interface-name>`). There is currently no logic to unmarshal the remote IP address from the netlink response and no tests to assert against the result, so that might be handy here.

ddyzhang:
Update: found the reason why the remote addr isn't being set correctly. On [L2437 of link_linux.go](https://github.com/shassard/netlink/commit/30a7e7e40bb8d4a97dee7f08b2a54a88915bca64#diff-350920f2216bdd28aa0ddd41f71bcab0R2437), we actually need to feed `ip4` into the data object, not `ip`. Similarly, we need to feed the result of `ip.To16()` into the data object for `IFLA_GENEVE_REMOTE6`.

shassard:
The Linux kernel geneve VNI is limited to 23bits from what I’ve observed. Does that work better than maxing out at 24bits?

ddyzhang:
I don't believe we're doing anything right now to restrict the input to 24 bits or less, but the parsing of the netlink response is only reading 24 bits. The issue I was encountering was when creating Geneve tunnels with VNIs that are quite a bit smaller than the max 24-bit value.

aboch:
LGTM
@vishvananda 's request was addressed
merging

PR #569, opened 2020-09-30
Add whole route metrics for netlink

bianze:
There are lots of route metrics to be added, such as rto_min,
quick_ack and so on, and these metrics are useful for users to
tune network performance, so we add all these metrics.

Signed-off-by: bianze <bianze1993@gmail.com>

### DIFF
--- a/route.go
+++ b/route.go
@@ -29,25 +29,38 @@ type Encap interface {
 
 // Route represents a netlink route.
 type Route struct {
-	LinkIndex  int
-	ILinkIndex int
-	Scope      Scope
-	Dst        *net.IPNet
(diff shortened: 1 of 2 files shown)


bianze:
> There are lots of route metrics to be added, such as rto_min,
> quick_ack and so on, and these metrics are useful for users to
> tune network performance, so we add all these metrics.
> 
> Signed-off-by: bianze [bianze1993@gmail.com](mailto:bianze1993@gmail.com)

#575

bianze:
Please take a look @vishvananda / @aboch . Thanks in advance !

PR #572, opened 2020-10-01
SocketDiagTCPInfo: Add support for BBR information

taktv6:


taktv6:
Any chance I could get some feedback?

vishvananda:
This is looking good to me. Any way to add a test for the functionality? Also, when you are ready to go, please squash your changes down into a single commit.

taktv6:
Thank you for your feedback. I've refactored the SocketDiagTCPInfo function and added some tests.
Unfortunately testify doesn't support Go 1.12 anymore. What would you like me to use as an alternative?

leoluk:
This would be super useful to have - any news on this?

taktv6:
@vishvananda feedback pls?

socket_linux.go
```diff
@@ -184,7 +184,7 @@ func SocketDiagTCPInfo(family uint8) ([]*InetDiagTCPInfoResp, error) {
 	req.AddData(&socketRequest{
 		Family:   family,
 		Protocol: unix.IPPROTO_TCP,
-		Ext:      INET_DIAG_INFO,
+		Ext:      (1 << (INET_DIAG_VEGASINFO - 1)) | (1 << (INET_DIAG_INFO - 1)),
```
aboch, line 187: Is it ok for `SocketDiagTCPInfo()` to now return both _INET_DIAG_INFO_ and _INET_VEGAS_INFO_?
I see there is a bunch of DIAG_INFO options. Would the two above go hand in hand for TCP or could other ones also apply? Otherwise should we export a function that let the caller decide which INFO to pull, the Ext bitbask?

taktv6: It would be more flexible if the user could specify the options/bitmask.
However, specifying flags that are not parsed (yet) could then be interpreted as a bug.
Also dealing with bitmasks is not very straight forward for many. So I think this lib should definitely hide it somehow.

I think primary question is: Do we care if the routine responds with additional data that a user didn't ask for explicitly?
If yes I'd make that configurable via an Options struct/parameter.

aboch: >Also dealing with bitmasks is not very straight forward for many. So I think this lib should definitely hide it somehow.

Yes, I see in the respective kernel code the bitmask creation is a bit quirky (`1<< (enum -1)`).
I was hoping in this library we could define those const values directly as bitmask, so then user can call our method with `const1 | const2 | ...` as they are already exported
```
const (
	INET_DIAG_NONE = 1
	INET_DIAG_MEMINFO = 2
        INET_DIAG_INFO = 4
...
```
This way we do not need to define any option setter function or option parsing logic.
WDYT?

taktv6: Honestly, I think when someone takes a look into TCP connection statistics they are usually happy to also get congestion avoidance aglorithm data like those for BBR. On the other hand I can't really think of anyone being unhappy about that.

So my conclusion is to just leave it as it is now as it is probably not getting any simpler.

aboch: True, actually the function is for `TCPInfo`. It makes sense to return any diag info related to TCP.

aboch:
LGTM

Please squash these changes in one commit ("Squash and merge" is not enabled here).

taktv6:
Okay. But any idea about the testify issue?
See https://github.com/vishvananda/netlink/issues/596

aboch:
>But any idea about the testify issue?

Yes, I'd suggest to remove the testify import from these changes. Check the struct equality in the test code.

I would stick to the builtin go testing package, because of the simplicity of the functions that needs to be tested here and for keeping this small library code base  lean.

taktv6:
Okay. I'll sort that out tomorrow. Thanks!

aboch:
LGTM

PR #578, opened 2020-10-13
VF stats Deserialization method and struct added. 

killianmuldoon:
VfStats now included in the VfInfo struct and a DeserializeVfStats method is added in order to add the stats to the VFInfo struct.


killianmuldoon:
I'm not sure why the build is failing for go 1.14 - the test seems to be unrelated to the changes made in this PR

PR #582, opened 2020-10-19
Fixed endianity problem with Geneve tunnel VNI

mlitvin:
It seems that VNI (ID in the code) is passed in host order.

The change include tests to verify that we use the correct endianity: It create a tunnel using `ip link` and read the ID of that tunnel proving that reading has the correct endianity. The test that set and read proves that the write endianity is the same as the read.

### DIFF
--- a/link_linux.go
+++ b/link_linux.go
@@ -2476,7 +2476,7 @@ func addGeneveAttrs(geneve *Geneve, linkInfo *nl.RtAttr) {
 	}
 
 	if geneve.ID != 0 {
-		data.AddRtAttr(nl.IFLA_GENEVE_ID, htonl(geneve.ID))
+		data.AddRtAttr(nl.IFLA_GENEVE_ID, nl.Uint32Attr(geneve.ID))
 	}
 
 	if geneve.Dport != 0 {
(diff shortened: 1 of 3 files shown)


mlitvin:
Test failure is because the tests are running on ubuntu 16.04 where iproute2 doesn't support dstport for Geneve.

So the options are:
1. Move to more modern ubuntu version in the travis configuration.
2. Remove the part about dstport for that test
3. attempt to identify the iproute version in the unit test code.
4. give up on comparing the code to iproute2 (it is not "pure" unit testing)

vishvananda:
can you try modifying the travis config in your branch to use a newer version of ubuntu? That way we can see if any other tests start failing.

mlitvin:
Doesn't look good, we have problems with modprobe nf_conntrack_ipv4 and qdisc

mlitvin:
Pull request was changes so we don't attempt to explicitly set the distort. It rely (and verify) the implicit Geneve port so we don't actually loose a lot in testing.

link_linux.go
```diff
@@ -2448,6 +2454,58 @@ func linkFlags(rawFlags uint32) net.Flags {
...
+
+	if geneve.FlowBased {
+		// In flow based mode, no other attributes need to be configured
+		linkInfo.AddRtAttr(nl.IFLA_GENEVE_COLLECT_METADATA, boolAttr(geneve.FlowBased))
+		return
+	}
+
+	if ip := geneve.Remote; ip != nil {
+		if ip4 := ip.To4(); ip4 != nil {
+			data.AddRtAttr(nl.IFLA_GENEVE_REMOTE, []byte(ip4))
```
aboch, line 2468: why converting a `net.IP` to a byte slice? `net.IP` is  a `[]byte`

aboch:
please rebase

mlitvin:
Most of the support was done in a separate commit, but the endianity problem need to be fixed

aboch:
Thanks.
Please squash your commits into one

aboch:
LGTM

PR #583, opened 2020-10-20
tap device: set MTU on tap devices

maiqueb:
Sets the MTU on tap devices.

Adds a tests that sets/asserts the tap device MTU.

Signed-off-by: Miguel Duarte Barroso <mdbarroso@redhat.com>


maiqueb:
Abandoning since I can use the `LinkSetMTU` function defined @ https://github.com/vishvananda/netlink/blob/d40f9887b852ec4ce633f2a46734422c87fc8336/link_linux.go#L366 to achieve the same result.

PR #585, opened 2020-10-29
Adding support for RTA_VIA

fach:
RTA_VIA is the netlink attribute type used to specify a gateway for a route using a different address family than said route. An example of this is using a IPv4 next hop for an IPv6 route. This PR adds support to set/get routes containing the RTA netlink attribute type.

Signed-off-by: Steve Shaw <shaw38@gmail.com>

### DIFF
--- a/route_test.go
+++ b/route_test.go
@@ -1339,3 +1339,69 @@ func TestMTURouteAddDel(t *testing.T) {
 		t.Fatal("Route not removed properly")
 	}
 }
+
+func TestRouteViaAddDel(t *testing.T) {
+	minKernelRequired(t, 5, 4)
+	tearDown := setUpNetlinkTest(t)
+	defer tearDown()
+
+	_, err := RouteList(nil, FAMILY_V4)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	link, err := LinkByName("lo")
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	if err := LinkSetUp(link); err != nil {
+		t.Fatal(err)
+	}
+
+	route := &Route{
+		LinkIndex: link.Attrs().Index,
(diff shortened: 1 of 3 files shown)


aboch:
Can you please add or extend existing UT functions to test the programming and reading of the `via` setting

route_test.go
```diff
@@ -1339,3 +1340,68 @@ func TestMTURouteAddDel(t *testing.T) {
...
+	}
+	routes, err = RouteList(link, FAMILY_V4)
+	if err != nil {
+		t.Fatal(err)
+	}
+	if len(routes) != 0 {
+		t.Fatal("Route not removed properly")
+	}
+
+	tearDown()
```
aboch, line 1406: any reason for not to using `defer` ?

fach: Nope, good catch.

route_test.go
```diff
@@ -1339,3 +1340,67 @@ func TestMTURouteAddDel(t *testing.T) {
...
+
+	routes, err := RouteList(link, FAMILY_V4)
+	if err != nil {
+		t.Fatal(err)
+	}
+	if len(routes) != 1 {
+		t.Fatal("Route not added properly")
+	}
+
+	if !cmp.Equal(route.MultiPath[0].Via, routes[0].Via) {
```
aboch, line 1391: do we really need to import this go-cmp/cmp to compare a two fields structure also considering the structure already has a `Equal()` method? 

Unless widely used and needed, I'd prefer to keep this project lean from outside imports

fach: Nope, it can be removed.

aboch:
Thanks @fach please squash your commits into one and force push.

fach:
Tests were failing due to the kernel version on the test runner. Added minimum kernel version to the unit tests for this.

PR #581, opened 2020-10-19
Add additional attributes for fou

corny:

PR #597, opened 2020-11-23
Optimizations

corny:


nl/nl_linux.go
```diff
@@ -362,46 +362,44 @@ func (a *RtAttr) Serialize() []byte {
 
 type NetlinkRequest struct {
```
aboch, line 363: This is an exported type. Other project may already be using it and the members you are removing are unfortunately exported as well.

nl/nl_linux.go
```diff
@@ -362,46 +362,44 @@ func (a *RtAttr) Serialize() []byte {
...
-	return b
+	copy(data[:unix.SizeofNlMsghdr], hdr)
+
+	return data
 }
 
-func (req *NetlinkRequest) AddData(data NetlinkRequestData) {
-	req.Data = append(req.Data, data)
+// AddRtAttr adds an RtAttr to the request
+func (req *NetlinkRequest) AddRtAttr(attrType int, data []byte) *NetlinkRequest {
```
aboch, line 385: This one also changes existing exported methods.

aboch (commented):
Thank you for the change.
We need to see how to deal with the API changes

aboch:
Looks like the Travis failed at the same test as in #581 
Please double check, it may be indeed caused by your changes, or the test needs to be updated

aboch:
pls fix conflicts

aboch:
ping...

PR #598, opened 2020-11-26
Adds String methods to several bond params and fixes Packers -> Packets

RiccardoManfrin:


aboch:
@RiccardoManfrin  go1.13 and go1.14 CIs failed at

image

Please also squash you commits into one

RiccardoManfrin:
I've squashed the commits and removed
```
case unix.RTPROT_KEEPALIVED:
    return "keepalived"
```

aboch:
You are proposing 2 commits. One that adds x number of things and one that remove something from x.

Please have only one commit.

If you are concerned about leaving a trace about why RT_PROT is not stringified, leave it in the code commented out, with explanation. So somebody can uncomment it in future when the current three go release used support that unix pkg field.

aboch:
Grazie @RiccardoManfrin 

LGTM

PR #603, opened 2020-12-15
Add support for addition FQ Codel knobs

sargun:
These knobs are supported in recent kernels, and are useful for tuning
in the datacenter.


aboch:
LGTM

PR #604, opened 2020-12-16
Allow building on non-linux machine

mlitvin:
Currently build is broken on non-linux machine, which is annoying when using netlink in projects that can be mostly be built and tested (or even just edited) on other platforms.

Most of the netlink code already work this way, there are few places where code needed to be moved for *_linux.go to regular files and vice versa.


aboch:
LGTM

PR #607, opened 2020-12-31
Add ipvtap link support

RichardWeiYang:
ipvtap is a similar link type as ipvlan with tap interface.

This patch enables it just like macvtap.

Signed-off-by: Wei Yang <richard.weiyang@gmail.com>


aboch:
Can you add or extend the existing UT to do some basic testing of this?
Program, read, make sure params matches.

Issue #613, opened 2021-01-25
Allowing to set socket timeout for default pkgHandle


ArthurChiao:
Hi,

Noticed that lots of downstream projects perform netlink operations with `netlink.XXX` API directly, which in the background uses the **default pkg handle** [`pkgHandle`](https://github.com/vishvananda/netlink/blob/master/handle_linux.go#L13). 

While it's the most intuitive and straight-forward way to use this package (for most scenarios), and works wonderfully most of the time, there are still probabilities of failures (**hanging**, specifically) due to **socket timeout option not set** for `pkgHandle`, such as cases in popular projects [Calico (issue)](#254) and [Cilium (issue)](https://github.com/cilium/cilium/pull/14502#issuecomment-756548459).

So I'm wondering if we could improve it by allowing users to set socket timeout for the default handle. Such as the following draft:

```
diff --git a/handle_linux.go b/handle_linux.go
index 26887b7..e536310 100644
--- a/handle_linux.go
+++ b/handle_linux.go
@@ -10,7 +10,8 @@ import (
 )

 // Empty handle used by the netlink package methods
-var pkgHandle = &Handle{}
+var pkgHandle *Handle = nil
+var socketTimeout = 5 * time.Minute

+func init() {
+       h, err := NewHandle()
+       if err != nil {
+               panic("Create default pkg handle failed")
+       }
+
+       pkgHandle = h
+}
+
+func SetSocketTimeout(t time.Duration) error {
+       if pkgHandle == nil {
+               return fmt.Errorf("Default pkg handle not initialized")
+       }
+
+       socketTimeout = t
+       return pkgHandle.SetSocketTimeout(socketTimeout)
+}
+
+func GetSocketTimeout() (time.Duration, error) {
+       if pkgHandle == nil {
+               return 0, fmt.Errorf("Default pkg handle not initialized")
+       }
+
+       return socketTimeout, nil
+}
```

With this option, users no longer need to refactor their existing code by initializing their own pkg handles, which will save much effort for them.

Thanks!

PR #614, opened 2021-01-27
Allow to set/get socket timeout option for default handle

ArthurChiao:
Ref: https://github.com/vishvananda/netlink/issues/613

Signed-off-by: arthurchiao <arthurchiao@hotmail.com>

### DIFF
--- a/handle_linux.go
+++ b/handle_linux.go
@@ -21,6 +21,22 @@ type Handle struct {
 	lookupByDump bool
 }
 
+// SetSocketTimeout configures timeout for default netlink sockets
+func SetSocketTimeout(to time.Duration) error {
+	if to < time.Microsecond {
+		return fmt.Errorf("invalid timeout, minimul value is %s", time.Microsecond)
+	}
+
+	nl.SocketTimeoutTv = unix.NsecToTimeval(to.Nanoseconds())
+	return nil
+}
+
+// GetSocketTimeout returns the timeout value used by default netlink sockets
+func GetSocketTimeout() time.Duration {
+	nsec := unix.TimevalToNsec(nl.SocketTimeoutTv)
+	return time.Duration(nsec) * time.Nanosecond
+}
+
 // SupportsNetlinkFamily reports whether the passed netlink family is supported by this Handle
 func (h *Handle) SupportsNetlinkFamily(nlFamily int) bool {
 	_, ok := h.sockets[nlFamily]
--- a/handle_linux_test.go
+++ b/handle_linux_test.go
@@ -1,0 +1,17 @@
+package netlink
+
+import (
+	"testing"
+	"time"
+)
+
+func TestSetGetSocketTimeout(t *testing.T) {
+	timeout := 10 * time.Second
+	if err := SetSocketTimeout(10 * time.Second); err != nil {
+		t.Fatalf("Set socket timeout for default handle failed: %v", err)
+	}
+
+	if val := GetSocketTimeout(); val != timeout {
+		t.Fatalf("Unexpcted socket timeout value: got=%v, expected=%v", val, timeout)
+	}
+}
--- a/nl/nl_linux.go
+++ b/nl/nl_linux.go
@@ -35,6 +35,9 @@ var SupportedNlFamilies = []int{unix.NETLINK_ROUTE, unix.NETLINK_XFRM, unix.NETL
 
 var nextSeqNr uint32
 
+// Default netlink socket timeout, 60s
+var SocketTimeoutTv = unix.Timeval{Sec: 60, Usec: 0}
+
 // GetIPFamily returns the family type of a net.IP.
 func GetIPFamily(ip net.IP) int {
 	if len(ip) <= net.IPv4len {
@@ -426,6 +429,14 @@ func (req *NetlinkRequest) Execute(sockType int, resType uint16) ([][]byte, erro
 		if err != nil {
 			return nil, err
 		}
+
+		if err := s.SetSendTimeout(&SocketTimeoutTv); err != nil {
+			return nil, err
+		}
+		if err := s.SetReceiveTimeout(&SocketTimeoutTv); err != nil {
+			return nil, err
+		}
+
 		defer s.Close()
 	} else {
 		s.Lock()


brb:
Hey @aboch , could you take a look?

handle_linux.go
```diff
@@ -9,8 +9,11 @@ import (
 	"golang.org/x/sys/unix"
 )
 
-// Empty handle used by the netlink package methods
-var pkgHandle = &Handle{}
+// Default handle used by the netlink package methods
+var pkgHandle *Handle = nil
```
aboch, line 13: Why do we need to make this a pointer?
Can't we still set the socket TO in _init()_?

ArthurChiao: Hi @aboch thanks for reviewing!

Not sure if I understand your question corrctly: if `pkgHandle` is initialized as `&Handle{}` as previously, the `pkgHandle.sockets` will be an empty map, and we can not set TO values for these sockets.

aboch: Thanks @ArturChiao now I remember.

With your proposed change, any binary depending on _vishvananda/netlink_ will open 3 nl sockets, even when no netlink request is actually being executed (https://github.com/vishvananda/netlink/blob/v1.1.0/nl/nl_linux.go#L34).

Current behavior of this library has been: Open the nl socket only when you execute a pkg method (AddAddr(), AddLink(),...).

The addition of the _Handle_ thing was done in a way so that above behavior was not modified.

I know it is limiting, but have you considered setting a default TO in `func (req *NetlinkRequest) Execute()` at https://github.com/vishvananda/netlink/blob/v1.1.0/nl/nl_linux.go#L406 instead ?

At the end of the day, a minute timeout (the default you are proposing) should satisfy most if not all use-cases.

ArthurChiao: Thanks for the explanations!  @aboch 

Set TO for [each socket](https://github.com/vishvananda/netlink/blob/v1.1.0/nl/nl_linux.go#L402) we're going to use makes sense to me.
But I'm not sure if you would like to **make the TO value configurable or not** (e.g. hardcode 1 minute).

Ewocker:
I was trying to see what was the problem for the patch.
For test case **TestAddrAddReplace** the output of AddList originally is `[127.0.0.2/24 lo]` whereas with the patch there is an extra entry `[127.0.0.1/8 lo 127.0.0.2/24 lo]`.

ArthurChiao:
Hi @Ewocker, thanks for your reminder, I'll take a look at it!

Ewocker:
> Hi @Ewocker, thanks for your reminder, I'll take a look at it!

@ArthurChiao, asking out of curiosity, why was test TestAddrSubscribeWithOptions in original PR hitting the 1m TO? Also, why would the AddrAdd call create `127.0.0.1/8` in test TestAddrAddReplace ?

don't know too much about netlink, would really appreciate if you can answer my questions 😁

ArthurChiao:
@Ewocker actually I can't reproduce it in my develop env :(

And I found that even **without my changes**, many tests would fail in my dev node (already with root privilege), but with different reasons, such as the following one:

```shell
$ go test -v -run TestAddrAdd
=== RUN   TestAddrAdd
    TestAddrAdd: addr_test.go:90: Address flags not set properly, got=128, expected=140
--- FAIL: TestAddrAdd (0.00s)
=== RUN   TestAddrAddReplace
--- PASS: TestAddrAddReplace (0.00s)
FAIL
exit status 1
FAIL    github.com/vishvananda/netlink  0.004s
```

and this one:

```shell
--- FAIL: TestSEG6RouteAddDel (0.00s)
panic: runtime error: invalid memory address or nil pointer dereference [recovered]
        panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x38 pc=0x64863e]

goroutine 230 [running]:
testing.tRunner.func1.1(0x693980, 0x8ece40)
        /usr/local/go/src/testing/testing.go:940 +0x2f5
testing.tRunner.func1(0xc000305440)
        /usr/local/go/src/testing/testing.go:943 +0x3f9
panic(0x693980, 0x8ece40)
        /usr/local/go/src/runtime/panic.go:969 +0x166
github.com/vishvananda/netlink.TestSEG6RouteAddDel(0xc000305440)
        /home/arthurchiao/go/src/github.com/vishvananda/netlink/route_test.go:1176 +0xa9e
```

Haven't dug into these problems.

handle_linux.go
ArthurChiao: Refactored to set TO after we got a socket, please have a review, thanks! @aboch

brb (approved):
Thanks, LGTM!

nr17:
> @Ewocker actually I can't reproduce it in my develop env :(
> 
> And I found that even **without my changes**, many tests would fail in my dev node (already with root privilege), but with different reasons, such as the following one:
> 
> ```shell
> $ go test -v -run TestAddrAdd
> === RUN   TestAddrAdd
>     TestAddrAdd: addr_test.go:90: Address flags not set properly, got=128, expected=140
> --- FAIL: TestAddrAdd (0.00s)
> === RUN   TestAddrAddReplace
> --- PASS: TestAddrAddReplace (0.00s)
> FAIL
> exit status 1
> FAIL    github.com/vishvananda/netlink  0.004s
> ```
> 
> and this one:
> 
> ```shell
> --- FAIL: TestSEG6RouteAddDel (0.00s)
> panic: runtime error: invalid memory address or nil pointer dereference [recovered]
>         panic: runtime error: invalid memory address or nil pointer dereference
> [signal SIGSEGV: segmentation violation code=0x1 addr=0x38 pc=0x64863e]
> 
> goroutine 230 [running]:
> testing.tRunner.func1.1(0x693980, 0x8ece40)
>         /usr/local/go/src/testing/testing.go:940 +0x2f5
> testing.tRunner.func1(0xc000305440)
>         /usr/local/go/src/testing/testing.go:943 +0x3f9
> panic(0x693980, 0x8ece40)
>         /usr/local/go/src/runtime/panic.go:969 +0x166
> github.com/vishvananda/netlink.TestSEG6RouteAddDel(0xc000305440)
>         /home/arthurchiao/go/src/github.com/vishvananda/netlink/route_test.go:1176 +0xa9e
> ```
> 
> Haven't dug into these problems.

The Makefile (line#22) runs 4 tests in parallel. Tests setup and cleanup the new ns using setUpNetlinkTest.(netlink/netlink_test.go). Although the ns is supposed to be thread-local by making use of runtime.LockOSThread(), if the test goroutines running in parallel share the same underlying OS thread, they will interfere with each other.

handle_linux.go
brb: PTAL @aboch

PR #616, opened 2021-02-16
Properly handle uint64 for HTB class rate/ceil

imilchev:
Currently, providing a rate or a ceil with a value higher than the `uint32` max results in the value overflowing and a HTB class with wrong ceil/rate is created without any errors. The kernel already supports `uint64` values so adding support for it to the library was relatively easy.

I also adjusted the test to use large values to verify that it works properly. With the old implementation, the same test would fail

### DIFF
--- a/class_test.go
+++ b/class_test.go
@@ -268,7 +268,8 @@ func TestHtbClassAddHtbClassChangeDel(t *testing.T) {
 	}
 
 	htbclassattrs := HtbClassAttrs{
-		Rate:    1234000,
+		Rate:    uint64(1<<32) + 10,
+		Ceil:    uint64(1<<32) + 20,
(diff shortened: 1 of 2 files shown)


imilchev:
@aboch @vishvananda could one of you have a look at this PR?

class_test.go
```diff
@@ -268,7 +268,8 @@ func TestHtbClassAddHtbClassChangeDel(t *testing.T) {
 	}
 
 	htbclassattrs := HtbClassAttrs{
-		Rate:    1234000,
+		Rate:    uint64(1<<32) + 10,
+		Ceil:    uint64(1<<32) + 20,
```
aboch, line 272: :-)

aboch:
LGTM

PR #627, opened 2021-03-26
conntrack: allow to filter by subnet

aojea:
Add a new conntrack filter to be able to filter by subnet, in
addition to current IP address filter.
This keeps compatibility, since the IP filter is just an IP subnet filter with full mask.

Signed-off-by: Antonio Ojea <aojea@redhat.com>

### DIFF
--- a/conntrack_linux.go
+++ b/conntrack_linux.go
@@ -346,21 +346,32 @@ type CustomConntrackFilter interface {
 }
 
 type ConntrackFilter struct {
-	ipFilter    map[ConntrackFilterType]net.IP
+	ipNetFilter map[ConntrackFilterType]*net.IPNet
 	portFilter  map[ConntrackFilterType]uint16
 	protoFilter uint8
 }
 
-// AddIP adds an IP to the conntrack filter
-func (f *ConntrackFilter) AddIP(tp ConntrackFilterType, ip net.IP) error {
-	if f.ipFilter == nil {
-		f.ipFilter = make(map[ConntrackFilterType]net.IP)
+// AddIPNet adds a IP subnet to the conntrack filter
+func (f *ConntrackFilter) AddIPNet(tp ConntrackFilterType, ipNet *net.IPNet) error {
+	if ipNet == nil {
+		return fmt.Errorf("Filter attribute empty")
 	}
-	if _, ok := f.ipFilter[tp]; ok {
+	if f.ipNetFilter == nil {
+		f.ipNetFilter = make(map[ConntrackFilterType]*net.IPNet)
+	}
+	if _, ok := f.ipNetFilter[tp]; ok {
 		return errors.New("Filter attribute already present")
 	}
-	f.ipFilter[tp] = ip
+	f.ipNetFilter[tp] = ipNet
 	return nil
+}
+
+// AddIP adds an IP to the conntrack filter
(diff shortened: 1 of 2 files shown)


conntrack_linux.go
```diff
@@ -346,23 +346,32 @@ type CustomConntrackFilter interface {
...
 	}
-	f.ipFilter[tp] = ip
+	f.ipNetFilter[tp] = ipNet
 	return nil
 }
 
+// AddIP adds an IP to the conntrack filter
+func (f *ConntrackFilter) AddIP(tp ConntrackFilterType, ip net.IP) error {
+	// Convert IP to a subnet with a full mask
+	iplen := 8 * len(ip)
```
dcbw, line 369: @aojea can you explain a bit more what you're doing here? THere's gotta be something we can do that I can understand in 2 minutes of looking at it. Also, we're not To4() or To16() canonicalizing anything here; does the Contains() down below handle that automatically for us?

aojea: This creates a full mask, I tested locally but indeed is a bit opaque because I think it depends on the golang net IP internal types. I will make it explicit so there is no doubt, if you have an ipv4 it add /32 if you have ipv6 it adds /128

aojea: The contains check if an IP belongs to an IP subnet , the IP 192.268.2.1 belongs to the subnet 192.168.2.1/32

conntrack_linux.go
```diff
@@ -346,23 +346,45 @@ type CustomConntrackFilter interface {
...
+	f.ipNetFilter[tp] = ipNet
 	return nil
 }
 
+// AddIP adds an IP to the conntrack filter
+func (f *ConntrackFilter) AddIP(tp ConntrackFilterType, ip net.IP) error {
+	if ip == nil {
+		return fmt.Errorf("Filter attribute empty")
+	}
+	// Convert IP to a subnet with a full mask
```
danwinship, line 374: There's already a function that does that.

    return f.AddIPNet(tp, NewIPNet(ip))

conntrack_test.go
```diff
@@ -479,6 +479,65 @@ func TestConntrackFilter(t *testing.T) {
 		t.Fatalf("Error, there should be an exact match, v4:%d, v6:%d", v4Match, v6Match)
 	}
 
+	// SrcIPNet filter
+	filterV4 = &ConntrackFilter{}
+	filterV4.AddIPNet(ConntrackOrigSrcIP, &net.IPNet{IP: net.ParseIP("10.0.0.1"), Mask: net.CIDRMask(12, 32)})
```
danwinship, line 484: and there's a function for that too

```suggestion
	filterV4.AddIPNet(ConntrackOrigSrcIP, ParseIPNet("10.0.0.1/12"))
```

though presumably you mean `"10.0.0.0/12"` anyway?

actually all of the ipnet filters here look wrong; they should have 0s in the low bits

conntrack_linux.go
aojea: great

danwinship:
lgtm

dcbw:
LGTM

conntrack_linux.go
dcbw: thanks for the explanation, new push looks good to me

aboch:
please push again to trigger CI

PR #628, opened 2021-03-28
Support STP parameters in bridge

daniel-noland:
Added basic support for some basic spanning tree protocol configuration
parameters on the bridge device.

### DIFF
--- a/link_linux.go
+++ b/link_linux.go
@@ -303,6 +303,46 @@ func (h *Handle) BridgeSetVlanFiltering(link Link, on bool) error {
 	return h.linkModify(bridge, unix.NLM_F_ACK)
 }
 
+func BridgeSetVlanProtocol(link Link, protocol string) error {
+	return pkgHandle.BridgeSetVlanProtocol(link, protocol)
+}
+
+func (h *Handle) BridgeSetVlanProtocol(link Link, protocol string) error {
+	bridge := link.(*Bridge)
+	bridge.VlanProtocol = &protocol
+	return h.linkModify(bridge, unix.NLM_F_ACK)
+}
+
+func (h *Handle) BridgeSetStpState(link Link, state uint32) error {
+	bridge := link.(*Bridge)
+	bridge.StpState = &state
+	return h.linkModify(bridge, unix.NLM_F_ACK)
+}
+
+func BridgeSetStpState(link Link, state uint32) error {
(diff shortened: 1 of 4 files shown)


daniel-noland:
Notes:

1. I am very new to Go, so if I am deviating from any idioms or best practices please let me know and I will do my best to correct.
2. When I run the full `go-netlink` test suite some of the tests fail due to parameter off-by-one issues.  I am running linux kernel 5.11.10 at the time of writing.  I don't know why this is happening for me but I will assume this is not happening elsewhere until I dig in a little more and find out what is going on.  I mention this here because I deliberately wrote a couple more tests which fail for similar off-by-one issues under the assumption that they won't fail on other machines.  I am leaving this as a draft PR till I find out what is going on.

daniel-noland:
*Warning:*

My colleague ran the tests I included in this patch on his machine (running linux kernel 5.4) and all tests pass.

I am upgrading this PR from draft in the hopes that others will run the tests included in this patch and confirm.

I suggest that we make a point of running this test suite on various linux kernels.

*The tests fail for me on kernel 5.11.10*

The failure symptom is that all timer values I specify are accepted by the kernel (assuming they were otherwise valid).  However, on query every timer is reported as the specified value less one.  E.g., if I specify the max age stp parameter as 200 then the operation is is accepted but on querying the current value it returns as 199.

link_linux.go
```diff
@@ -303,6 +303,46 @@ func (h *Handle) BridgeSetVlanFiltering(link Link, on bool) error {
...
+	return pkgHandle.BridgeSetVlanProtocol(link, protocol)
+}
+
+func (h *Handle) BridgeSetVlanProtocol(link Link, protocol string) error {
+	bridge := link.(*Bridge)
+	bridge.VlanProtocol = &protocol
+	return h.linkModify(bridge, unix.NLM_F_ACK)
+}
+
+func (h *Handle) BridgeSetStpState(link Link, state uint32) error {
```
aboch, line 316: What values can _state_ take?

aboch:
also pls push again so CI is triggered

aboch:
ping...

vishvananda:
Looks like the tests are failing in CI here

aboch:
@daniel-noland

aboch:
@daniel-noland

vishvananda:
Closing due to inactivity. Feel free to reopen.

PR #635, opened 2021-04-07
Provide the allmulticast option as own link attribute

bjosv:
The `ip link` option allmulticast enables or disables the reception of all hardware multicast packets.

This PR provide the status of the allmulticast option via a highlevel link attribute instead of requiring raw flag handling. This also hides the dependencies to platform specific files for the user of this library.

### DIFF
--- a/link.go
+++ b/link.go
@@ -35,6 +35,7 @@ type LinkAttrs struct {
 	Alias        string
 	Statistics   *LinkStatistics
 	Promisc      int
+	Allmulti     int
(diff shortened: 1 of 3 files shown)


bjosv:
Updated testcase `TestLinkSetAllmulticast` to avoid intermittent faults due to independent change of flag `IFF_RUNNING`
Can be seen at https://github.com/vishvananda/netlink/runs/2403394603 for PR #646

PR #636, opened 2021-04-07
Correction required to build via Makefile

bjosv:
A Go formatting problem in route_linux.go seems to break builds using the Makefile.

This PR contains the trivial formatting correction, which possibly will make Travis CI builds pass again.


bjosv:
Issue corrected now by #646
Closing this PR

PR #645, opened 2021-04-21
[lwtunnel bpf] add support for encap BPF

bersoare:
this patch adds a new type for encap BPF, allowing FD and NAME to be set for programs, as well as the XMIT headroom.
bpf loading not included in this patch, but the bpf program can be loaded using a 3rd party library

example usage:
```
	fmt.Printf("prog fd: %d\n", prog.GetFd()) // loaded program, exposing its FD (it is pinned in this case)
	encap := netlink.BPFEncap{}
	err = encap.SetProg(nl.LWT_BPF_XMIT, prog.GetFd(), "mytest123")
	if err != nil {
		panic(err)
	}
	err = encap.SetXmitHeadroom(128)
	if err != nil {
		panic(err)
	}
	_, dst, _ := net.ParseCIDR("192.168.123.12/32")
	newRoute := &netlink.Route{Dst: dst, LinkIndex: link.Attrs().Index, Encap: &encap}
	err = handle.RouteAdd(newRoute)
	if err != nil {
		panic(err)
	}
```
result of the above code:
```
# ip ro sh | grep mytest
192.168.123.12  encap bpf xmit mytest123[fd:3] dev bareudp0
```


bersoare:
Alt Text

aboch:
please squash your commits into one

Issue #509, opened 2020-01-15
parseIptunData caused `index out of range` panic


AirbornWdd:
The `ip link` command dump:
```
13: tunl0@NONE: <NOARP> mtu 1480 qdisc noop state DOWN mode DEFAULT group default qlen 1000
    link/ipip 0.0.0.0 brd 0.0.0.0
14: ipip0@NONE: <NOARP,UP,LOWER_UP> mtu 1480 qdisc noqueue state UNKNOWN mode DEFAULT group default qlen 1000
    link/ipip 0.0.0.0 brd 0.0.0.0
15: ip6tnl0@NONE: <NOARP> mtu 1452 qdisc noop state DOWN mode DEFAULT group default qlen 1000
    link/tunnel6 :: brd ::
16: ipip60@NONE: <NOARP,UP,LOWER_UP> mtu 1452 qdisc noqueue state UNKNOWN mode DEFAULT group default qlen 1000
    link/tunnel6 :: brd ::
```

The panic backtrace are:
```
panic: runtime error: index out of range

 github.com/vishvananda/netlink.parseIptunData(0xcc4b40, 0xc000214580, 0xc00028ee00, 0xd, 0x10)
         /go/src/github.com/vishvananda/netlink/link_linux.go:2580 +0x2c1
 github.com/vishvananda/netlink.LinkDeserialize(0x0, 0xc00028497c, 0x590, 0x113c, 0x9, 0xc00028c700, 0x8, 0x10)
         /go/src/github.com/vishvananda/netlink/link_linux.go:1608 +0x1482
 github.com/vishvananda/netlink.(*Handle).LinkList(0x1359fd0, 0xbff8b3, 0x9, 0x21a, 0xc000034f70, 0xc000034f00)
         /go/src/github.com/vishvananda/netlink/link_linux.go:1813 +0x2e6
 github.com/vishvananda/netlink.LinkList(...)
         /go/src/github.com/vishvananda/netlink/link_linux.go:1791

```

AirbornWdd:
`dlv` dumps:
```
(dlv) bt
 0  0x00000000009772c3 in github.com/vishvananda/netlink.parseIptunData
    at /go/src/github.com/vishvananda/netlink/link_linux.go:2557
 1  0x000000000096b7e2 in github.com/vishvananda/netlink.LinkDeserialize
    at /go/src/github.com/vishvananda/netlink/link_linux.go:1608
[...]

PR #647, opened 2021-05-06
Fix another panic case of #509.

Dongdong-Wang:
Signed-off-by: Dongdong Wang <wangdongdong.6@bytedance.com>


bjosv:
The flaky testcase `TestLinkSetAllmulticast` that fails in this CI run is handled by vishvananda/netlink#635

PR #650, opened 2021-05-14
Make IPSet actually support cidrs, port,mark

lrh3321:
Supported set types:

- hash:mac
- hash:net,iface
- hash:net,port
- hash:net,port,net
- hash:net,net
- hash:ip,mark
- ...


aboch:
Can you move "add IpsetRename, IpsetSwap" to a another PR?
One commit per PR rule here.

PR #651, opened 2021-05-18
Remove trailing null char from string in devlink

adrianchiris:
netlink(kernel) returns the string values in a c-style
manner terminating with null.

when converting to go string these need to be removed
as done in other places in the project.

keeping the null terminating char prevents comparing
devlink dev/port string attributes as the `==` string
operand will fail.

Signed-off-by: Adrian Chiris <adrianc@nvidia.com>


adrianchiris:
cc @moshe010 @paravmellanox

paravmellanox (approved):
Reviewed-by: Parav Pandit <parav@nvidia.com>

PR #653, opened 2021-05-20
Ability to set multicast attribute on a link

dennisafa:
We are able to set the ALLMULTI attribute on a link, via LinkSetAllmulticast(). This change allows you to set the regular multicast attribute with LinkSetMulticast()


dennisafa:
No pressure, but was just wondering if this can be looked at @aboch @vishvananda thanks!

PR #661, opened 2021-06-12
Add linux route get oif option.

ywc689:
This option is equivalent to "ip route get ... dev ...", which is useful when the route has multipaths.

```bash
# ip route show
default 
	nexthop via x.x.68.62 dev eth0 weight 1 
	nexthop via x.x.69.62 dev eth1 weight 1
# ip route get 8.8.8.8 dev eth0
8.8.8.8 via x.x.68.62 dev eth0 src x.x.68.1 uid 0 
# ip route get 8.8.8.8 dev eth1
8.8.8.8 via x.x.69.62 dev eth1 src x.x.69.1 uid 0 
```


aboch:
Please add or extend existing UT to cover this code

ywc689:
> Please add or extend existing UT to cover this code

done

aboch:

ywc689:

done

Issue #664, opened 2021-06-25
ppp0 modem interface causes parseAddr panic


jswantek:
The parseAddr function (https://github.com/vishvananda/netlink/blob/master/addr_linux.go#L216) assumes that the `unix.IFA_ADDRESS` case will be hit, setting a valid value to the `dst` variable.  However, if the `unix.IFA_ADDRESS` case is _not_ hit then `dst` will remain unset (null) and cause a panic (https://github.com/vishvananda/netlink/blob/master/addr_linux.go#L271).

```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x1a31f90]

goroutine 20887 [running]:
github.com/rancher/k3s/vendor/github.com/vishvananda/netlink.parseAddr(0x400648bdec, 0x40, 0x40, 0x0, 0x400867a0a8, 0x4, 0x280, 0x0, 0x0, 0x400648be00, ...)
        /go/src/github.com/rancher/k3s/vendor/github.com/vishvananda/netlink/addr_linux.go:274 +0x170
github.com/rancher/k3s/vendor/github.com/vishvananda/netlink.(*Handle).AddrList(0x6721590, 0x43eff40, 0x400f0859e0, 0x2, 0x0, 0x4, 0x2bc, 0xfe, 0x1)
        /go/src/github.com/rancher/k3s/vendor/github.com/vishvananda/netlink/addr_linux.go:199 +0x180
github.com/rancher/k3s/vendor/github.com/vishvananda/netlink.AddrList(...)
        /go/src/github.com/rancher/k3s/vendor/github.com/vishvananda/netlink/addr_linux.go:174
github.com/rancher/k3s/vendor/github.com/coreos/flannel/pkg/ip.getIfaceAddrs(0x400e1e22c0, 0x0, 0x0, 0x0, 0x0, 0x0)
        /go/src/github.com/rancher/k3s/vendor/github.com/coreos/flannel/pkg/ip/iface.go:36 +0x68
github.com/rancher/k3s/vendor/github.com/coreos/flannel/pkg/ip.GetIfaceIP4Addr(0x400e1e22c0, 0x0, 0x0, 0x400775ee08, 0x1, 0x1)
        /go/src/github.com/rancher/k3s/vendor/github.com/coreos/flannel/pkg/ip/iface.go:40 +0x30
[...]

PR #665, opened 2021-06-28
Correct panic when IPv4 lacks IFA_ADDRESS address

jswantek:
IFA_ADDRESS is to be used as the peer address if it differs from IFA_LOCAL.
Therefore, include the check for "no IFA_ADDRESS" in the difference check.

Example: ppp interfaces can contain IFA_LOCAL and no IFA_ADDRESS attribute

Signed-off-by: Joe Swantek <joseph.swantek@northwestern.edu>

### DIFF
--- a/addr_linux.go
+++ b/addr_linux.go
@@ -268,7 +268,7 @@ func parseAddr(m []byte) (addr Addr, family int, err error) {
 	// But obviously, as there are IPv6 PtP addresses, too,
 	// IFA_LOCAL should also be handled for IPv6.
 	if local != nil {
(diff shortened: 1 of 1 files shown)


jswantek:
Addresses issue: https://github.com/vishvananda/netlink/issues/664

jswantek:
It was discovered that this does resolve a potential panic but there is other elements in the code-base that assume `IFA_ADDRESS` will be present.  Maybe a larger fix to remove that assumption is necessary?

aboch:
LGTM

Issue #588, opened 2020-10-30
filter: Add Vlan action support


DmytroLinkin:
Vlan action is the one of the actions, alongside with tunnel encapsulation, which allow enhanced traffic manipulations using Traffic Control in Linux.

PR #666, opened 2021-07-09
[WIP]add vlan filter action support

BSWANG:
#588
- [x] support vlan action
- [x] support vlan pop action
- [ ] support all vlan actions

### DIFF
--- a/filter.go
+++ b/filter.go
@@ -309,6 +309,41 @@ func (filter *BpfFilter) Attrs() *FilterAttrs {
 	return &filter.FilterAttrs
 }
 
+type VlanKeyAct uint8
+
+const (
+	TCA_VLAN_KEY_POP VlanKeyAct = 1 // pop vlan
+)
+
+func (k VlanKeyAct) String() string {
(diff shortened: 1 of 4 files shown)


vishvananda:
Can you fix conflicts and reopen? Note that the default branch is `main` now.

rschmied:
Did this go anywhere?  The reason I'm asking is that I am seeing some weird behavior using netlink:

- I create a veth pair
- both listen for raw packets, `conn, err = packet.Listen(iface, packet.Raw, 0x0003, nil)`  (with `ETH_P_ALL   0x0003`)
- both forward those packets to the other veth member, respectively

What I observe is that when I do have 802.1Q tagged frames, those tags are gone -- they seem to be removed.  Is there some default "VLAN action" which would do this?

So, to be clear, the frames are still there, they come in as ETHERTYPE 0x8100 (802.1Q) and leave as ETHERTYPE 0x8000 (IPv4) which is the encapsulated frame.  So _something_ removed the VLAN tag...!?

PR #667, opened 2021-07-13
feat: add proc events support

btwotch:


aboch:

vishvananda:
@btwotch Your new test appears to be timing out in CI. Can you take a look?

btwotch:
@vishvananda Thx for the hint; unfortunately the test passes locally; I added more debug output and pushed again - perhaps I now see more ...

vishvananda:
it could be an issue with the version of the kernel in travis? It would be nice if it would fail instead of hanging forever though. If we can validate that and get it to fail appropriately we could disable the test in travis if necessary

btwotch:
@vishvananda I tried out some more things ;-)
So, if I run only me new test locally, then it succeeds.
When I run all tests locally, then my test fails.

In strace I see:
Running only my test:
```149198 bind(6<NETLINK:[1386512]>, {sa_family=AF_NETLINK, nl_pid=0, nl_groups=0x000001}, 12 <unfinished ...>
149199 <... nanosleep resumed>NULL)     = 0
149198 <... bind resumed>)              = 0
149198 getpid()                         = 149198
149199 nanosleep({tv_sec=0, tv_nsec=5120000},  <unfinished ...>
149198 sendto(6<NETLINK:[CONNECTOR:149198]>, [{nlmsg_len=44, nlmsg_type=NLMSG_DONE, nlmsg_flags=0, nlmsg_seq=0, nlmsg_pid=149198}, "\x01\x00\x00\x00\x01\x00\x00\x00\x01\x00\x00\x00\x00\x00\x00\x00\x04\x00\x00\x00\x01\x00\x00\x00\x01\x00\x00\x00"], 44, 0, {sa_family=AF_NETLINK, nl_pid=0, nl_groups=0x000001}, 12) = 44
```

Running all tests:
```
148946 bind(25<NETLINK:[1382105]>, {sa_family=AF_NETLINK, nl_pid=0, nl_groups=0x000001}, 12) = 0
148946 getpid()                         = 148935
148946 sendto(25<NETLINK:[1382105]>, [{nlmsg_len=44, nlmsg_type=NLMSG_DONE, nlmsg_flags=0, nlmsg_seq=0, nlmsg_pid=148935}, "\x01\x00\x00\x00\x01\x00\x00\x00\x01\x00\x00\x00\x00\x00\x00\x00\x04\x00\x00\x00\x01\x00\x00\x00\x01\x00\x00\x00"], 44, 0, {sa_family=AF_NETLINK, nl_pid=0, nl_groups=0x000001}, 12) = -1 ECONNREFUSED (Connection refused)
```

vishvananda:
Possibly an issue because other tests are using network namespaces? This looks like it is possibly related? https://stackoverflow.com/questions/35987286/sendmsg-to-netlink-fails-with-econnrefused-if-application-is-running-in-lxc-cont

vishvananda:
If that is an issue, you maybe able to solve it by adding a `defer setUpNetlinkTest(t)()` at the beginning of your test, which lock the os thread and force a single namespace. Can you give that a try?

btwotch:
I am not sure, if I use `defer setUpNetlinkTest(t)()` correctly, but with it, it always fails.

btwotch:
`setUpNetlinkTest(t)` creates a new network namespace.

vishvananda:
Yes I wonder if you need to ensure sure the `false` command you are executing executes in the new network namespace as well.

btwotch:
But the `false` process is a child of it, so it should automatically be in the same network namespace.

I also tried out: `unshare -n ./netlink.test -test.v -test.timeout 60s -test.run TestSubscribeProcEvent`
And this one doesn't work either, while without `unshare -n` it works.

btwotch:
@vishvananda I added some code (in the test) to join the network namespace of PID 1 - do you think that's a way to go, or is it too much of a hack?

vishvananda:
Our CI was broken for a while since travis has not been helpful. I just switched over to github actions which means a new kernel and a new go version . It would be good to check if the hack is still necessary

vishvananda:
I think you would still need these three lines:
`	skipUnlessRoot(t)
	runtime.LockOSThread()
	defer runtime.UnlockOSThread()`

btwotch:
> whoops, youi need to uncomment the import for runtime

Aaah, sorry!

vishvananda:
looks like your workaround is still needed. If you want to add it back in, I'll merge it if it passes ci

btwotch:
> looks like your workaround is still needed. If you want to add it back in, I'll merge it if it passes ci

Nice! Thank you very much for your review!

vishvananda:
Yes your message size is incorrect
```
  fmt.Printf("%+v\n", binary.Size(x)+binary.Size(x.op)) // 28
  fmt.Printf("%+v\n", binary.Size(x))                   // 24
  fmt.Printf("%+v\n", binary.Size(x.CnMsg))             // 20
  fmt.Printf("%+v\n", binary.Size(x.op))                // 4
  ```

vishvananda:
This is how you would do it to match the way other serialization in nl is done: https://gist.github.com/vishvananda/ba20c1a97f3d06e9f0494c2f47d4c2ef

Would you mind applying that diff suashing and repushing?

btwotch:
> This is how you would do it to match the way other serialization in nl is done: https://gist.github.com/vishvananda/ba20c1a97f3d06e9f0494c2f47d4c2ef
> 
> Would you mind applying that diff suashing and repushing?

Thank you for your help; applied and built.

PR #669, opened 2021-07-22
Define more for conntact_unspecified

sc68cal:
This allows compilation on non-linux hosts for the containerlab project

### DIFF
--- a/conntrack_unspecified.go
+++ b/conntrack_unspecified.go
@@ -2,8 +2,14 @@
 
 package netlink
 
+import "net"
+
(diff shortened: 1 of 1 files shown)


sc68cal:
It's pretty ugly because i don't program in Go very much, but this is something I hacked up, to try and get containerlab to compile on OS X so I could hack at it some more

aboch:

vishvananda:
Some of this stubbing seems incorrect. The linux specific constants should not be in the unspecified file. Having unimplemented method calls seems fine, but the calling code should be updated to not depend on linux constants in an osx build.

PR #674, opened 2021-08-11
Use SyscallConn() instead of Fd() to persist TUN/TAP interface

montag451:
os.File.Fd() puts back the file descriptor in blocking mode which is
pretty annoying as SetDeadline will stop working


aboch:

montag451:

Done!

PR #687, opened 2021-08-26
Add mode/type support for IP tunnels using IPv4 addresses

byteocean:
The code supporting tunnel mode (see: https://man7.org/linux/man-pages/man8/ip-tunnel.8.html) configuration exists for IP tunnels using IPv6 addresses, but not exist for the ones using IPv4 addresses. This patch provides such support, and the experiment shows it works.

### DIFF
--- a/link.go
+++ b/link.go
@@ -1036,6 +1036,7 @@ type Iptun struct {
 	EncapType  uint16
 	EncapFlags uint16
 	FlowBased  bool
+	Proto      uint8
 }
 
 func (iptun *Iptun) Attrs() *LinkAttrs {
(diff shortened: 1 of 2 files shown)


aboch:

vishvananda:
also please squash into a single commit. Thanks!

byteocean:
> also please squash into a single commit. Thanks!
 Both requests are done. If anything more needs to be done, pl let me know. thanks!

hardikdr:
@vishvananda  Can you please review the PR again? 
Thanks in advance.  :)

vishvananda:
This looks correct. Can you try adding the proto value to a test in link_test.go and check it so we can be sure it is transiting the kernel correctly?

hardikdr:
cc @byteocean

PR #688, opened 2021-08-26
`Rule.String()` method returns `.Dst`

qdm12:
- Return `to <all | ipNetString>` in string
- Return `all` for `from` if `rule.Src == nil`
- Return `all` for `to` if `rule.Dst == nil`

### DIFF
--- a/rule.go
+++ b/rule.go
@@ -28,7 +28,18 @@ type Rule struct {
 }
 
 func (r Rule) String() string {
-	return fmt.Sprintf("ip rule %d: from %s table %d", r.Priority, r.Src, r.Table)
+	from := "all"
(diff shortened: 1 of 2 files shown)


aboch:

aboch:
Please squash your commits into one then force push. (This repo does not have "Squash-Merge" option)

qdm12:
> Please squash your commits into one then force push.

Done 😉

> This repo does not have "Squash-Merge" option enabled

Why not out of curiosity 🤔 ?

aboch:
>Why not out of curiosity thinking ?

It was like this since I came around here.

I think Vish wanted to have a clear 1-to-1 mapping between commit and PR. Which I also like.

PR #693, opened 2021-09-13
Add trailing null char to string in devlink request

zhangzujian:
Fix function `getEswitchAttrs` by adding trailing null char to string in devlink request.

Related PR: #651 .


aboch:

PR #702, opened 2021-10-14
Support for devlink info command

ipatrykx:
This PR introduces support for [devlink info command](https://www.kernel.org/doc/html/latest/networking/devlink/devlink-info.html). This command can be used by NIC manufacturers to report hardware info such as firmware version, device serial number and much more.


vishvananda:
lgtm. Thanks for the test

Issue #710, opened 2021-10-29
Add Rule.Type to allow adding/listing unreachable (RTN_UNREACHABLE) rules?


bradfitz:
The Rule type has no `Type` field or similar to permit creating RTN_UNREACHABLE rules.

Looking at `func ruleHandle`, there doesn't seem to be a way to set the `msg.Type` to anything other than `unix.RTN_UNSPEC`, `unix.RTN_UNICAST`, or `nl.FR_ACT_GOTO`.

PR #711, opened 2021-10-29
Add Rule.Type to support RTN_UNREACHABLE etc rules.

bradfitz:
Fixes vishvananda/netlink#710

### DIFF
--- a/rule.go
+++ b/rule.go
@@ -3,6 +3,8 @@ package netlink
 import (
 	"fmt"
 	"net"
+
+	"golang.org/x/sys/unix"
 )
 
 // Rule represents a netlink rule.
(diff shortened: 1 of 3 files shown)


bradfitz:
@vishvananda, thoughts?

We need this for @Tailscale in https://github.com/tailscale/tailscale/pull/3218 to drop our shelling out to the `ip` command.

vishvananda:
This is also causing a test failure:
`TestRuleAddDel:  rule_test.go:52: Rule has diffrent options than one added`

bradfitz:
We no longer need this (we went a different direction) so I'll just close this rather than try to remember the state of it. Feel free to take it over if it's something you want, though.

PR #716, opened 2021-11-24
Fix Travic CI integration

aboch:


aboch:
These changes are already in master.

PR #723, opened 2022-01-05
Add support for VCAN

maxcleme:
Dummy strategy in order to make `vcan` interpreted as `can`.

Does it sounds acceptable ? Or would you rather use a dedicated struct for handling `vcan` interfaces ?

Cheers.

### DIFF
--- a/link_linux.go
+++ b/link_linux.go
@@ -1853,6 +1853,8 @@ func LinkDeserialize(hdr *unix.NlMsghdr, m []byte) (Link, error) {
 						link = &Tuntap{}
 					case "ipoib":
 						link = &IPoIB{}
+					case "vcan":
+						fallthrough
 					case "can":
 						link = &Can{}
 					case "bareudp":


maxcleme:
@vishvananda 

Afaik `vcan` is a subset of `can`, options like `bitrate` cannot be set on `vcan` interface.

I also could add an empty struct for handling `vcan`, as you wish.

vishvananda:
We prefer to keep the main branch clean without merge commits. Can you rebase your changes please?

aboch:
@maxcleme please fix your branch's history

maxcleme:
@aboch @vishvananda Looks like I miss that comment from March 🤦 

PR has been rebased!

aboch:
What is this change going to give us?
Ability to parse a `vcan` type link as `can`, but no ability to program a `vcan` type link.

maxcleme:
@aboch Agreed, my use case was only reading CAN interface, however when debugging I was setting up VCAN interface directly on my machine instead of relying on hardware. I used my fork for doing such.

Feel free to close if you think programability should come with this PR.

aboch:
> Feel free to close if you think programability should come with this PR.

Or feel free to improve this change and add the full capability

maxcleme:
@aboch To be honest I won't 😅

PR #727, opened 2022-01-17
Only set XFRMA_IF_ID if not 0

tklauser:
A value of 0 for XFRMA_IF_ID is not valid and will be rejected as of
Linux kernel commit torvalds/linux@68ac0f3810e7 ("xfrm: state and policy
should fail if XFRMA_IF_ID 0"), which will return EINVAL when setting
XFRMA_IF_ID = 0. Thus, only set the XFRMA_IF_ID attribute in case
XfrmState.Ifid or XfrmPolicy.Ifid are != 0.


aboch:
LGTM

PR #729, opened 2022-01-25
add build on macOS test

twpayne:
netlink is Linux-only, but adding this test ensures that netlink builds
without error on macOS, which helps catch missing build tags.

cc @aditighag who found the problem and is submitting a PR to fix the missing build tags.


aditighag:
Thanks, Tom. Here is the fix - https://github.com/vishvananda/netlink/pull/730.

vishvananda:
included in #730

twpayne:
Thanks for the fast response 👍

PR #730, opened 2022-01-25
Fix darwin target compilation

aditighag:
dd687eb broken compilation on mac OS (darwin target). Rename the
linux specific files accordingly.

Fixes: dd687eb


aditighag:
@riccardomanfrin @aboch

vishvananda:
The macos build doesn't actually do anything, but I suppose it is fine to have the test in there so we remember to annotate the linux build files.

PR #731, opened 2022-01-30
Annotate errors using netlink error message.

yzp0n:
Background: I was having trouble finding out why an XDP program doesn't attach to a link on a particular host. If I can log the netlink error message like `ip` command does, it will be much easier to debug issues.

This PR consists of two patches:
- `setsockopt` the netlink socket, to inform the kernel that it should respond with the extended ACK TLVs
- `NetlinkRequest.Execute`'s response message handler logic modified, so it can parse and process the `nlmsgerr` attributes.

I'd appreciate your thoughts on this :)

### DIFF
--- a/nl/nl_linux.go
+++ b/nl/nl_linux.go
@@ -39,6 +39,9 @@ var nextSeqNr uint32
 // Default netlink socket timeout, 60s
 var SocketTimeoutTv = unix.Timeval{Sec: 60, Usec: 0}
 
+// ErrorMessageReporting is the default error message reporting configuration for the new netlink sockets
+var EnableErrorMessageReporting bool = false
+
 // GetIPFamily returns the family type of a net.IP.
 func GetIPFamily(ip net.IP) int {
 	if len(ip) <= net.IPv4len {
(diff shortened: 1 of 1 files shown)


yzp0n:
I believe I'm still encountering CI failures, but I'm having trouble reproducing them on local Linux env. Any help is appreciated.

yzp0n:
Another ping on this.
@vishvananda @aboch Can I have your review?

vishvananda:
Hi @yzp0n . A couple of things:
1. Can you squash the changes into a single commit?
2. I'm a little concerned that you had to change the error type check on the test. This signals to me that we might break existing code with this change. For this reason I think it might be better to set extended error reporting to false by default and document how to turn it on. We could switch the default to true in the future as part of a version bump of the library.

yzp0n:
Thanks for your review @vishvananda!
Would you mind taking another look?

> 1. Can you squash the changes into a single commit?

Done.

> 2. I'm a little concerned that you had to change the error type check on the test. This signals to me that we might break existing code with this change. For this reason I think it might be better to set extended error reporting to false by default and document how to turn it on. We could switch the default to true in the future as part of a version bump of the library.

Sure. I've turned `EnableErrorMessageReporting` default to false for now. This makes the test changes no longer necessary.

PR #734, opened 2022-02-18
sources: add SetInet6AddrGenMode() func

zemeteri:
Signed-off-by: Dmitry Petrov <dpetrov@nfware.com>


link_linux.go
```diff
@@ -3462,3 +3462,41 @@ func parseBareUDPData(link Link, data []syscall.NetlinkRouteAttr) {
...
 }
+
+func SetInet6AddrGenMode(link Link, mode int) error {
+	if mode < nl.IN6_ADDR_GEN_MODE_EUI64 || mode > nl.IN6_ADDR_GEN_MODE_RANDOM {
+		return fmt.Errorf("invalid addrgenmode: %d", mode)
+	}
+	return pkgHandle.SetInet6AddrGenMode(link, mode)
+}
+
+func (h *Handle) SetInet6AddrGenMode(link Link, mode int) error {
```
vishvananda, line 3473: Am I missing something here? Why have both a public and private version of this method if the public just delegates directly to the private?

vishvananda (changes requested):
Style nit inline.

link_linux.go
zemeteri: I've just done as I noticed about functions LinkSetARPOn()/LinkSetARPOff(), LinkSetAllmulticastOn()/LinkSetAllmulticastOff(), etc. I am newbie in Go language, so I've just tried to do as most of the functions int this file almost written. I have no idea why so. If I can do it better way - I would be pleased if You explain me the magic :)

PR #735, opened 2022-02-19
Only set IFLA_XFRM_IF_ID if not 0

tklauser:
A value of 0 for IFLA_XFRM_IF_ID is not valid and will be rejected by
the kernel by returning EINVAL, see [1]. Thus, only set the
IFLA_XFRM_IF_ID attribute in case Xfrmi.Ifid != 0.

[1] https://github.com/torvalds/linux/blob/8efd0d9c316af470377894a6a0f9ff63ce18c177/net/xfrm/xfrm_interface.c#L645-L648


tklauser:
@vishvananda @aboch could one of you please take a look?

aboch (approved):

PR #741, opened 2022-03-04
addr: add receive timeout option for AddrSubscribe

dcbw:
Otherwise addrSubscribeAt()'s goroutine will block in s.Receive()
and potentially call callbacks long after the done channel has
been closed by the caller.


tssurya (commented):
LGTM

Issue #751, opened 2022-04-02
`setsockopt` returns `ENOPROTOOPT` for `NETLINK_EXT_ACK`


hmahmood:
This is on older kernels, specifically centos 7.6. Line: https://github.com/vishvananda/netlink/blob/main/nl/nl_linux.go#L801

PR #752, opened 2022-04-02
Ignore ENOPROTOOPT error when setting NETLINK_EXT_ACK to 0

hmahmood:
Fix for https://github.com/vishvananda/netlink/issues/751

### DIFF
--- a/nl/nl_linux.go
+++ b/nl/nl_linux.go
@@ -798,7 +798,13 @@ func (s *NetlinkSocket) SetExtAck(enable bool) error {
 		enableN = 1
 	}
 
-	return unix.SetsockoptInt(int(s.fd), unix.SOL_NETLINK, unix.NETLINK_EXT_ACK, enableN)
+	var err error
+	if err = unix.SetsockoptInt(int(s.fd), unix.SOL_NETLINK, unix.NETLINK_EXT_ACK, enableN); err == unix.ENOPROTOOPT && !enable {
+		// older kernels may not support NETLINK_EXT_ACK
+		return nil
+	}
+
+	return err
 }
 
 func (s *NetlinkSocket) GetPid() (uint32, error) {


nl/nl_linux.go
```diff
@@ -798,7 +798,13 @@ func (s *NetlinkSocket) SetExtAck(enable bool) error {
 		enableN = 1
 	}
 
-	return unix.SetsockoptInt(int(s.fd), unix.SOL_NETLINK, unix.NETLINK_EXT_ACK, enableN)
+	var err error
+	if err = unix.SetsockoptInt(int(s.fd), unix.SOL_NETLINK, unix.NETLINK_EXT_ACK, enableN); err == unix.ENOPROTOOPT && !enable {
```
aboch, line 802: Why suppressing the error only in the `enable == false` case?

hmahmood: If the caller requires this option to present, i.e. they are setting to true, then it is important to communicate the error back. In the false case, since they don't want the option anyway, the error doesn't matter.

vishvananda:
Now that I think about it, I think the original logic for this feature is wrong. If the error checking isn't enabled, we shouldn't be setting the sock option at all, instead of trying to turn off the option and erroring. Ithink I'd prefer this changes to simply not call the method if enabled is false. I know this means you can't turn off the feature later but doing this via a package level variable is the wrong approach if we want to allow people to toggle it on and off.

yzp0n:
I think we can change

https://github.com/vishvananda/netlink/blob/main/nl/nl_linux.go#L507 to
```go
if EnableErrorMessageReporting {
		if err := s.SetExtAck(true); err != nil {
			return nil, err
		}
}
```

I believe this is inline with what @vishvananda is suggesting earlier.

vishvananda:
I accidentally merged this without asking for a squash first. Going to repush it with a squashed version

vishvananda:
New commit is here: https://github.com/vishvananda/netlink/commit/5e915e0149386ce3d02379ff93f4c0a5601779d5

hmahmood:
Thanks @vishvananda

PR #753, opened 2022-04-08
Add ReceiveBufferSize to LinkSubscribeOptions

danobi:
Sometimes with `ListExisting` enabled and enough interfaces configured,
netlink will try to send more data than the buffer can hold and the
caller will get back ENOBUFS. Thus, it's useful to be able to configure
the netlink buffer size.

### DIFF
--- a/link_linux.go
+++ b/link_linux.go
@@ -2069,21 +2069,22 @@ type LinkUpdate struct {
 // LinkSubscribe takes a chan down which notifications will be sent
 // when links change.  Close the 'done' chan to stop subscription.
 func LinkSubscribe(ch chan<- LinkUpdate, done <-chan struct{}) error {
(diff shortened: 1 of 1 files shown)


aboch:
Sorry, I do not see why this change is needed.

This library already gives you a package level function to set the receiver buffer size for all the sockets:


image

Your project code just has to use it before calling `LinkSubscribe()`.

danobi:
Hi @aboch , sorry about the much delayed reply.

`SetSocketReceiveBufferSize()` requires a pointer receiver. There is currently no way (as far as I can tell) to access the `pkgHandle` defined in handle_linux.go.

Maybe instead we can add a getter to return the package handle? What do you think?

aboch:
pls fix the conflict

danobi:
This was merged in with  #773

PR #760, opened 2022-04-28
rule: add support for uidrange

manjirigadagkar:


rule.go
```diff
@@ -26,6 +26,7 @@ type Rule struct {
 	Dport             *RulePortRange
 	Sport             *RulePortRange
 	IPProto           int
+	UID               *RuleUIDRange
```
aboch, line 29: Can you please point me to the respective iproute2 cli command option?
I want to understand whether this should be simply called UID even when it is actually specifying a range of uids

aboch: Thanks @manjirigadagkar 

Given 
```
              uidrange NUMBER-NUMBER
                     select the uid value to match.
```

Can we rename `UID` to `UIDRange` or similar?

manjirigadagkar: I wonder if that won't make it incongruent with `Dport` and `Sport` which are also `RulePortRange` type.

manjirigadagkar: `Dport             *RulePortRange`

vs

`UIDRange               *RuleUIDRange`

Wouldn't it be confusing to reader why, when both are "range" types, only `UIDRange` is named this way? One might wonder, why isn't the name `DportRange`? Your thoughts?

This is why I chose `UID` over `UIDRange` in the beginning.

Your feedback indicates that the preference is to match CLI naming convention despite the fact that `Dport` and `Sport` do not support CLI's NUMBER style input, only NUMBER-NUMBER i.e. range input. Can you confirm before I proceed with the changes?

aboch: I agree with you. Because of the previous logic chosen for the ports fields.

Solution would be to define new `DPortRange` and `SPortRange` along with existing ones and deprecate old ones.

And we would need to handle the case where caller mistakenly sets both.

aboch: > I have renamed UID to UIDRange.

Thank you.
Please squash the two commits in one. This project does not have the Squash and Merge enabled.

manjirigadagkar: Done!

aboch:
LGTM

Issue #477, opened 2019-08-20
"No buffer space available" error on large queries


diconico07:
Large queries tends to fill the socket receive buffer causing the receive on socket to raise an `ENOBUFS` meaning some messages have been dropped due to lack of space.

This can be solved by increasing the `SO_RCVBUF` socket option value (as the relevant value will vary between uses of the library it would be best IMHO to make this value configurable).

aboch:
At least for the handle APIs, it should be possible to set now:
https://github.com/vishvananda/netlink/blob/f504738125a57f35f87fc30fb69b8df75237ccde/handle_linux.go#L61

diconico07:
Got the issue on a subscribe call, and expected to see a method on NetlinkSocket so missed this one. So we only need a configuration knob for subscribe calls for subscribe calls , and I see there is already one for AddrSubscribe but not for NeighSubscribe nor LinkSubscribe

PR #773, opened 2022-06-03
Add ReceiveBufferSize, ReceiveBufferSizeForce, and ReceiveTimeout options to *Subscribe functions

daveset:
We were seeing NeighSubscribe with ListExisting failing with ENOBUFS errors as described in Issue #477. In our case we initially were not handling the error properly and the goroutine died leaving another channel without a reader causing the application to hang.  

This PR incorporates the LinkSubscribe fix from https://github.com/danobi in PR https://github.com/vishvananda/netlink/pull/753 and adds another commit to: 

- Apply subscribe options consistently across AddrSubscribe, NeighSubscribe, LinkSubscribe, and RouteSubscribe functions. 
- Support ReceiveBufferSize, ReceiveBufferSizeForce, and ReceiveTimeout options
- Only set the receive buffer size on the specific socket associated with the subscription (vs. using pkghandle)

Although the go code is pretty straightforward, I am not intimately familiar with Netlink and all of its interactions, so please let me know if I've misinterpreted something. In particular i'm not sure I fully understand the use of pkghandle and if my choice to set the ReceiveBufferSize on the specific socket is correct.

I have not had an opportunity to write generic unit tests for this functionality. It is running successfully in our application where we are detecting the ENOBUFS failure and then re-subscribing with the ReceiveBufferSize doubled (using an initial default matching the /proc/sys/net/core/rmem_default value on our system) until it is successful or it exceeds a max buffer size constant (5 MB).

### DIFF
--- a/addr_linux.go
+++ b/addr_linux.go
@@ -296,23 +296,24 @@ type AddrUpdate struct {
 // AddrSubscribe takes a chan down which notifications will be sent
 // when addresses change.  Close the 'done' chan to stop subscription.
 func AddrSubscribe(ch chan<- AddrUpdate, done <-chan struct{}) error {
-	return addrSubscribeAt(netns.None(), netns.None(), ch, done, nil, false, 0, nil)
+	return addrSubscribeAt(netns.None(), netns.None(), ch, done, nil, false, 0, nil, false)
 }
 
 // AddrSubscribeAt works like AddrSubscribe plus it allows the caller
 // to choose the network namespace in which to subscribe (ns).
 func AddrSubscribeAt(ns netns.NsHandle, ch chan<- AddrUpdate, done <-chan struct{}) error {
-	return addrSubscribeAt(ns, netns.None(), ch, done, nil, false, 0, nil)
+	return addrSubscribeAt(ns, netns.None(), ch, done, nil, false, 0, nil, false)
 }
 
 // AddrSubscribeOptions contains a set of options to use with
 // AddrSubscribeWithOptions.
 type AddrSubscribeOptions struct {
-	Namespace         *netns.NsHandle
-	ErrorCallback     func(error)
-	ListExisting      bool
-	ReceiveBufferSize int
-	ReceiveTimeout    *unix.Timeval
+	Namespace              *netns.NsHandle
+	ErrorCallback          func(error)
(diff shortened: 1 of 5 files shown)


daveset:
Hi @aboch. I'm looking to get our fork in sync with upstream and this and associated PR #753 are still open. Can you take another look at these PRs that take the solution to ENOBUFS errors on subscribe operations previously implemented on `AddrSubscribeOptions` and extend it to the other Subscribe functions? 

Note that the suggestion in #753 to use `func (h *Handle) SetSocketReceiveBufferSize(size int, force bool) error` does not work here because the Subscribe functions don't operate on socket associated with a Handle but instead create and manage the lifespan of their own socket via `nl.SubscribeAt`

aboch:
lgtm

tklauser:
This PR seems to have broken `TestRouteExtraFields` on `main`, see https://github.com/vishvananda/netlink/actions/runs/6269527708/job/17026085340

aboch:
Thanks @tklauser 

@daveset  Can you please take a look?

daveset:
@aboch At first glance it's not clear how the changes in this PR could affect the `TestRouteExtraFields` test, but i'll dig into it and see what I can find.

daveset:
@aboch It looks to me like the Github Actions image bumped the kernel version from 5.15.x to 6.2.x in https://github.com/actions/runner-images/commit/514557cf72d78ca4ce87e6d91a32f4e09f2fa3aa#diff-d3f496e99457b699df2a5621c8f59cb52f60a050e1a88318af7d6f7069c12b2eR9 and this is likely causing the CI failure in `TestRouteExtraFields`.

This is consistent with my testing tonight on an Ubuntu 22.04 with kernel 5.15.0-25 where it passes, vs testing on a Manjaro workstation with kernel 6.1.53 that fails the same test.

tklauser:
Apologies for blaming the test failure on this PR. I've traced the `TestRouteExtraFields` failure down to kernel commit https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=f55fbb6afb8d which started rejecting TOS values with ECN bits set and openend a PR fixing the failing tests: https://github.com/vishvananda/netlink/pull/921

PR #777, opened 2022-06-20
Add IpsetSwap()

damz:
This implements the `IPSET_CMD_SWAP` command as `IpsetSwap()`. That command swaps two ipsets atomically.


aboch:
LGTM

Issue #542, opened 2020-05-25
*Subscribe* family of functions leaks goroutines


jncornett:
Contingent on how long it takes to actually start reading from the socket fd, the following code will never return:

```
package main

import (
	"time"

	"github.com/vishvananda/netlink"
)

func main() {
	updates := make(chan netlink.AddrUpdate)
	done := make(chan struct{})
	netlink.AddrSubscribe(updates, done)
	time.Sleep(time.Second)
	close(done)
	<-updates
}
```

This is apparently because:
1. there is no cancellation mechanism in the call to `s.Receive()`.
2. the initialization code does not set a read timeout for the socket (via, e.g. setsockopt).

This isn't a big deal for short-lived programs (as the goroutines will get cleaned up on exit), but I was writing a long lived daemon that creates and removes subscriptions based on configuration file updates. With the current behavior, it is not possible to write such a program without leaking goroutines via the above mechanism.

jncornett:
I don't have the context to know if this makes sense for all use cases, but a simple fix would be to duplicate `nl.NetlinkSocket.Receive` to a new method `nl.NetlinkSocket.ReceiveWithTimeout`. Inside of this new method set a userspace timeout via `unix.Select(...)` immediately prior to the call to `unix.Recvfrom(...)`. The `*Subscribe*` family of functions could use this new `ReceiveWithTimeout` method in place of `Receive`. If the return value of the error is `Timeout` (I know, checking error return values is anti-pattern), then the loop would just `continue`. As far as I know, this seems like the least invasive change to "cancel" a `recvfrom()`/`recvmsg()` call.

vincentbernat:
`unix.RecvFrom()` can be unblocked by calling `unix.Shutdown()` with `unix.SHUT_RD` (or `unix.SHUT_RDWR` if you also want to unblock `unix.SendTo()`).

vincentbernat:
I didn't test, but I think this patch would help:

```diff
diff --git a/nl/nl_linux.go b/nl/nl_linux.go
index 600b942b1785..72104b93ef2b 100644
--- a/nl/nl_linux.go
+++ b/nl/nl_linux.go
@@ -734,6 +734,7 @@ func SubscribeAt(newNs, curNs netns.NsHandle, protocol int, groups ...uint) (*Ne

 func (s *NetlinkSocket) Close() {
        fd := int(atomic.SwapInt32(&s.fd, -1))
+       unix.Shutdown(fd, unix.SHUT_RDWR)
[...]

PR #779, opened 2022-06-26
Add a test to check if we can cleanly close subscribe goroutine

vincentbernat:
This is for #542. It seems to work just fine this way. If not, this
should always work with the following patch:

```diff
diff --git a/nl/nl_linux.go b/nl/nl_linux.go
index 600b942b1785..72104b93ef2b 100644
--- a/nl/nl_linux.go
+++ b/nl/nl_linux.go
@@ -734,6 +734,7 @@ func SubscribeAt(newNs, curNs netns.NsHandle, protocol int, groups ...uint) (*Ne

 func (s *NetlinkSocket) Close() {
        fd := int(atomic.SwapInt32(&s.fd, -1))
+       unix.Shutdown(fd, unix.SHUT_RDWR)
        unix.Close(fd)
 }
```

Let's see the result of the tests.

### DIFF
--- a/nl/nl_linux.go
+++ b/nl/nl_linux.go
@@ -734,6 +734,7 @@ func SubscribeAt(newNs, curNs netns.NsHandle, protocol int, groups ...uint) (*Ne
 
 func (s *NetlinkSocket) Close() {
 	fd := int(atomic.SwapInt32(&s.fd, -1))
+	unix.Shutdown(fd, unix.SHUT_RDWR)
 	unix.Close(fd)
 }
 
--- a/route_test.go
+++ b/route_test.go
@@ -355,6 +355,24 @@ func TestRouteSubscribe(t *testing.T) {
 	}
 }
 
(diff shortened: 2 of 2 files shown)


vincentbernat:
Humm, this does not seem to work. I would need to investigate why.

Hendrik-H:
The behavior is discussed in https://stackoverflow.com/questions/6389970/unblock-recvfrom-when-socket-is-closed . Even though the accepted answer looks pretty much the same as this change there is a comment that there would still be a race condition.
According to https://man7.org/linux/man-pages/man2/close.2.html#:~:text=Multithreaded%20processes%20and%20close() the `close()` call is expected to hang on Linux.

A solution might be to use `EpollWait` before the actual `Recvfrom` call. This way it should be possible to prevent long hangs.

vincentbernat:
Looking at Linux code, `shutdown()` should unblock receiving. However, for Netlink, shutdown is not supported.

PR #784, opened 2022-07-04
Add labelFilter for conntrack

tssurya:
This PR adds support for filtering flows
based on conntrack labels. It adds two
filters `ConntrackMatchLabels` &&
`ConntrackUnmatchLabels` through which user can
provide a list of labels as type "bytes" which
will then be compared to flow.Labels to see if
any matches were found.

Since netfilter conntrack labels is just a 128 bit/16bytes
variable, its possible to store multiple labels in that field.
Hence we allow passing multiple labels for comparison

`ConntrackMatchLabels`: Every label passed should
be contained in flow.Labels for a match to be true

`ConntrackUmmatchLabels`: Every label passed should
not be contained in the flow.Labels for a match to
be true

### DIFF
--- a/conntrack_linux.go
+++ b/conntrack_linux.go
@@ -149,20 +149,26 @@ type ConntrackFlow struct {
 	TimeStart  uint64
 	TimeStop   uint64
 	TimeOut    uint32
+	Labels     []byte
 }
 
 func (s *ConntrackFlow) String() string {
 	// conntrack cmd output:
-	// udp      17 src=127.0.0.1 dst=127.0.0.1 sport=4001 dport=1234 packets=5 bytes=532 [UNREPLIED] src=127.0.0.1 dst=127.0.0.1 sport=1234 dport=4001 packets=10 bytes=1078 mark=0
+	// udp      17 src=127.0.0.1 dst=127.0.0.1 sport=4001 dport=1234 packets=5 bytes=532 [UNREPLIED] src=127.0.0.1 dst=127.0.0.1 sport=1234 dport=4001 packets=10 bytes=1078 mark=0 labels=0x00000000050012ac4202010000000000
 	//             start=2019-07-26 01:26:21.557800506 +0000 UTC stop=1970-01-01 00:00:00 +0000 UTC timeout=30(sec)
 	start := time.Unix(0, int64(s.TimeStart))
 	stop := time.Unix(0, int64(s.TimeStop))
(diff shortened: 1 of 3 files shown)


conntrack_linux.go
```diff
@@ -474,10 +491,22 @@ func (f *ConntrackFilter) AddProtocol(proto uint8) error {
 	return nil
 }
 
+// AddLabel adds the label value to the conntrack filter
+func (f *ConntrackFilter) AddLabel(tp ConntrackFilterType, labels [][]byte) error {
```
dcbw, line 495: @tssurya The comment seems a bit misleading; it actually adds either zero or more label values right? Should it be `AddLabels` instead?

conntrack_linux.go
```diff
@@ -531,6 +560,29 @@ func (f *ConntrackFilter) MatchConntrackFlow(flow *ConntrackFlow) bool {
...
+			// --label label1,label2 in conn entry;
+			// every label passed should be not contained in flow.Labels for a match to be true
+			if elem, found := f.labelFilter[ConntrackUnmatchLabels]; match && found {
+				for _, label := range elem {
+					match = match && !(bytes.Contains(flow.Labels, label))
+				}
+			}
+		} else {
+			// flow doesn't contain labels, so it doesn't contain or notContain any provided matches
+			match = false
```
dcbw, line 582: @tssurya shouldn't this case of zero labels in the flow also match if `len(f.labelFilter) == 1` and that single item is `ConntrackUnmatchLabels` ? Because by definition zero flow labels won't match anything.

conntrack_linux.go
tssurya: oh good point! let me fix this; yes it should be `AddLabels`

conntrack_linux.go
tssurya: ack thanks Dan, let me update that function!

tssurya:
@aboch @vishvananda can you please take a look?

dcbw (approved):
Looks good to me.

tssurya:
hey @aboch or @vishvananda Is there something we could do to get some eyes on this PR? Any advice is appreciated.

tssurya:
@aboch or @vishvananda : PTAL

tssurya:
Wondering if there is something I can do from my side to get some reviews on this, we are currently vendoring a fork with this PR and we'd like to stop doing that and try and see if we can get this in.

PR #788, opened 2022-07-21
Add support for IPv6 IPsets

damz:
Add support for the IPv6 family of IPsets.

This requires two changes:

* Adding an explicit `Family` property in `IpsetCreateOptions`, which defaults to `unix.AF_INET` (i.e. ipv4) for backwards compatibility;
* Fixing the serialization/deserialization of `IPSetEntry.IP` and `IPSetEntry.IP2`

The serialization currently confuses `IPSET_ATTR_IP` / `IPSET_ATTR_IP2` (the attribute identifier) with `IPSET_ATTR_IPADDR_IPV4` / `IPSET_ATTR_IPADDR_IPV6` (the type of the attribute). The only reason it works right now is that both `IPSET_ATTR_IP` and `IPSET_ATTR_IPADDR_IPV4` are 1.

### DIFF
--- a/ipset_linux.go
+++ b/ipset_linux.go
@@ -67,6 +67,7 @@ type IpsetCreateOptions struct {
 	Comments bool
 	Skbinfo  bool
 
+	Family   uint8
 	Revision uint8
 	IPFrom   net.IP
 	IPTo     net.IP
@@ -158,7 +159,10 @@ func (h *Handle) IpsetCreate(setname, typename string, options IpsetCreateOption
 		data.AddChild(nl.NewRtAttr(nl.IPSET_ATTR_PORT_FROM|int(nl.NLA_F_NET_BYTEORDER), buf[:2]))
 		data.AddChild(nl.NewRtAttr(nl.IPSET_ATTR_PORT_TO|int(nl.NLA_F_NET_BYTEORDER), buf[2:]))
 	default:
-		family = unix.AF_INET
+		family = options.Family
+		if family == 0 {
+			family = unix.AF_INET
+		}
 	}
 
 	req.AddData(nl.NewRtAttr(nl.IPSET_ATTR_FAMILY, nl.Uint8Attr(family)))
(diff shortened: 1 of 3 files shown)


damz:
@aboch Any feedback on this?

aboch:
>  The only reason it works right now is that both IPSET_ATTR_IP and IPSET_ATTR_IPADDR_IPV4 are 1.

:-))

ipset_linux.go
```diff
@@ -249,6 +254,20 @@ func (h *Handle) IpsetDel(setname string, entry *IPSetEntry) error {
 	return h.ipsetAddDel(nl.IPSET_CMD_DEL, setname, entry)
 }
 
+func encodeIP(ip net.IP) (*nl.RtAttr, error) {
+	typ := int(nl.NLA_F_NET_BYTEORDER)
+	switch len(ip) {
+	case 16:
```
aboch, line 260: A `net.IP` variable populated via `net.ParseIP()` will be 16 bytes long even if it is parsing a IPv4 in string form.
This check may be weak.

aboch: One better option is to test the o/p of `func (ip IP) To4() IP`

damz: Indeed, and the reason that it works right now is because the tests have explicit `.To4()` calls all over. A nice side effect of this will be that entries containing IPv4 addresses that are not explicitly normalized will now work.

I updated the logic there, and removed the `.To4()` calls in the tests.

damz: @aboch Tests are green, back to you :)

aboch:
Thank you.

This project follows a "1 commit per PR" rule and GH "Squash-Merge" is not enabled:

image

damz:
@aboch It is a bit weird to squash two commits that are cleanly separate, but in any case, it is done.

aboch:
> @aboch It is a bit weird to squash two commits that are cleanly separate, but in any case, it is done.

Agree.
In this case second commit was minimal and somewhat related to the other commit.
In other cleanly separated commits, what we suggest is to have two distinct PRs.

PR #790, opened 2022-07-28
RouteGetWithOptions: Add UID option

manjirigadagkar:
Usage: 
       ip route get [ ROUTE_GET_FLAGS ] ADDRESS
                            [ from ADDRESS iif STRING ]
                            [ oif STRING ] [ tos TOS ]
                            [ mark NUMBER ] [ vrf NAME ]
                            **[ uid NUMBER ]** [ ipproto PROTOCOL ]
                            [ sport NUMBER ] [ dport NUMBER ]

Note: Current uid applies when the uid option is not used.

### DIFF
--- a/route_linux.go
+++ b/route_linux.go
@@ -1298,6 +1298,7 @@ type RouteGetOptions struct {
 	Oif     string
 	VrfName string
 	SrcAddr net.IP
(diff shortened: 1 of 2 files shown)


aboch:
LGTM

PR #793, opened 2022-07-31
Fix recvfrom goroutine leak

### LINKED ISSUE #792: routeSubscribeAt not closed
code:

```go
package main

import (
	"net/http"
	_ "net/http/pprof"
	"time"

)

func main() {
	ch := make(chan netlink.RouteUpdate, 2)
	done := make(chan struct{})
	netlink.RouteSubscribe(ch, done)
	time.Sleep(time.Second)
	close(done)
	println("closed")
	http.ListenAndServe(":8080", nil)
}
```

documentation says: `close the 'done' chan to stop subscription.`

pprof result:

Image

nekohasekai:
unix.Recvfrom does not return after fd close.

Fixes #792

### DIFF
--- a/nl/nl_linux.go
+++ b/nl/nl_linux.go
@@ -6,6 +6,7 @@ import (
 	"encoding/binary"
 	"fmt"
 	"net"
+	"os"
 	"runtime"
 	"sync"
 	"sync/atomic"
@@ -607,8 +608,9 @@ func NewNetlinkRequest(proto, flags int) *NetlinkRequest {
 }
 
 type NetlinkSocket struct {
-	fd  int32
-	lsa unix.SockaddrNetlink
+	fd   int32
+	file *os.File
+	lsa  unix.SockaddrNetlink
 	sync.Mutex
 }
 
@@ -617,8 +619,13 @@ func getNetlinkSocket(protocol int) (*NetlinkSocket, error) {
 	if err != nil {
 		return nil, err
 	}
+	err = unix.SetNonblock(fd, true)
+	if err != nil {
+		return nil, err
+	}
 	s := &NetlinkSocket{
-		fd: int32(fd),
+		fd:   int32(fd),
+		file: os.NewFile(uintptr(fd), "netlink"),
 	}
 	s.lsa.Family = unix.AF_NETLINK
 	if err := unix.Bind(fd, &s.lsa); err != nil {
@@ -703,8 +710,13 @@ func Subscribe(protocol int, groups ...uint) (*NetlinkSocket, error) {
 	if err != nil {
 		return nil, err
 	}
+	err = unix.SetNonblock(fd, true)
+	if err != nil {
+		return nil, err
+	}
 	s := &NetlinkSocket{
-		fd: int32(fd),
+		fd:   int32(fd),
+		file: os.NewFile(uintptr(fd), "netlink"),
 	}
 	s.lsa.Family = unix.AF_NETLINK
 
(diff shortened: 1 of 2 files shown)


nekohasekai:
About why TestIfSocketCloses passed:

`SetReceiveTimeout` make `Receive()` always return `unix.EAGAIN`

thaJeztah:
Looks like this one may have introduced a regression; see https://github.com/moby/moby/pull/46982#issuecomment-2315840500

> EINTR on netlink sockets is a new one. I suspect it has more to do with the netlink dependency bump you pulled in when rebasing than on the Go toolchain bump. I think the bug is here: https://github.com/vishvananda/netlink/blob/92645823f36c7ed03faf4baa566078d9d5e06fda/nl/nl_linux.go#L821-L824 It retries on `EWOULDBLOCK` (a.k.a. `EAGAIN`) but neglects to retry on `EINTR`.

database64128:
@thaJeztah IO calls on non-blocking sockets will never return `-EINTR`. The problem here is that Moby calls `SetSocketTimeout`, which sets `SO_SNDTIMEO` and `SO_RCVTIMEO`. These socket options are only useful for sockets in blocking mode. Setting these probably places the socket back into blocking mode.

https://github.com/moby/moby/blob/92a05cf4148adc26054a5b52b191e0b4d13112cd/libnetwork/ns/init_linux.go#L25-L40

https://github.com/vishvananda/netlink/blob/92645823f36c7ed03faf4baa566078d9d5e06fda/nl/nl_linux.go#L848-L860

I'm not very familiar with this project, but it seems to me that before this PR, only blocking IO is used. This PR adds calls to set the socket as non-blocking, but still allows setting the timeout socket options.

thaJeztah:
Thanks; let me check with the networking folks about those parts.

corhere:
Tracing through the kernel sources, I can confidently claim that the timeout sockopts have _no_ effect when the `O_NONBLOCK` fcntl is enabled. That's a red herring.
- [`O_NONBLOCK` on the socket forces the `MSG_DONTWAIT` flag to be set on all `recv[from]` syscalls](https://elixir.bootlin.com/linux/v6.10.6/source/net/socket.c#L2241)
- [`recv` on a netlink socket is implemented in terms of `skb_recv_datagram`](https://elixir.bootlin.com/linux/v6.10.6/source/net/netlink/af_netlink.c#L1927)
- [The implementation of `skb_recv_datagram`](https://elixir.bootlin.com/linux/v6.10.6/source/net/core/datagram.c#L293) only blocks if the timeout is non-zero

The timeout for a non-blocking socket would be zero due to the `MSG_DONTWAIT` flag being forced on; therefore setting the timeout sockopts must have no effect on the behavior of the `recv` syscall.

Need further proof? [The timeouts on new sockets are initialized to `LONG_MAX`.](https://elixir.bootlin.com/linux/v6.10.6/source/net/core/sock.c#L3491) And [there is no magic hiding in the `setsockopt` syscall.](https://elixir.bootlin.com/linux/v6.10.6/source/net/core/sock.c#L1357)

System calls that "shouldn't" return `-EINTR` have been known to return `-EINTR` in practice when a signal is delivered to the task while in a syscall. See e.g. https://go.dev/issue/38033. Maybe it's a kernel bug which results in a recv on a non-blocking socket returning with `-EINTR` upon receiving a signal despite the `SA_RESTART` sigaction when a blocking recv would have been implicitly restarted?

database64128:
This all still feels very strange to me.

I checked [`tokio`'s implementation of `TcpStream`](https://github.com/tokio-rs/tokio/blob/master/tokio/src/net/tcp/stream.rs) and followed the references all the way down to the libc calls. There's no handling of `-EINTR` (`std::io::ErrorKind::Interrupted`) at all. The [netlink-sys](https://github.com/rust-netlink/netlink-sys) crate's tokio socket implementation does not handle `-EINTR` either.

I actually [rolled my own](https://github.com/database64128/ddns-go/blob/main/producer/netlink/internal/rtnetlink/conn_linux.go) `rtnetlink` Go package a while ago, and it never had to deal with `-EINTR`.

corhere:
Unlike Go programs, Rust programs don't (inherently) [have a runtime which sends SIGURG to the process](https://go.dev/doc/go1.14#runtime) baked in.

database64128:
> Unlike Go programs, Rust programs don't (inherently) [have a runtime which sends SIGURG to the process](https://go.dev/doc/go1.14#runtime) baked in.

Tokio does allow you to [register for receiving signals](https://docs.rs/tokio/latest/tokio/signal/unix/struct.Signal.html). It's actually quite similar to `os/signal`. I imagine there are programs that use Tokio for network IO and receiving signals at the same time.

robmry:
This change breaks the timeouts, I can raise a PR to fix that.

But, the `EINTR` we've been seeing in moby CI is unrelated - it's due to https://github.com/vishvananda/netlink/pull/925, exposing an NLM_F_DUMP_INTR reported in the netlink response as an EINTR ... https://github.com/vishvananda/netlink/blob/92645823f36c7ed03faf4baa566078d9d5e06fda/nl/nl_linux.go#L582-L584

PR #794, opened 2022-08-02
"FlowBased" fix and addition for GRE

ebirger:
- Fix Gretap FlowBased support by not sending a payload when configuring the IFLA_GRE_COLLECT_METADATA attribute
- Add FlowBased support for Gretun


aboch:
This project follows a "one fix/addition per PR"
Please have two PRs, one for each commit.

Thank you

ebirger:
Will submit separately. Thanks!

PR #795, opened 2022-08-02
link: fix `FlowBased` support in Gretap

ebirger:
IFLA_GRE_COLLECT_METADATA is a "flag" netlink attribute. As such it should
not be sent with a data payload.

Fix gretap link to send this flag correctly and re-enable tests.

Signed-off-by: Eyal Birger <eyal.birger@gmail.com>


aboch:
LGTM

PR #798, opened 2022-08-05
Use IFLA_* consts defined in golang.org/x/sys/unix

tklauser:
Use the constants provided in the golang.org/x/sys/unix package instead
of locally defining them, as done in commit c94808a88ba4 ("link: add
support for IFLA_GRO_MAX_SIZE").


aboch:
LGTM

PR #806, opened 2022-08-22
Support skip_hw/skip_sw flags

e0ne:


nl/tc_linux.go
```diff
@@ -1022,6 +1029,9 @@ const (
 	__TCA_FLOWER_MAX
 )
 
+const TCA_CLS_FLAGS_SKIP_HW = 1 << 0 /* don't offload action to HW */
```
adrianchiris, line 1039: nit:
```
const (
  TCA_CLS_FLAGS_SKIP_HW = 1 << 0 /* dont .... */
  TCA_CLS_FLAGS_SKIP_SW = 1 << 1 /* dont ... */
)
```

also i think this applies to classification (filter) and not the action within a filter. so prehaps change the comment.

filter_linux.go
```diff
@@ -130,6 +132,22 @@ func (filter *Flower) encode(parent *nl.RtAttr) error {
 		parent.AddRtAttr(nl.TCA_FLOWER_KEY_ENC_KEY_ID, htonl(filter.EncKeyId))
 	}
 
+	if filter.SkipHw {
```
adrianchiris, line 135: should they both be added to the same TCA_FLOWER_FLAGS attribute ?

e0ne: What do you mean? We can't add both TCA_CLS_FLAGS_SKIP_HW and TCA_CLS_FLAGS_SKIP_SW at the same time

adrianchiris: Yes i realize its invalid config (having both set).

I think, in general, if there are multiple fields on the same attribute they should be set together, if its invalid then netlink(kernel) will fail it (i hope).

does it fail when both are set on the same attribute ?

nl/tc_linux.go
e0ne: > so prehaps change the comment.
what to you mean? the comment is copied from iprote2

filter_linux.go
e0ne: `~# tc filter add dev enp3s0f0np0 protocol ip parent ffff: flower ip_proto 3 skip_hw skip_sw action drop
RTNETLINK answers: Invalid argument
We have an error talking to the kernel`

nl/tc_linux.go
adrianchiris: https://github.com/shemminger/iproute2/blob/87dfa589f6c3e9ab1ac2bfb9b5f8591a60320103/include/uapi/linux/pkt_cls.h#L206

filter_linux.go
adrianchiris: ack, thx for clarifying. i see you also changed flags it to be on the same attr thx !

qdisc_linux.go
```diff
@@ -448,6 +451,11 @@ func (h *Handle) QdiscList(link Link) ([]Qdisc, error) {
 
 					// no options for ingress
 				}
+			case nl.TCA_INGRESS_BLOCK:
+				bitfield := nl.DeserializeUint32Bitfield(attr.Value)
```
adrianchiris, line 455: i think you just need to cast to uint32 here.

aboch:
@e0ne Please have only one commit per PR. Please open a new PR for your second fix, the qdisc ingress block

qdisc_linux.go
```diff
@@ -448,6 +452,10 @@ func (h *Handle) QdiscList(link Link) ([]Qdisc, error) {
 
 					// no options for ingress
 				}
+			case nl.TCA_INGRESS_BLOCK:
+				val := new(uint32)
+				*val = binary.LittleEndian.Uint32(attr.Value)
```
adrianchiris, line 457: you should use `netlink.native.Uint32`

nl/tc_linux.go
adrianchiris: @e0ne could you align comment with what is defined in iproute2 ? i added link in the comment above.
also while at it, maybe change the const to be a block instead of per line :) not a must by any means.

e0ne: done

adrianchiris (approved):
Thx for addressing my comments !

aboch:
LGTM

PR #813, opened 2022-09-26
Implement chains support

e0ne:


chain.go
```diff
@@ -0,0 +1,22 @@
+package netlink
+
+import (
+	"fmt"
+)
+
+// ChainAttrs are the attributes of a Chain
+type Chain struct {
```
adrianchiris, line 8: i wonder if its worth, similar to Filter[1] and Qdisc[2] to abstract this with an interface for consistency.

on the other hand, in contrast to Filter and Qdisc objects we dont have different types of  chains (currently...) so maybe its not worth it. 

just a thought :)

[1] https://github.com/vishvananda/netlink/blob/8e1ce9665a8f8ab9ed94d8830f08781c2951351c/filter.go#L8
[2] https://github.com/vishvananda/netlink/blob/8e1ce9665a8f8ab9ed94d8830f08781c2951351c/qdisc.go#L26

chain.go
```diff
@@ -0,0 +1,22 @@
+package netlink
+
+import (
+	"fmt"
+)
+
+// ChainAttrs are the attributes of a Chain
```
adrianchiris, line 7: nit: ChainAttr -> Chain ?

chain_linux.go
```diff
@@ -0,0 +1,114 @@
+package netlink
+
+import (
+	"fmt"
+	"github.com/vishvananda/netlink/nl"
+	"golang.org/x/sys/unix"
+)
+
```
adrianchiris, line 8: would it make sense to add Unit test for this file ?

chain_linux.go
```diff
@@ -0,0 +1,114 @@
...
+	for _, m := range msgs {
+		msg := nl.DeserializeTcMsg(m)
+
+		attrs, err := nl.ParseRouteAttr(m[msg.Len():])
+		if err != nil {
+			return nil, err
+		}
+
+		// skip chains from other interfaces
+		if link != nil && msg.Ifindex != index {
```
adrianchiris, line 98: not a netlink expert :)  but i see you provide at L76 the ifIndex in the tc msg.
does the kernel returns chains that dont belong to the provided interface index ?

adrianchiris:
@e0ne it would be great to add support for specifying chain in filter (for list/get/modify) as well as part of this PR or a separate one :)

see:
https://github.com/shemminger/iproute2/blob/9f366536edb5158343152604e82b968be46dbf26/tc/tc_filter.c#L189

chain_linux.go
```diff
@@ -0,0 +1,114 @@
...
+	msg := &nl.TcMsg{
+		Family:  nl.FAMILY_ALL,
+		Ifindex: index,
+		Parent:  chain.Parent,
+	}
+	req.AddData(msg)
+	req.AddData(nl.NewRtAttr(nl.TCA_CHAIN, nl.Uint16Attr(chain.Chain)))
+
+	_, err := req.Execute(unix.NETLINK_ROUTE, 0)
+	fmt.Println("6")
```
adrianchiris, line 54: nit: remove this

chain_linux.go
```diff
@@ -0,0 +1,114 @@
...
+		h.ensureIndex(base)
+		index = int32(base.Index)
+	}
+	msg := &nl.TcMsg{
+		Family:  nl.FAMILY_ALL,
+		Ifindex: index,
+		Parent:  chain.Parent,
+	}
+	req.AddData(msg)
+	req.AddData(nl.NewRtAttr(nl.TCA_CHAIN, nl.Uint16Attr(chain.Chain)))
```
adrianchiris, line 51: Chain attribute is 32 bits.

https://github.com/shemminger/iproute2/blob/49c63bc775d610c3dfd3db0d6501ac29c519967f/tc/tc_filter.c#L189

chain.go
adrianchiris: ack, makes sense to me.

chain_test.go
```diff
@@ -0,0 +1,99 @@
...
+		t.Fatal(err)
+	}
+	if len(chains) != 1 {
+		t.Fatal("Failed to add chain")
+	}
+	if chains[0].Chain != *chainVal {
+		t.Fatalf("Incorrect chain added")
+	}
+	classId := MakeHandle(1, 1)
+	filter := &U32{
```
adrianchiris, line 74: testing filter with chain attr should be part of filter_test.go IMO

chain_test.go
```diff
@@ -0,0 +1,99 @@
...
+		QdiscAttrs: QdiscAttrs{
+			LinkIndex: link.Attrs().Index,
+			Handle:    MakeHandle(0xffff, 0),
+			Parent:    HANDLE_INGRESS,
+		},
+	}
+	if err := QdiscAdd(qdisc); err != nil {
+		t.Fatal(err)
+	}
+	qdiscs, err := SafeQdiscList(link)
```
adrianchiris, line 45: i think we can assume `QdiscAdd` works as expected (as its functionality is covered by Qdisc related tests.
so IMO you can remove L45-55 WDYT ?

cmd/ipset-test/main.go
```diff
@@ -123,6 +125,22 @@ func cmdListAll(args []string) {
 	}
 }
 
+func cmdListFilter(args []string) {
```
adrianchiris, line 128: how is this related to ipset-test ?

filter_linux.go
```diff
@@ -446,6 +450,10 @@ func (h *Handle) FilterList(link Link, parent uint32) ([]Filter, error) {
 				default:
 					detailed = true
 				}
+			case nl.TCA_CHAIN:
+				val := new(uint32)
+				*val = binary.LittleEndian.Uint32(attr.Value)
```
adrianchiris, line 455: system may be big-endian (e.g there are powerpc cpus that are big endian)

you can just use `netlink.native.Uint32()` like is being used in other places.

chain_test.go
e0ne: I prefer to keep these lines for consistency and to make TestChainAddDel more explicit in case of failure

chain_test.go
e0ne: done

chain_linux.go
e0ne: tests added

chain_test.go
e0ne: even more: such tests helped me to find an issue with filter chain implementation

chain_linux.go
adrianchiris: i see its done the same in QdiscList[1]
but not done in FilterList[2]

is it really needed? 
if there is no reason to keep this check, lets remove it.

[1] https://github.com/vishvananda/netlink/blob/e0988fd129c1858c092c73fd2c1bfd1a9d2af45d/qdisc_linux.go#L352
[2] https://github.com/vishvananda/netlink/blob/e0988fd129c1858c092c73fd2c1bfd1a9d2af45d/filter_linux.go#L363

chain_test.go
```diff
@@ -0,0 +1,81 @@
...
+		t.Fatal(err)
+	}
+	link, err := LinkByName("foo")
+	if err != nil {
+		t.Fatal(err)
+	}
+	if err := LinkSetUp(link); err != nil {
+		t.Fatal(err)
+	}
+	redir, err := LinkByName("bar")
```
adrianchiris, line 26: why do you need "bar" link ?

chain_test.go
```diff
@@ -0,0 +1,81 @@
...
+	chains, err := ChainList(link, HANDLE_INGRESS)
+	if err != nil {
+		t.Fatal(err)
+	}
+	if len(chains) != 1 {
+		t.Fatal("Failed to add chain")
+	}
+	if chains[0].Chain != *chainVal {
+		t.Fatalf("Incorrect chain added")
+	}
```
adrianchiris, line 70: can you also test for `chains[0].Parent` ?

filter_linux.go
```diff
@@ -235,6 +223,9 @@ func (h *Handle) filterModify(filter Filter, flags int) error {
 		Info:    MakeHandle(base.Priority, nl.Swap16(base.Protocol)),
 	}
 	req.AddData(msg)
+	if filter.Attrs().Chain != nil {
+		req.AddData(nl.NewRtAttr(nl.TCA_CHAIN, nl.Uint32Attr(*filter.Attrs().Chain)))
+	}
 	req.AddData(nl.NewRtAttr(nl.TCA_KIND, nl.ZeroTerminated(filter.Type())))
```
adrianchiris, line 230: in case proto is `unix.RTM_DELTFILTER` i dont think we want to add all the below RTAttrs below

e0ne: In an original implementation in iproute2 there's no specific case for deletion: [1] and [2]


[1] https://github.com/shemminger/iproute2/blob/main/tc/tc_filter.c#L761
[2] https://github.com/shemminger/iproute2/blob/main/tc/tc_filter.c#L67-L86

adrianchiris: ack

adrianchiris (approved):
Thx for addressing my comments !

aboch:
please squash your commits into one

e0ne:
> please squash your commits into one

done. thanks for the feedback, @aboch

PR #816, opened 2022-10-05
- fixes  vishvananda/netlink#815

### LINKED ISSUE #815: xfrm:  v4 over v6 IPsec and v6 over v4 tunnel don't work
currently when creating v4 over v6 tunnel, there are two issues to prevent it from working:

1. the IPv6 tmpl src/dst address of the xfrm policy will be incorrectly treated as IPv4 address 
2. the selector src/dst of the xfrm state is always ::/0, while it should be IPv4

similar things happen with v6 over v4 tunnel

hujun-open:
Fixes  #815 

- changes: 1. set userTmpl.Family to correct family 2. add Selector to XfrmState which is corresponding to XfrmUsersaInfo.Sel

### DIFF
--- a/xfrm_state_linux.go
+++ b/xfrm_state_linux.go
@@ -209,7 +209,6 @@ func (h *Handle) xfrmStateAllocSpi(state *XfrmState) (*XfrmState, error) {
 	msg.Min = 0x100
 	msg.Max = 0xffffffff
 	req.AddData(msg)
-
 	if state.Mark != nil {
 		out := nl.NewRtAttr(nl.XFRMA_MARK, writeMark(state.Mark))
 		req.AddData(out)
@@ -337,7 +336,6 @@ var familyError = fmt.Errorf("family error")
 
 func xfrmStateFromXfrmUsersaInfo(msg *nl.XfrmUsersaInfo) *XfrmState {
 	var state XfrmState
-
 	state.Dst = msg.Id.Daddr.ToIP()
 	state.Src = msg.Saddr.ToIP()
 	state.Proto = Proto(msg.Id.Proto)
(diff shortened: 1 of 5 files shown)


hujun-open:
This fixes issue #815

xfrm_state_linux.go
```diff
@@ -515,6 +515,10 @@ func xfrmUsersaInfoFromXfrmState(state *XfrmState) *nl.XfrmUsersaInfo {
 	msg.Id.Spi = nl.Swap32(uint32(state.Spi))
 	msg.Reqid = uint32(state.Reqid)
 	msg.ReplayWindow = uint8(state.ReplayWindow)
-
+	if state.Selector != nil {
+		selFromPolicy(&msg.Sel, state.Selector)
+	} else {
+		msg.Sel = nl.XfrmSelector{}
+	}
```
aboch, line 522: for readability, can you 
```
msg.Sel = nl.XfrmSelector{}
if state.Selector != nil {
	selFromPolicy(&msg.Sel, state.Selector)
}
```
instead?

aboch:
is it possible to extend existing UT function to cover this?

hujun-open:
> is it possible to extend existing UT function to cover this?

don't know what does "UT" means, could you point to the specific function?

aboch:
> > is it possible to extend existing UT function to cover this?
> 
> don't know what does "UT" means, could you point to the specific function?

Unit Test
I believe you would add it [here](https://github.com/vishvananda/netlink/blob/main/xfrm_state_test.go) or [here](https://github.com/vishvananda/netlink/blob/main/xfrm_policy_test.go) for your change

hujun-open:
ok, updated based on above discussion

aboch:
Thanks @hujun-open 

hujun-open:
squashed 

> Thanks @hujun-open Please squash the two commits into one.

aboch:
LGTM

PR #818, opened 2022-10-09
Add firewall mark option to RouteGetOptions

stv0g:
This option allows performing FIB lookups for a particular firewall mark. It is equivalent to iproute2's 'ip route get mark' option.

### DIFF
--- a/route_test.go
+++ b/route_test.go
@@ -1748,3 +1748,98 @@ func TestRouteUIDOption(t *testing.T) {
 		t.Fatal(routes)
 	}
 }
+
+func TestRouteFWMarkOption(t *testing.T) {
+	tearDown := setUpNetlinkTest(t)
+	defer tearDown()
+
+	// setup eth0 so that network is reachable
+	err := LinkAdd(&Dummy{LinkAttrs{Name: "eth0"}})
+	if err != nil {
+		t.Fatal(err)
+	}
+	link, err := LinkByName("eth0")
+	if err != nil {
+		t.Fatal(err)
+	}
+	if err = LinkSetUp(link); err != nil {
+		t.Fatal(err)
+	}
+	addr := &Addr{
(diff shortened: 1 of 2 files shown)


aboch:
I do not see the changes toparse the Mark when reading the route from kernel.
Also, please add or extend existing UT to test this programming.

stv0g:
> I do not see the changes toparse the Mark when reading the route from kernel.

Can you explain why this is required? This PR will only add a new option to filter routes.

aboch:
> > I do not see the changes toparse the Mark when reading the route from kernel.
> 
> Can you explain why this is required? This PR will only add a new option to filter routes.

I see, this is for the fwmark likely added via policy routing. Thanks.
We do support [`ip rule` programming](https://github.com/vishvananda/netlink/blob/main/rule.go#L9)
Would you be able to add a UT for your change?

stv0g:
Hi @aboch,

> Would you be able to add a UT for your change?

Sure, I will add one this week.

stv0g:
@aboch I added the requested UT

aboch:
> @aboch I added the requested UT

Thank you

route_test.go
```diff
@@ -1748,3 +1748,98 @@ func TestRouteUIDOption(t *testing.T) {
...
+func TestRouteFWMarkOption(t *testing.T) {
+	tearDown := setUpNetlinkTest(t)
+	defer tearDown()
+
+	// setup eth0 so that network is reachable
+	err := LinkAdd(&Dummy{LinkAttrs{Name: "eth0"}})
+	if err != nil {
+		t.Fatal(err)
+	}
+	link, err := LinkByName("eth0")
```
aboch, line 1761: A bit worried about assuming we have `eth0` named interface, but looks like our testing VMs do, UT passed:

image

Issue #817, opened 2022-10-08
Is RouteList incorrectly setting RT_FILTER_OIF


rickpayne:
First time go user, so this could be me - but I was trying to get a list of the routes on my box. Should be simple?
```
package main

import (
    "fmt"
    "github.com/vishvananda/netlink"
)

func main() {
    routes,err := netlink.RouteList(nil, netlink.FAMILY_V4)
    if err != nil {
            fmt.Printf("could not list: %v\n", err)
    }
    for k, v := range routes {
        fmt.Printf("%d: %v\n", k, v)
    }
}
```
I get nothing back. Some debugging shows that the parsed routes are discarded because the OIF is 2, rather than 0 (because my box has a few interfaces and the output interface is not the first one I guess). I think the issue is because we end up here:
```
func (h *Handle) RouteList(link Link, family int) ([]Route, error) {
	routeFilter := &Route{}
	if link != nil {
		routeFilter.LinkIndex = link.Attrs().Index
	}
	return h.RouteListFiltered(family, routeFilter, RT_FILTER_OIF)
}
```
So even if link is nil, an RT_FILTER_OIF is set. I'd have expected a call to RouteList with a nil filter should do just that? What have I missed?

cjmakes:
This used to be the default behavior and was broken in #782 and will be fixed in #825

PR #825, opened 2022-10-26
fix RouteList can't list routes properly when don't specify link

chengzhycn:
Signed-off-by: chengzhycn <chengzhycn@gmail.com>

pr #782 introduced a problem that `RouteList` couldn't show routes properly when we didn't specify a link. 

It changed `routeFilter` from a nil pointer to an empty `Route{}` by default, so in `RouteListFiltered` it couldn't skip the filter logic when we didn't specify a link and we could only get an empty route list returned by it.

cc @cjmakes

### DIFF
--- a/route_linux.go
+++ b/route_linux.go
@@ -1004,8 +1004,10 @@ func (h *Handle) RouteList(link Link, family int) ([]Route, error) {
 	routeFilter := &Route{}
 	if link != nil {
 		routeFilter.LinkIndex = link.Attrs().Index
+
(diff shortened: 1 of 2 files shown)


chengzhycn:
@vishvananda @aboch could you help me review this pr? :)

cjmakes (approved):
Thanks for fixing this, sorry for the breaking change.

aboch:
LGTM

aboch:
Fixes #817

PR #827, opened 2022-11-15
Fix bug in FOU netlink parser

martin31821:
Netlink API states that the TLV tuples should be padded to 4 bytes, which was implemented incorrectly in the FOU case.
On recent kernels, ListFou fails because the kernel returns more attributes than the parser expects.


aboch:
@martin31821  pls squash your commits into one

aboch:
@martin31821 please squash your commits into one

aboch:
@martin31821

aboch:
@martin31821 please fix the conflicts

PR #835, opened 2023-01-09
Allow netlink with no owner/group

mmirecki:
Currently it is only possible to create a tuntap with owner/group specified. If not set explicitly, the default value of 0 will be used. It's therefore not possible to create a tuntap without owner/group. This PR changes this to allow this. It changes the type of owner/group from uint to int to allow setting owner/group to -1, indicating no owner/group.

### DIFF
--- a/link.go
+++ b/link.go
@@ -339,8 +339,8 @@ type Tuntap struct {
 	NonPersist bool
 	Queues     int
 	Fds        []*os.File
-	Owner      uint32
-	Group      uint32
+	Owner      int64 // -1 for no owner
+	Group      int64 // -1 for no group
 }
 
 func (tuntap *Tuntap) Attrs() *LinkAttrs {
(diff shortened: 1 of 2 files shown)


mmirecki:
@vishvananda @aboch Could you please take a look at this? Thanks

dcbw:
looks good to me, though it would be a downstream code change if for some reason consumers relied on the uint64 type

mmirecki:
> looks good to me, though it would be a downstream code change if for some reason consumers relied on the uint64 type

Unfortunately we need to change the api one way or another to pass information about "no value". 
The other options are:
- use *uint with nil for no owner/group - but we don't use pointers in the api so it could be confusing
- add new attributes hasOwner/hasGroup, this would be backwards compatible, but would add redundant parameters

dcbw:
Yeah, probably OK since it's a fairly minor change and adds a useful benefit.

LGTM

aboch:
> [...] it would be a downstream code change if for some reason consumers relied on the uint64 type

Yes, that is the problem

mmirecki:
> > [...] it would be a downstream code change if for some reason consumers relied on the uint64 type
> 
> Yes, that is the problem

Fixing the problem will need to break the backwards compatibility in some way (except if we add new parameters to indicated the condition, which would probably be even uglier)
Any suggestions on which way of breaking it would be the least evil?

PR #847, opened 2023-02-26
supports AF_XDP socket diagnosis

thediveo:
- dumps the diagnosis data of all sockets of type AF_XDP in the current network namespace.
- dumps the diagnosis data of only a single socket that matches a specific inode number an/or socket cookie. Now, while the XDP socket diagnosis request allows for specifying a inode number and socket cookie this information doesn't seem to be used in the Linux kernel xsk_diag.c implementation anywhere and tests show that no filtering takes place.

Supplying tests is rather complex in the case of AF_XDP sockets and while I have tests, these pull in lots of other code because of the long-winded way these sockets need to be properly set up and rely on the Ginkgo/Gomega TDD framework in order to reduce lots of boilerplate, so I'm refraining from polluting the netlink module.

### DIFF
--- a/socket.go
+++ b/socket.go
@@ -35,3 +35,70 @@ type UnixSocket struct {
 	INode  uint32
 	Cookie [2]uint32
 }
+
+// XDPSocket represents an XDP socket (and the common diagnosis part in
+// particular). Please note that in contrast to [UnixSocket] the XDPSocket type
+// does not feature “State” information.
+type XDPSocket struct {
+	// xdp_diag_msg
+	// https://elixir.bootlin.com/linux/v6.2/source/include/uapi/linux/xdp_diag.h#L21
+	Family uint8
+	Type   uint8
+	pad    uint16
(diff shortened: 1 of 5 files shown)


aboch:
@thediveo please resolve the conflicts

thediveo:
@aboch any idea why ...

```
=== RUN   TestSocketXDPGetInfo
    socket_xdp_linux_test.go:30: no such file or directory
--- FAIL: TestSocketXDPGetInfo (0.00s)
```

...fails on the Github runner, but succeeds on both a Ubuntu 24.04 (kernel 6.8) and 22.04LTS (kernel 5.15.0)?

thediveo:
@aboch skipping XDP socket diagnosis test if kernel reports ENOENT, thus lacking XSK diagnosis.

aboch:
please squash the two commits into one

thediveo:
any issues, or can this be please merged?

PR #850, opened 2023-03-06
Add field for permanent hardware address

lorenz:
Linux 5.6 and higher support IFLA_PERM_ADDRESS, which contains the permanent hardware address of the interface if an interface has such an address. This can be used to identify interfaces even when the normal hardware address has been changed.

### DIFF
--- a/link.go
+++ b/link.go
@@ -54,6 +54,7 @@ type LinkAttrs struct {
 	GROIPv4MaxSize uint32
 	Vfs            []VfInfo // virtual functions available on link
 	Group          uint32
+	PermHWAddr     net.HardwareAddr
 	Slave          LinkSlave
 }
 
--- a/link_linux.go
+++ b/link_linux.go
@@ -2100,6 +2100,13 @@ func LinkDeserialize(hdr *unix.NlMsghdr, m []byte) (Link, error) {
 			base.NumRxQueues = int(native.Uint32(attr.Value[0:4]))
 		case unix.IFLA_GROUP:
 			base.Group = native.Uint32(attr.Value[0:4])
+		case unix.IFLA_PERM_ADDRESS:
+			for _, b := range attr.Value {
+				if b != 0 {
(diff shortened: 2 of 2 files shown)


link.go
```diff
@@ -50,7 +50,9 @@ type LinkAttrs struct {
 	GROMaxSize   uint32
 	Vfs          []VfInfo // virtual functions available on link
 	Group        uint32
-	Slave        LinkSlave
+
```
aboch, line 53: I assume you added this empty line to avoid the reformat of all existing fields?
Would name the new field `PermHWAddr` good enough?

lorenz:
Rebased and applied naming suggestion

link_linux.go
```diff
@@ -2100,6 +2100,16 @@ func LinkDeserialize(hdr *unix.NlMsghdr, m []byte) (Link, error) {
 			base.NumRxQueues = int(native.Uint32(attr.Value[0:4]))
 		case unix.IFLA_GROUP:
 			base.Group = native.Uint32(attr.Value[0:4])
+		case unix.IFLA_PERM_ADDRESS:
+			var nonzero bool
+			for _, b := range attr.Value {
+				if b != 0 {
+					nonzero = true
```
aboch, line 2107: Could we just
```
base.PermHWAddr = attr.Value[:]
break
```
and get rid of nonzero variable?

link_linux.go
```diff
@@ -2100,6 +2100,16 @@ func LinkDeserialize(hdr *unix.NlMsghdr, m []byte) (Link, error) {
 			base.NumRxQueues = int(native.Uint32(attr.Value[0:4]))
 		case unix.IFLA_GROUP:
 			base.Group = native.Uint32(attr.Value[0:4])
+		case unix.IFLA_PERM_ADDRESS:
+			var nonzero bool
```
aboch, line 2104: pls remove this variable

aboch:
LGTM

Issue #670, opened 2021-07-26
param check bug


hititou:
func (h *Handle) routeHandle(route *Route, req *nl.NetlinkRequest, msg *nl.RtMsg) error {
	if (route.Dst == nil || route.Dst.IP == nil) && route.Src == nil && route.Gw == nil && route.MPLSDst == nil {
		return fmt.Errorf("one of Dst.IP, Src, or Gw must not be nil")
	}

delete this param check ??
default route like this can not add or del by this func, but ‘ip route add/del default dev eth0’ command can

[root@VM-0-7-centos ~]# ip route list table 100
default dev eth1 scope link 
10.55.45.0/24 dev br-c163e6327285 proto kernel scope link src 10.55.45.1 
192.168.1.0/24 dev eth1 proto kernel scope link src 192.168.1.4

rajal-amzn:
Facing the same issue. Is there a workaround for this?

vishvananda:
Have you tried setting dst to 0.0.0.0/0?

rajal-amzn:
Yes Here is my rough code
```
link, _ := netlink.LinkByName("lo")
netlink.LinkSetUp(link)
_, dst, _ := net.ParseCIDR("0.0.0.0/0")
routeToBeAdded := &netlink.Route{
        Dst:      dst,
        Scope:     unix.RT_SCOPE_HOST,
	Type:      unix.RTN_LOCAL,
	Table:     100,
	LinkIndex: link.Attrs().Index,
}
netlink.RouteAdd(routeToBeAdded)
handle, _ := netlink.NewHandle()
route := &netlink.Route{
	Table:     100,
	LinkIndex: link.Attrs().Index,
}
routes, _ := handle.RouteListFiltered(unix.AF_INET, route, netlink.RT_FILTER_TABLE|netlink.RT_FILTER_OIF) // This returns 1 route with Dst set to nil
for _, r := range routes {
	routeToBeDeleted := r
	netlink.RouteDel(&routeToBeDeleted) // This returns error
}
```

Issue #611, opened 2021-01-19
"blackhole default" route cannot be deleted via netlink package


matt-johnston-ecommistry:
A rule that shows up in "ip route" as just "blackhole default" cannot be removed via this package when retrieved from netlink.RouteList or netlink.RouteListFiltered:

`routes, err := netlink.RouteListFiltered(netlink.FAMILY_V4, &netlink.Route{
		Table: 101,
	}, netlink.RT_FILTER_TABLE)
	if err != nil {
		return fmt.Errorf("failed retrieving routes for table %d", v.VnetId)
	}

	for _, r := range routes {
		err = netlink.RouteDel(&r)
		if err != nil {
			return fmt.Errorf("failed deleting route '%s' for %d: %w", r.String(), 101, err)
		}
	}`
	
output:
`failed deleting route '{Ifindex: 0 Dst: <nil> Src: <nil> Gw: <nil> Flags: [] Table: 101}' for 101: one of Dst.IP, Src, or Gw must not be nil`
	
ip route output:
`ip route show table 101
blackhole default 
172.31.0.0/24 dev bond0.101 src 172.31.0.1 `

Temporary workaround - for our code since we know that only these two rule types have this scenario.
`if (r.Type == unix.RTN_PROHIBIT || r.Type == unix.RTN_BLACKHOLE) && (r.Dst == nil || r.Dst.IP == nil) && r.Src == nil && r.Gw == nil && r.MPLSDst == nil {
			r.Dst = broadcast // broadcast is 0.0.0.0/0 constant
		}`

PR #852, opened 2023-03-08
Add "default" dst while no DST attr responsed from nl

PlatformLC:
For #670 && #611 are both introduced by "default" route. 

There is indeed no attribute tlv encoded for RTA_DST while doing RTM_GETROUTE requests, but while we doing ADD/DELETE opearation, we do need  encode such "default" dst (i.e., 0.0.0.0/0 or ::/0).
This introduced the inconsistence between "read" and "write" and lead failure for flushing the route table(deleting all routes from what listed)

Refer to [iproute2](https://github.com/shemminger/iproute2/blob/main/ip/iproute.c#L820) implementation, add the same logic for `deserializeRoute`

### DIFF
--- a/route_linux.go
+++ b/route_linux.go
@@ -1068,6 +1068,9 @@ func (h *Handle) RouteListFiltered(family int, filter *Route, filterMask uint64)
 				continue
 			case filterMask&RT_FILTER_DST != 0:
 				if filter.MPLSDst == nil || route.MPLSDst == nil || (*filter.MPLSDst) != (*route.MPLSDst) {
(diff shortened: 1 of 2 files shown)


PlatformLC:
@vishvananda Would you have a look at this?

vishvananda:
this looks good, if you squash as @aboch requested we can merge it.

PlatformLC:
> @PlatformLC Please squash your two commits into one. This project follows a one-commit-per-PR scheme

Updated. PTAL

PR #855, opened 2023-03-10
rule: add support for FRA_PROTOCOL

NikAleksandrov:
Add support for ip rules' FRA_PROTOCOL attribute and also check for it when testing rules. The default ip rule protocol is RTPROT_UNSPEC (0) so we set the attribute only when it is >0.


tklauser (approved):

tklauser:
@vishvananda @aboch PTAL

PR #861, opened 2023-04-03
Support ipset test entry existence

major1201:
This PR supports ipset test whether an entry is in a set or not.

Func added:

- IpsetTest(setname string, entry *IPSetEntry) (bool, error)


aboch:
LGTM

PR #862, opened 2023-04-03
Add support for alternative names

SchSeba:
This commit add the support for alternative names.

exposing the attributes in the link object.
also adding the add and delete functions

### DIFF
--- a/link.go
+++ b/link.go
@@ -33,6 +33,7 @@ type LinkAttrs struct {
 	MasterIndex    int         // must be the index of a bridge
 	Namespace      interface{} // nil | NsPid | NsFd
 	Alias          string
+	AltNames       []string
 	Statistics     *LinkStatistics
 	Promisc        int
 	Allmulti       int
--- a/link_linux.go
+++ b/link_linux.go
@@ -497,6 +497,58 @@ func (h *Handle) LinkSetAlias(link Link, name string) error {
 	return err
 }
 
+// LinkAddAltName adds a new alternative name for the link device.
+// Equivalent to: `ip link property add $link altname $name`
+func LinkAddAltName(link Link, name string) error {
+	return pkgHandle.LinkAddAltName(link, name)
+}
+
+// LinkAddAltName adds a new alternative name for the link device.
+// Equivalent to: `ip link property add $link altname $name`
+func (h *Handle) LinkAddAltName(link Link, name string) error {
+	base := link.Attrs()
+	h.ensureIndex(base)
+	req := h.newNetlinkRequest(unix.RTM_NEWLINKPROP, unix.NLM_F_ACK)
+
+	msg := nl.NewIfInfomsg(unix.AF_UNSPEC)
+	msg.Index = int32(base.Index)
+	req.AddData(msg)
+
+	data := nl.NewRtAttr(unix.IFLA_PROP_LIST|unix.NLA_F_NESTED, nil)
+	data.AddRtAttr(unix.IFLA_ALT_IFNAME, []byte(name))
+
+	req.AddData(data)
+
+	_, err := req.Execute(unix.NETLINK_ROUTE, 0)
+	return err
+}
+
+// LinkDelAltName delete an alternative name for the link device.
+// Equivalent to: `ip link property del $link altname $name`
+func LinkDelAltName(link Link, name string) error {
+	return pkgHandle.LinkDelAltName(link, name)
+}
+
+// LinkDelAltName delete an alternative name for the link device.
+// Equivalent to: `ip link property del $link altname $name`
+func (h *Handle) LinkDelAltName(link Link, name string) error {
+	base := link.Attrs()
+	h.ensureIndex(base)
+	req := h.newNetlinkRequest(unix.RTM_DELLINKPROP, unix.NLM_F_ACK)
+
+	msg := nl.NewIfInfomsg(unix.AF_UNSPEC)
+	msg.Index = int32(base.Index)
+	req.AddData(msg)
+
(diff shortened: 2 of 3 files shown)


aboch:
Thank you.
Can you pelase add some UT as well?

hellt:
I would love to see this merged. Need it to set user-friendly names for otherwise restricted link length

hellt:
@SchSeba I wonder, which func would one use to get the interface by its altname?

SchSeba:
Hi @hellt,
it will be just as part of the link attrs you will have a list of them https://github.com/vishvananda/netlink/pull/862/files#diff-ff692b0e80c3a46490bdaf004f6628c003a4736c549aa8050a430246e70d6703R36

hellt:
Hi @SchSeba 
So to get a link by an alias one would need to list all links and then do a loop to the desired altname to be present in []AltNames?

I wonder, since we have LinkByAlias, wouldn't it be nice to have LinkByAltName?

SchSeba:
The problem is that `LinkByAltName` will have to do "list all links and then do a loop to the desired altname to be present in []AltNames" because there is no API right now in the kernel.

btw I didn't try that but I think you can just use the LinkByName with the Alt name and it should work can you give it a try?

SchSeba:
@aboch @vishvananda could you take a look at this PR ? :)

SchSeba:
Hi @hellt be aware that now it works link netlink does you can do LinkByName and it will find the device base on the name or the altName

```
ip link show enp8s0u2u1u2
26: ens1u2u1u2: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc fq_codel state UP mode DEFAULT group default qlen 1000
    link/ether 08:3a:88:61:c6:bf brd ff:ff:ff:ff:ff:ff
    altname enp8s0u2u1u2
ip link show enp8s0u2u1u2
26: ens1u2u1u2: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc fq_codel state UP mode DEFAULT group default qlen 1000
    link/ether 08:3a:88:61:c6:bf brd ff:ff:ff:ff:ff:ff
    altname enp8s0u2u1u2
```

link.go
```diff
@@ -33,6 +33,7 @@ type LinkAttrs struct {
 	MasterIndex    int         // must be the index of a bridge
 	Namespace      interface{} // nil | NsPid | NsFd
 	Alias          string
+	AltNames     []string
```
adrianchiris, line 36: nit: align indent

link_linux.go
```diff
@@ -497,6 +497,58 @@ func (h *Handle) LinkSetAlias(link Link, name string) error {
...
+	return err
+}
+
+// LinkDelAltName delete a new alternative name for the link device.
+// Equivalent to: `ip link property del $link altname $name`
+func LinkDelAltName(link Link, name string) error {
+	return pkgHandle.LinkDelAltName(link, name)
+}
+
+// LinkDelAltName delete a new alternative name for the link device.
```
adrianchiris, line 532: same as above

link_linux.go
```diff
@@ -996,28 +1048,28 @@ func LinkSetXdpFdWithFlags(link Link, fd, flags int) error {
 // LinkSetGSOMaxSegs sets the GSO maximum segment count of the link device.
 // Equivalent to: `ip link set $link gso_max_segs $maxSegs`
 func LinkSetGSOMaxSegs(link Link, maxSegs int) error {
-       return pkgHandle.LinkSetGSOMaxSegs(link, maxSegs)
+	return pkgHandle.LinkSetGSOMaxSegs(link, maxSegs)
```
adrianchiris, line 1051: i believe you replaced spaces with tabs here ?

link_linux.go
```diff
@@ -2136,6 +2195,18 @@ func LinkDeserialize(hdr *unix.NlMsghdr, m []byte) (Link, error) {
...
+		case unix.IFLA_PROP_LIST | unix.NLA_F_NESTED:
+			attrs, err := nl.ParseRouteAttr(attr.Value[:])
+			if err != nil {
+				return nil, err
+			}
+
+			base.AltNames = []string{}
+			for _, attr := range attrs {
+				if attr.Attr.Type == unix.IFLA_ALT_IFNAME {
+					base.AltNames = append(base.AltNames, string(attr.Value[:len(attr.Value)-1]))
```
adrianchiris, line 2207: you can use: https://github.com/vishvananda/netlink/blob/857968af1192b419c958984e96fa5154ada6baff/nl/nl_linux.go#L869

which takes care of the `0` char

but i see most of the code here does as you do.

link_test.go
```diff
@@ -1883,6 +1883,53 @@ func TestLinkSet(t *testing.T) {
 	if link.Attrs().Group != 42 {
 		t.Fatal("Link group not changed")
 	}
+
+	err = LinkAddAltName(link, "altname")
```
adrianchiris, line 1887: any chance to move this to a separate test ?

feels like this tests is already testing quite alot.

link_linux.go
SchSeba: the IDE did that I didn't change the code here

SchSeba:
@adrianchiris please take another look when you have time

link_linux.go
adrianchiris: ack, seems rest of file is tab spaced so your IDED just aligned.

adrianchiris:
@SchSeba code changes look OK. thx for addressing my comments.

could you please update commit message to mention the changes in `LinkByName()`

SchSeba:
Done thanks @adrianchiris !

adrianchiris (approved):
lgtm

PR #809, opened 2022-09-03
Support src_mac, dst_mac and vlan id in flower filters

shuowang-123:
See title, so users are able to filter flower rules based on src/dst mac address. I ran the test on an Oracle Linux 8 virtual machine and it passed:

[] go test -run "TestFilterFlowerAddDel"
PASS
ok  	github.com/vishvananda/netlink	0.017s

I also ran the test (without cleanup) on the same machine, was able to see the flower rule was created correctly, as reported by "tc":

[] ip netns exec netns1 tc filter show dev foo ingress
filter parent ffff: protocol all pref 1 flower chain 0
filter parent ffff: protocol all pref 1 flower chain 0 handle 0x1
  dst_mac 2c:54:91:88:c9:e5
  src_mac 2c:54:91:88:c9:e3
  eth_type ipv4
  dst_ip 1.0.0.1/24
  src_ip 2.0.0.1/24
  enc_dst_ip 3.0.0.1/24
  enc_src_ip 4.0.0.1/24
  enc_key_id 1234
  enc_dst_port 8472
  not_in_hw
	action order 1: mirred (Egress Redirect to device bar) stolen
	index 1 ref 1 bind 1

PR #872, opened 2023-05-24
flower: add support for mac address matching

rjarry:
Allow matching ethernet addresses in tc flower filters.


rjarry:
My bad, this is duplicate of #809

PR #873, opened 2023-05-24
Bridge port isolate support

rjarry:
Allow isolating bridge ports with `LinkSetIsolated`. This is the equivalent of:

```
bridge link set dev <foo> isolated on|off
```


aboch:
Can you please add a UT for this, usually program to and read it back from kernel.
Thanks

aboch:
@rjarry

rjarry:
Hi @aboch sorry for the delay. I had forgotten about this.

Let me know if I need to change anything.

PR #879, opened 2023-06-13
link: add LinkSetGSOIPv4MaxSize and LinkSetGROIPv4MaxSize

borkmann:
Add two new methods to allow setting GSO and GRO max size attributes only. They make it much easier to enable IPv4 BIG TCP [0].

The equivalent iproute2 commands are:

$ ip link set $link gso_ipv4_max_size $maxSize
$ ip link set $link gro_ipv4_max_size $maxSize

Also add tests for them. We already do support the IPv6 counterpart via 543bb1cadec5 ("link: add LinkSetGSOMaxSize and LinkSetGROMaxSize").

  [0] https://lore.kernel.org/netdev/cover.1674921359.git.lucien.xin@gmail.com/

### DIFF
--- a/link_linux.go
+++ b/link_linux.go
@@ -956,13 +956,13 @@ func LinkSetXdpFdWithFlags(link Link, fd, flags int) error {
 	return err
 }
 
-// LinkSetGSOMaxSize sets the GSO maximum size of the link device.
+// LinkSetGSOMaxSize sets the IPv6 GSO maximum size of the link device.
 // Equivalent to: `ip link set $link gso_max_size $maxSize`
 func LinkSetGSOMaxSize(link Link, maxSize int) error {
 	return pkgHandle.LinkSetGSOMaxSize(link, maxSize)
 }
 
-// LinkSetGSOMaxSize sets the GSO maximum size of the link device.
+// LinkSetGSOMaxSize sets the IPv6 GSO maximum size of the link device.
 // Equivalent to: `ip link set $link gso_max_size $maxSize`
 func (h *Handle) LinkSetGSOMaxSize(link Link, maxSize int) error {
 	base := link.Attrs()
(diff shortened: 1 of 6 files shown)


borkmann:
Hi :wave: @aboch, could you pls take a look? Thanks so much!

link_linux.go
```diff
@@ -956,13 +956,13 @@ func LinkSetXdpFdWithFlags(link Link, fd, flags int) error {
 	return err
 }
 
-// LinkSetGSOMaxSize sets the GSO maximum size of the link device.
+// LinkSetGSOMaxSize sets the IPv6 GSO maximum size of the link device.
```
aboch, line 959: Is this true for the iproute2 command as well, that it is for ipv6 only?

Also, I do not find the gso_ipv4_max_size option in iproute2 shipped in RHEL 9. 

Is this a very recent addition?

borkmann: BIG TCP was initially targeted only for IPv6-only: https://lore.kernel.org/netdev/20220513183408.686447-3-eric.dumazet@gmail.com/T/  . This went into 5.19 kernel.

Only later IPv4 BIG TCP got added in 6.3 kernel: https://lore.kernel.org/netdev/cover.1674921359.git.lucien.xin@gmail.com/

I doubt this is anywhere in RHEL at this point given it's fairly recent.

borkmann: Forgot to add the iproute2 link: https://git.kernel.org/pub/scm/network/iproute2/iproute2.git/commit/?id=1dafe448c7a2f2be5dfddd8da250980708a48c41

@aboch let me know if you need any other info.

aboch: Thank you @borkmann that's the context I needed.

link_test.go
```diff
@@ -2148,6 +2148,66 @@ func TestLinkSetGROMaxSize(t *testing.T) {
...
+	if err := LinkAdd(iface); err != nil {
+		t.Fatal(err)
+	}
+
+	link, err := LinkByName("foo")
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	err = LinkSetGSOIPv4MaxSize(link, 32768)
```
aboch, line 2166: Many times people would learn a functionality by looking at its UT.
Is there any reason for choosing an exact power of 2?
This may mislead people; one would expect 2^N -1, as in networking and computing max sizes are usually because of a N bytes long protocol field/buffer.

borkmann: Power of 2 is okay here. I used the same value as in the existing IPv6 test further above, and the kernel default is also power of 2.

PR #881, opened 2023-06-19
Use IFLA_G{S,R}O_IPV4_MAX_SIZE constans from golang.org/x/sys/unix

tklauser:
Commit 2b008399a406 ("link: add LinkSetGSOIPv4MaxSize and LinkSetGROIPv4MaxSize") introduced the use of these constants, but they were added only recently to the golang.org/x/sys/unix package. Update its version in go.mod and use the pre-defined constants instead of duplicating them.


borkmann:
Awesome, looks good! Thanks Tobias!

borkmann:
Cc @aboch

borkmann (approved):

PR #882, opened 2023-06-23
Seg6local: add support for SEG6_LOCAL_ACTION_END_BPF

shu1r0:
This commit add the support for End.BPF. 

With this commit, the eBPF program (BPF_PROG_TYPE_LWT_SEG6LOCAL) is loaded by a third-party library, and the eBPF can be configured by setting FD and name.

For reference, the iproute2 command to set End.BPF is as follows: :
```
 ip -6 route add 2001:db8::6 encap seg6local action End.BPF endpoint object ex-obj.o section ex-prog dev eth0
```

When I try this commit in my environment, the following route is set: (eBPF section name is `lwt_seg6local/example`)
```
2001:db8:20::2  encap seg6local action End.BPF endpoint lwt_seg6local/example dev r1_h2 metric 1024 pref medium
```

### DIFF
--- a/nl/seg6local_linux.go
+++ b/nl/seg6local_linux.go
@@ -12,6 +12,7 @@ const (
 	SEG6_LOCAL_NH6
 	SEG6_LOCAL_IIF
 	SEG6_LOCAL_OIF
+	SEG6_LOCAL_BPF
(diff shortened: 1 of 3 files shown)


aboch:
LGTM

PR #889, opened 2023-07-10
Linux rule list results have Family populated

### LINKED ISSUE #708: [ip-rule] return rules with correct family
Hello,

I'm working on some code which manipulates the RPDB for both IPv6 and IPv4. 

To do this I perform a perform a list of rules like so:

```
	move := func(family int) error {
		rules, err := netlink.RuleList(family)
		if err != nil {
			return err
		}
```

Later on I pass the obtained rules to 'RuleAdd' and 'RuleDel' commands. 

To my surprise, the returned rule objects have a "Family" field set to zero, despite the filtering call specifying the family. 

This causes issues as I'm trying to write the same rule in separate RPDB's for ipv6 and ipv4 ( equivalent to adding the same policy rule prefixed with 'ip -4' and 'ip -6' - this will not conflict). 

I have to perform the following for the two calls of "move" to work correctly.

```
		rulesByPrio := map[int]*netlink.Rule{}
		for i := range rules {
			rulesByPrio[rules[i].Priority] = &rules[i]
			// oddly enough, the netlink library
			// does not return rules with the
			// family defined.
			// even despite filtering for the family
			// in the above RuleList call.
			rules[i].Family = family
		}
```

Would it be more intuitive for the returned rules to have the correct Family fields realized? Not requiring the caller to do this.

d2r:
Fixes #708

### DIFF
--- a/rule_linux.go
+++ b/rule_linux.go
@@ -223,6 +223,7 @@ func (h *Handle) RuleListFiltered(family int, filter *Rule, filterMask uint64) (
 		rule := NewRule()
 
 		rule.Invert = msg.Flags&FibRuleInvert > 0
+		rule.Family = int(msg.Family)
 		rule.Tos = uint(msg.Tos)
 
 		for j := range attrs {
--- a/rule_test.go
+++ b/rule_test.go
@@ -23,6 +23,7 @@ func TestRuleAddDel(t *testing.T) {
 	}
 
 	rule := NewRule()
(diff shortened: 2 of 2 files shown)


josh-chan:
@vishvananda Could you pls merge this PR? I need this!

vishvananda:
Thanks for including tests.
Lgtm

PR #895, opened 2023-07-20
Support tc flower action statistic

SirHao:
Enable tc filter statistics for flower action, equivalent to: `tc -s filter show`, including:
1. Classic statistics like `Sent 1024 bytes 1 pkt (dropped 0, overlimits 0 requeues 0)`
2. Hardward statistics like `Sent hardware 1024 bytes 1 pkt`
3. Time statistics like `ref 1 bind 1 installed 10 sec used 10 sec`

### DIFF
--- a/filter.go
+++ b/filter.go
@@ -118,16 +118,31 @@ func (a TcPolAct) String() string {
 }
 
 type ActionAttrs struct {
-	Index   int
-	Capab   int
-	Action  TcAct
-	Refcnt  int
-	Bindcnt int
+	Index      int
+	Capab      int
+	Action     TcAct
+	Refcnt     int
+	Bindcnt    int
+	Statistics *ActionStatistic
+	Timestamp  *ActionTimestamp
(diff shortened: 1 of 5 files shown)


filter.go
```diff
@@ -243,6 +256,9 @@ type MirredAction struct {
 	ActionAttrs
 	MirredAction MirredAct
 	Ifindex      int
+
+	Statistics *ActionStatistic
+	Timestamp  *ActionTimestamp
```
dakehero, line 261: Have you ever considered putting it into `ActionAttrs`?

dakehero (changes requested):
I'm not familiar with Actions

filter.go
SirHao: U are right, these statistics are indeed part of the `ActionAttrs`, which already included fields like `Index`, `Refcnt`, `Bindcnt`, etc. 
I have modified this point~

PR #896, opened 2023-07-24
add tc flower pedit action

chent1996:
this pr add support for pedit action in tc flower, it is useful to install hardware offload rules with netlink pedit action.

authored-by: tangchen <tangchen.1@bytedance.com>


chent1996:
WIP: refracting the code to fix dependency problem

PR #898, opened 2023-07-26
Fix FlowBased support for Geneve

rgo3:
The IFLA_GENEVE_COLLECT_METADATA netlink attribute shouldn't have any a payload. For Geneve devices also other attributes can be set next to FlowBased, however the VNI needs to be 0.

This commit also adds a test for creating a Geneve device in FlowBased mode.


rgo3:
Hello @aboch 
Any chance I can get a review? We are moving away from using iproute2 in cilium to create various devices and this PR is needed to move forward with cilium/cilium#27082. Thanks!

aboch:
LGTM

PR #904, opened 2023-08-14
change all ioutil calls to os

DAT4:
This pull request removes all calls and imports of ioutils which is deprecated.

Only ReadFile and WriteFile of ioutil is used and those are just imported from os instead.

_This is my first pull request on this branch, I don't know if you have some rules about how it should be done, please let me know if I need to do anything :-)_

### DIFF
--- a/ipset_linux_test.go
+++ b/ipset_linux_test.go
@@ -2,16 +2,17 @@ package netlink
 
 import (
 	"bytes"
-	"io/ioutil"
 	"net"
+	"os"
 	"testing"
 
-	"github.com/vishvananda/netlink/nl"
 	"golang.org/x/sys/unix"
+
+	"github.com/vishvananda/netlink/nl"
 )
 
 func TestParseIpsetProtocolResult(t *testing.T) {
-	msgBytes, err := ioutil.ReadFile("testdata/ipset_protocol_result")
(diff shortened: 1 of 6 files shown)


DAT4:
Oh I just realized the autoformatting kicked in and added linebreaks on longer function calls. Let me know if this is a problem, then I will go back and change it.

ipset_linux_test.go
```diff
@@ -76,14 +77,22 @@ func TestParseIpsetListResult(t *testing.T) {
 	}
 	expectedMAC := net.HardwareAddr{0xde, 0xad, 0x0, 0x0, 0xbe, 0xef}
 	if !bytes.Equal(ent.MAC, expectedMAC) {
-		t.Errorf("expected MAC for first entry to be %s, got %s", expectedMAC.String(), ent.MAC.String())
+		t.Errorf(
+			"expected MAC for first entry to be %s, got %s",
+			expectedMAC.String(),
+			ent.MAC.String(),
+		)
```
thaJeztah, line 84: I'm not a maintainer on this repo, but I'd prefer the changes to be reverted; keeping these lines as a single line makes it easier to "skimp" over them when reading the code ("we're returning an error"); breaking it up over multiple lines doesn't _really_ improve readability, and introduces code-churn, which we should avoid.

PR #921, opened 2023-10-24
Fix TestRouteFilterAllTables and TestRouteExtraFields failures

tklauser:
Commit 1:

    Enable extended error message reporting in tests
    
    This causes the NETLINK_EXT_ACK socket option to be set and extended
    error messages to be reported in errors. This should help debug test
    failures.

Commit 2:

    nl: avoid trailing NULL byte in error messages
    
    Use unix.ByteSliceToString to convert the NULL-terminated
    NLMSGERR_ATTR_MSG error message.

Commit 3:

    Use valid Tos value in TestRouteFilterAllTables and TestRouteExtraFields
    
    TestRouteFilterAllTables and TestRouteExtraFields started failing a
    while ago after GitHub actions images bumped the kernel version from
    5.15.x to 6.2.x [1].
    
    This is because newer kernels containing commit [2] started rejecting
    Tos values with the ECN bits set to non-zero, this leads to the
    following error being reported when adding routes:
    
        invalid argument: Invalid dsfield (tos): ECN bits must be 0
    
    Fix this by using a valid Tos value in TestRouteFilterAllTables and
    TestRouteExtraFields.
    
    [1] https://github.com/vishvananda/netlink/pull/773#issuecomment-1754436653
    [2] https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=f55fbb6afb8d


tklauser:
@vishvananda @aboch PTAL. This should fix the CI failures currently seen on the `main` branch after GitHub actions runner images updated kernel versions, also see https://github.com/vishvananda/netlink/pull/773#issuecomment-1754436653

aboch:
Thanks @tklauser for your fixes!

Usually we want a commit per PR, but in this case the commits are clearly fixing/improving independent things. Since all changes pertain to UT fix, I will not ask for four independent PRs.

LGTM

tklauser:
> Usually we want a commit per PR, but in this case the commits are clearly fixing/improving independent things. Since all changes pertain to UT fix, I will not ask for four independent PRs.

Thanks, I wasn't aware of that rule. I'll keep it in mind for future changes.

PR #924, opened 2023-11-12
fix: check for ext ACK TLVs

adrianchiris:
checking for Extended ACK in nlmsghdr flags
should be done with bitwise AND op instead of OR op

### DIFF
--- a/nl/nl_linux.go
+++ b/nl/nl_linux.go
@@ -569,7 +569,7 @@ done:
 				err = syscall.Errno(-errno)
 
 				unreadData := m.Data[4:]
-				if m.Header.Flags|unix.NLM_F_ACK_TLVS != 0 && len(unreadData) > syscall.SizeofNlMsghdr {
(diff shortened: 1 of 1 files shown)


nl/nl_linux.go
```diff
@@ -661,12 +661,14 @@ func GetNetlinkSocketAt(newNs, curNs netns.NsHandle, protocol int) (*NetlinkSock
...
 // at the end of the code that needs to be executed in the network namespace.
 // Example:
-// func jobAt(...) error {
-//      d, err := executeInNetns(...)
-//      if err != nil { return err}
-//      defer d()
-//      < code which needs to be executed in specific netns>
-//  }
+//
+//	func jobAt(...) error {
```
adrianchiris, line 665: this change is unrelated to the fix but the linter keeps adjusting the style.

aboch:
Thanks for catching this one

aboch:
LGTM

aboch:
LGTM

PR #925, opened 2023-11-12
fix: check nlmsghdr flags for interrupt

adrianchiris:
fail if NLM_F_DUMP_INTR flag is set

### DIFF
--- a/nl/nl_linux.go
+++ b/nl/nl_linux.go
@@ -559,6 +559,11 @@ done:
 			if m.Header.Pid != pid {
 				continue
 			}
+
+			if m.Header.Flags&unix.NLM_F_DUMP_INTR != 0 {
+				return nil, syscall.Errno(unix.EINTR)
+			}
+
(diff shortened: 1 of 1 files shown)


nl/nl_linux.go
```diff
@@ -661,12 +666,14 @@ func GetNetlinkSocketAt(newNs, curNs netns.NsHandle, protocol int) (*NetlinkSock
...
 // at the end of the code that needs to be executed in the network namespace.
 // Example:
-// func jobAt(...) error {
-//      d, err := executeInNetns(...)
-//      if err != nil { return err}
-//      defer d()
-//      < code which needs to be executed in specific netns>
-//  }
+//
+//	func jobAt(...) error {
```
adrianchiris, line 670: while unrelated, linter keeps fixing this because of style issues.

adrianchiris:
this check is similar to what is perfomed in libmnl[1]

[1] https://github.com/justmirror/libmnl/blob/f14732339a77a2e6ba9ed4ca99b347ce1dc60801/src/callback.c#L70

aboch:
LGTM

PR #934, opened 2023-12-06
qdisc: Add Clsact

kevsecurity:
Qdisc implementation was missing the clsact qdisc which would allow adding filters to both ingress and egress. See https://lwn.net/Articles/671458/

We have been running a fork with this patch for approx 1 year. Seems stable enough to propose to upstream.


borkmann:
Hi 👋 @aboch, if you get a chance, please take a look. :)

aboch:
LGTM

PR #935, opened 2023-12-06
Allow NLMSG_DONE with no data

adrianchiris:
certain implementations in kernel return NLMSG_DONE with no data. If that is the case, assume no error occured.

currently we panic.

Note:
This behaviour exists in devlink subsystem, and possibly others.
clarification has been added to kernel docs[1]

[1] https://git.kernel.org/pub/scm/linux/kernel/git/netdev/net-next.git/commit/?id=8ad55b1e73c4e77656e2bea68fa2b484f33a6425


adrianchiris:
@aboch @vishvananda  PTAL :)

aboch:
LGTM

PR #938, opened 2023-12-24
add group_fwd_mask support for bridge and bridge port

hujun-open:
group_fwd_mask is useful to make certain Ethernet link local traffic like LLDP to pass-through bridge;
this PR provides capability equivalent to command `ip link add <name> type bridge group_fwd_mask <mask` and `ip link set <name> type bridge_slave group_fwd_mask <mask>`


aboch:
LGTM

PR #940, opened 2024-01-03
Add iterator-based version of RouteListFilter

fasaxc:
The RouteListFiltered method can be quite inefficient when there are many (say 100k) routes.  First all 100k routes are loaded into fragmented byte slices, appended to a slice (which must be grown through several orders of magnitude, requiring a lot of copying), then the whole `[][]byte` is kept in RAM while all the routes are parsed and returned in a second large slice.  

This PR adds a callback-based iterator equivalent: RouteListFilteredIter.  RouteListFilteredIter loads one route from the kernel, parses it fully and passes it to the callback, then it moves to the next route.  This should be a lot more efficient, especially when the caller is further converting the routes.  To avoid copying too much code, RouteListFiltered is now a simple wrapper around RouteListFilteredIter, which should be more efficient due to avoiding creation of the intermediate `[][]byte`.

Similarly, the main work of NetlinkRequest.Execute is moved to ExecuteIter.

### DIFF
--- a/netlink_test.go
+++ b/netlink_test.go
@@ -23,7 +23,7 @@ import (
 
 type tearDownNetlinkTest func()
 
-func skipUnlessRoot(t *testing.T) {
+func skipUnlessRoot(t testing.TB) {
 	t.Helper()
 
 	if os.Getuid() != 0 {
@@ -53,7 +53,7 @@ func skipUnlessKModuleLoaded(t *testing.T, module ...string) {
 	}
 }
 
-func setUpNetlinkTest(t *testing.T) tearDownNetlinkTest {
+func setUpNetlinkTest(t testing.TB) tearDownNetlinkTest {
 	skipUnlessRoot(t)
 
 	// new temporary namespace so we don't pollute the host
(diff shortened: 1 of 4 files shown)


fasaxc:
I ran some benchmarks to confirm:
```
BenchmarkRouteListFilteredOld
BenchmarkRouteListFilteredOld-12    	      12	  99193504 ns/op	164265036 B/op	  329341 allocs/op
BenchmarkRouteListFilteredNew
BenchmarkRouteListFilteredNew-12    	      12	  88368757 ns/op	155487153 B/op	  329313 allocs/op
BenchmarkRouteListIter
BenchmarkRouteListIter-12           	      27	  41802746 ns/op	27340148 B/op	  329285 allocs/op
PASS
```
BenchmarkRouteListFilteredOld/New compare the normal RouteListFiltered() method; the new approach seems to give a ~10% improvement even when building the output slice.   BenchmarkRouteListIter iterates the routes but discards them; it's 2x as fast and reduces RAM usage.

fasaxc:
@aboch @vishvananda Any interest in this kind of approach?  I think the reduction in GC burden is pretty large if you have a lot of routes and the performance is better whether you use the new API or the old.  It also lays the groundwork for adding `Iter` versions of any other API methods that would benefit.

fasaxc:
Rebased again, and looks like the build is now green.

aboch:
LGTM

fasaxc:
Thanks @aboch!

PR #941, opened 2024-01-08
Change behavior of `Subscribe` to non-blocking. Fix test.

kuroa-me:
When listening for IPv6 address changes, I found that a netlink socket that is subscribed to a group is not returning from `Receive` when there is message in the socket. After some researching, I found that libnl suggest setting socket to non-blocking when subscribing to groups. [Ref](https://www.infradead.org/~tgr/libnl/doc/core.html#:~:text=best%20to%20put%20the%20socket%20in%20non-blocking%20mode) This is probably caused by a bug in kernel. 

My test machine runs in 5.10, before, `addrSubscribeAt` would only exit on timeout; after the patch, it correctly sending addr to channels on each new addr.

Also fixed test related to BareUDP, which requires "bareudp" kmod. [Ref](https://www.kernelconfig.io/config_bareudp)

### DIFF
--- a/addr_linux.go
+++ b/addr_linux.go
@@ -365,6 +365,9 @@ func addrSubscribeAt(newNs, curNs netns.NsHandle, ch chan<- AddrUpdate, done <-c
 		for {
 			msgs, from, err := s.Receive()
 			if err != nil {
+				if err == syscall.EAGAIN {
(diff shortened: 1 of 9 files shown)


aboch:
LGTM

Issue #947, opened 2024-01-31
AddrAdd, AddReplace and AddrDel should not require a Link


cwmos:
I propose the three methods:
```
AddrAdd(link Link, addr *Addr) error
AddrReplace(link Link, addr *Addr) error
AddrDel(link Link, addr *Addr) error
```
Have their signatures changed to:
```
AddrAdd(addr *Addr) error
AddrReplace(addr *Addr) error
AddrDel(addr *Addr) error
```
The link arguments are not needed because the interface index is already in the Addr struct.

This would make it similar to how it works for Neighbor and Routes. It would also give the nice property that the information you get about an address in AddrList is exactly the same as you need to give to the functions here.

The implementation does also use Name of the link in order to test if the label has it as prefix:
```
base := link.Attrs()
if addr.Label != "" && !strings.HasPrefix(addr.Label, base.Name) {
  return fmt.Errorf("label must begin with interface name")
}
```
But presumably this check is not really needed? It seems like it is the kernel that should make this check if it is needed.

I would be happy to make a PR with this change.

aboch:
Please go ahead.
We cannot break current contract though, and golang does not support function overloading.

PR #948, opened 2024-01-31
Allow link argument to AddrAdd, AddReplace and AddrDel to be nil

cwmos:
This to address https://github.com/vishvananda/netlink/issues/947

### DIFF
--- a/addr_linux.go
+++ b/addr_linux.go
@@ -74,17 +74,19 @@ func (h *Handle) AddrDel(link Link, addr *Addr) error {
 }
 
 func (h *Handle) addrHandle(link Link, addr *Addr, req *nl.NetlinkRequest) error {
-	base := link.Attrs()
-	if addr.Label != "" && !strings.HasPrefix(addr.Label, base.Name) {
-		return fmt.Errorf("label must begin with interface name")
-	}
-	h.ensureIndex(base)
-
 	family := nl.GetIPFamily(addr.IP)
-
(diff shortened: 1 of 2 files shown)


jellonek:
lgtm

jellonek:
@aboch, @vishvananda  can you take a look? imo this change is clean and simple.

aboch:
Thank you @cwmos and @jellonek 
Would it be possible to add a UT function for this?
Programming and then reading from kernel an address where the link parameter is nil, if not there yet.

jellonek:
imo it is in the same bucket as checking if user provided non nil `addr`. we are not checking that what leads to panic, because developer calling this function did a horrible mistake.
UT would be nice to have, but imo is not needed. (i would merge that on your place :P )

aboch:
If you do not want to add the ut to cover the following two lines
```
	if link == nil {
		msg.Index = uint32(addr.LinkIndex)
```
then I have to do it myself.
Believe me, it is going to take longer than one of you guys doing it.
Please add it.

cwmos:
I agree a unit test is in place here - I don't want this use case to break in the future :-)

I will try to add one.

aboch:
Thank you @cwmos 
Please squash your two commits into one.
This project follows a "one commit per feature/fix/enhancement" approach

cwmos:
Done @aboch

aboch:
LGTM

PR #961, opened 2024-03-26
Fix: Cilium 30744

foyerunix:
Hello,
 
After analyzing Cilium issue [30744](https://github.com/cilium/cilium/issues/30744) we found that the root cause come from the "RouteListFiltered" function in this library.

The "RouteListFiltered" function take a parameter "family" to filter by IP family.
 
On a host with IPv6 enabled, filtering with family = 10 works correctly. On a host with IPv6 disabled, filtering with family = 10 return IPv4 routes. We also saw that using an unknow family (for example 55) return all routes. Using strace we can see that this come from the behavior of the RTM_GETROUTE operation.
 
We suggest to check that the family of each returned route match the family requested by the caller before returning the results.
 
Reproducer:

```
package main
 
import (
    "fmt"
)
 
func main() {
  routes, err := netlink.RouteListFiltered(55, &netlink.Route{Table: 2004}, netlink.RT_FILTER_TABLE)
 
  if err != nil {
    fmt.Errorf("Unable to list table routes: %s", err)
    return
  }
 
        for _, route := range routes {
                fmt.Printf("Destination: %s, Family: %d, Table: %d\n",
                        route.Dst, route.Family, route.Table)
        }
}
````

### CODE REVIEW

**Comment by aboch** [comment]
Thanks @foyerunix 
Can you confirm that the equivalent iproute2 command behave the same as your proposed fix?
Also, please rephrase your commit's title and do not refer tickets in other projects.
Something like `Filter RouteListFiltered response by requested IP family in ` would do it.

**Comment by foyerunix** [comment]
Hello @aboch 

If I add a route in the 2004 table using `ip route add local default dev lo table 2004` I can confirm that the command `ip -6 route show table 2004` doesn't return any route. However `reproducer.go`, when setting family = 10, return the route.

Interestingly using `strace -f -e sendto,recvmsg,recvfrom ip -6 route show table 2004` I can confirm that the AF_INET route is returned at this syscall:

 ```
recvmsg(3, {msg_name={sa_family=AF_NETLINK, nl_pid=0, nl_groups=00000000}, msg_namelen=12, msg_iov=[{iov_base=[{nlmsg_len=44, nlmsg_type=RTM_NEWROUTE, nlmsg_flags=NLM_F_MULTI|NLM_F_DUMP_FILTERED, nlmsg_seq=1711523424, nlmsg_pid=2665}, {rtm_family=AF_INET, rtm_dst_len=0, rtm_src_len=0, rtm_tos=0, rtm_table=RT_TABLE_COMPAT, rtm_protocol=RTPROT_BOOT, rtm_scope=RT_SCOPE_HOST, rtm_type=RTN_LOCAL, rtm_flags=0}, [[{nla_len=8, nla_type=RTA_TABLE}, 0x7d4], [{nla_len=8, nla_type=RTA_OIF}, if_nametoindex("lo")]]], iov_len=32768}], msg_iovlen=1, msg_controllen=0, msg_flags=0}, 0) = 44
```

However the library get the results from the syscall `recvfrom` and get every routes and not only the routes of the table 2004, whether IPv6 or IPv4. From the results of strace,  it seem that the design of the library is different from the design of iproue2.

Best Regards.

foyerunix:
Hi again @aboch 

I rephrased the commit, is is ok for you?

Best regards,

aboch:
LGTM

PR #962, opened 2024-03-27
rule: fix parsing zero priority rule

### LINKED ISSUE #709: [ip-rule] priority 0 rule returned as -1
On my machine, when I use the netlink.RuleList() method the 0 priority rule is returned as "-1". 

iproute2 command
```
  ip rule list
0:      from all lookup local
32766:  from all lookup main
32767:  from all lookup default
```

netlink library output:
```
=== RUN   TestNetLink
    manager_test.go:251: ip rule -1: from <nil> table 255
    manager_test.go:251: ip rule 32766: from <nil> table 254
    manager_test.go:251: ip rule 32767: from <nil> table 253
```

Test Code:
```
func TestNetLink(t *testing.T) {
	rules, err := netlink.RuleList(netlink.FAMILY_V4)
	if err != nil {
		t.Fatal(err)
	}
	for _, rule := range rules {
		t.Logf("%v", rule)
	}
}
```

jronak:
Fixes #709 (see issue for repro)

This happens as the rule with priority=0 don't contain `nl.FRA_PRIORITY` attribute, which leaves the priority in the struct as -1 which was set by `NewRule`. 

Wrote a test which fails before the fix:
```
--- FAIL: TestRuleListFiltered (0.01s)
    --- FAIL: TestRuleListFiltered/IPv4 (0.00s)
        --- FAIL: TestRuleListFiltered/IPv4/returns_rules_with_highest_priority (0.00s)
            rule_test.go:373: Expected len: 1, got: 0
    --- FAIL: TestRuleListFiltered/IPv6 (0.00s)
        --- FAIL: TestRuleListFiltered/IPv6/returns_rules_with_highest_priority (0.00s)
            rule_test.go:373: Expected len: 1, got: 0
```


rule_linux.go
```diff
@@ -279,6 +279,11 @@ func (h *Handle) RuleListFiltered(family int, filter *Rule, filterMask uint64) (
 			}
 		}
 
+		// Fallback to priority zero when priority is not set by attribute.
+		if rule.Priority == -1 {
+			rule.Priority = 0
```
aboch, line 284: Please move this right after the default rule struct we create at line 223:
```
rule := NewRule()
rule.Priority = 0    // The default priority is 0
```

rule_test.go
```diff
@@ -222,6 +222,26 @@ func runRuleListFiltered(t *testing.T, family int, srcNet, dstNet *net.IPNet) {
 				return rs, false
 			},
 		},
+		{
+			name:       "returns one rule filtered by Priority(0) and Table",
+			ruleFilter: &Rule{Priority: 0, Table: 1},
```
aboch, line 227: In my opinion, we should also test the situation where the rule priority is not specified by us when we program the kernel. That happens when we client uses the NewRule() method, and Priority is set to -1.

aboch:
@jronak

rule_test.go
jronak: Updated tests to add rule with priority unset, which results into rule creation with priority preceding main table (32766).

aboch:
Thanks. Please squash your two commits into one.

jronak:
Squashed commits into one, thanks!

aboch:
LGTM

PR #963, opened 2024-04-01
Add support for TCA_HTB_DIRECT_QLEN in HTB qdisc

bc-lee:
- Extend Htb struct in qdisc.go to include DirectQlen field
- Implement the DirectQlen option in qdisc_linux.go
- Modify TestHtbAddDel test to validate DirectQlen changes

### DIFF
--- a/qdisc.go
+++ b/qdisc.go
@@ -123,6 +123,7 @@ type Htb struct {
 	Defcls       uint32
 	Debug        uint32
 	DirectPkts   uint32
+	DirectQlen   *uint32
 }
 
 func NewHtb(attrs QdiscAttrs) *Htb {
@@ -133,6 +134,7 @@ func NewHtb(attrs QdiscAttrs) *Htb {
 		Rate2Quantum: 10,
 		Debug:        0,
 		DirectPkts:   0,
+		DirectQlen:   nil,
 	}
 }
 
--- a/qdisc_linux.go
+++ b/qdisc_linux.go
@@ -200,7 +200,9 @@ func qdiscPayload(req *nl.NetlinkRequest, qdisc Qdisc) error {
 		opt.Debug = qdisc.Debug
 		opt.DirectPkts = qdisc.DirectPkts
 		options.AddRtAttr(nl.TCA_HTB_INIT, opt.Serialize())
-		// options.AddRtAttr(nl.TCA_HTB_DIRECT_QLEN, opt.Serialize())
+		if qdisc.DirectQlen != nil {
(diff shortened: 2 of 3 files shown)


qdisc.go
```diff
@@ -133,6 +134,7 @@ func NewHtb(attrs QdiscAttrs) *Htb {
 		Rate2Quantum: 10,
 		Debug:        0,
 		DirectPkts:   0,
+		DirectQlen:   0xFFFFFFFF,
```
aboch, line 137: I assume `0` is a valid value for `DirectQlen` ?

bc-lee: > I assume `0` is a valid value for `DirectQlen` ?

Yes, 0 is a valid value for `DirectQlen`. Also, 0xFFFFFFFF(=4294967295) is a valid value for `DirectQlen` as well. (Although `tc` command doesn't allow to set `direct_qlen` to 4294967295)

```sh
# Create a qdisc
$ tc qdisc add dev enp6s0 root handle 1:0 htb
# Create a class
$ tc class add dev enp6s0 parent 1:0 classid 1:1 htb rate 1000mbit ceil 1000mbit
# Create a qdisc
$ tc qdisc add dev enp6s0 parent 1:1 handle 10:0 htb default 30 direct_qlen 4294967295
# Describe the qdisc
$ tc qdisc show dev enp6s0
qdisc htb 1: root refcnt 2 r2q 10 default 0 direct_packets_stat 226 direct_qlen 1000
qdisc htb 10: parent 1:1 r2q 10 default 0x30 direct_packets_stat 0 direct_qlen 1000
```
Check the last line of the output. The `direct_qlen` is set to 1000, not 4294967295.

However, after testing, `TCA_HTB_DIRECT_QLEN` itself doesn't have a restriction on the value of `direct_qlen`. So, I modified the patch to allow any value for `direct_qlen`.

aboch: As this would be the first optional HTB field, I just to make sure, existing fields are all mandatory?

image

bc-lee: I think so. All fields you mentioned are in [tc_htb_glob](https://github.com/torvalds/linux/blob/8f2c057754b25075aa3da132cd4fd4478cdab854/include/uapi/linux/pkt_sched.h#L418) struct, so we have to set them all.

aboch:
LGTM

aboch:
@bc-lee please squash the two commits into one

bc-lee:
> @bc-lee please squash the two commits into one

Done.

PR #965, opened 2024-04-07
Add zone filter to conntrack

hongliangl:


conntrack_linux.go
```diff
@@ -438,6 +450,7 @@ type ConntrackFilter struct {
 	portFilter  map[ConntrackFilterType]uint16
 	protoFilter uint8
 	labelFilter map[ConntrackFilterType][][]byte
+	zoneFilter  *uint16
```
aboch, line 453: I was a bit unsure about a pointer for this, but I get the logic now. Thanks

hongliangl:
Thanks for merging this @aboch !

PR #966, opened 2024-04-08
Allow a Police to be specified directly on a U32 filter

cwmos:
It is already possible to specify it on an action of a U32 filter.

The example at https://man7.org/linux/man-pages/man8/tc-police.8.html specifies it directly on the U32 filter as supported by this PR. So presumably such a configuration is very common.

The code in this PR is almost just copy pasted from the corresponding code for FwFilter.


aboch:
LGTM

PR #967, opened 2024-04-09
Fix go race detector races

cwmos:
This fixes an issue I had where the Go race detector reported races.

Even though there technically is a race, the existing code probably works fine. However, it is nice to be able to use the Go race detector without having races reported.

I considered using sync.Once instead. However, I wanted to preserve the existing behavior, that in case of problems reading "/proc/net/psched", we try to read it again next time.

### DIFF
--- a/qdisc_linux.go
+++ b/qdisc_linux.go
@@ -5,6 +5,7 @@ import (
 	"io/ioutil"
 	"strconv"
 	"strings"
+	"sync"
 	"syscall"
 
 	"github.com/vishvananda/netlink/nl"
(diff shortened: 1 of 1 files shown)


cwmos:
Someone may have concerns about using a mutex for this for efficiency reasons?

I think it should be very efficient as in most cases it will just use CompareAndSwapInt32 - see https://cs.opensource.google/go/go/+/refs/tags/go1.22.2:src/sync/mutex.go;l=81

I could instead put the all the values into a struct and store the struct or a pointer to the struct in an atomic.Value. Perhaps that would be a bit nicer?

Or use a typesafe atomic.Pointer instead, but that requires go 1.19.

Alternatively I could make a solution using math.Float64bits, math.Float64frombits, atomic.LoadUint64 and atomic.StoreUint64. But that seems a bit cumbersome..

PR #976, opened 2024-05-23
Revert "Change behavior of `Subscribe` to non-blocking. Fix test."

tnqn:
This reverts commit 916f9685fa42fc711f76a7f48d058e2656f7edf6.

It doesn't revert the BareUDP test fix.

For #975

PR #977, opened 2024-05-23
Eliminate busy loop when receiving messages from non-blocking socket

### LINKED ISSUE #975: Subscribe and its variants run into a busy loop taking 100% CPU when groups is not empty
My colleage @hongliangl tried to upgrade netlink to a recent commit to pick up a required change. However, our CI became flaky when validating the new netlink version. We identified the flake started from the commit merged via https://github.com/vishvananda/netlink/pull/941.

The above commit changes the socket created by `Subscribe` to non-blocking when groups are provided. However, it didn't change how the message was received from the socket, causing the receiver goroutine to run into a busy loop, taking 100% CPU.
https://github.com/vishvananda/netlink/blob/856e190dd707c02002dcdf6434424ef8af375ada/addr_linux.go#L367-L372

The issue can be reproduced by the following code:
```go
// netlink-reproducer.go
package main

import (
	"fmt"
	"os"

)

func main() {
	ch := make(chan netlink.AddrUpdate, 1000)
	stopCh := make(chan struct{})
	defer close(stopCh)
	if err := netlink.AddrSubscribeWithOptions(ch, stopCh, netlink.AddrSubscribeOptions{
		ErrorCallback: func(err error) {
			fmt.Fprintf(os.Stderr, "Received err: %v\n", err)
		},
	}); err != nil {
		fmt.Fprintf(os.Stderr, "Failed to subscribe: %v\n", err)
	}
	for {
		select {
		case update := <-ch:
			fmt.Fprintf(os.Stderr, "Received: %v\n", update)
		}
	}
}
```

The process would always take 100%+ CPU:
```
    PID USER      PR  NI    VIRT    RES    SHR S  %CPU  %MEM     TIME+ COMMAND
 412687 root      20   0 1968136   8172   2660 S 103.6   0.0   0:15.53 netlink-reprodu
```

To fix it, all subscribers' receiver goroutines should use poll or select to wait for events first before receiving messages from the socket. I could take a stab at fixing the implementation, but I wonder **if we could revert the commit that introduces the bug first** to unblock projects requiring other changes of the library.

tnqn:
For a non-blocking socket, it should wait for events first before receiving messages from the socket, otherwise it would receive empty message and run into a busy loop.

This is an alternative of https://github.com/vishvananda/netlink/pull/976.

Fixes #975

### DIFF
--- a/nl/nl_linux_test.go
+++ b/nl/nl_linux_test.go
@@ -75,7 +75,7 @@ func TestIfSocketCloses(t *testing.T) {
 		for {
 			_, _, err := sk.Receive()
 			// Receive returned because of a timeout and the FD == -1 means that the socket got closed
-			if err == unix.EAGAIN && nlSock.GetFd() == -1 {
+			if nlSock.GetFd() == -1 {
 				endCh <- err
 				return
 			}
--- a/nl/nl_linux.go
+++ b/nl/nl_linux.go
@@ -632,6 +632,10 @@ type NetlinkSocket struct {
 	fd  int32
 	lsa unix.SockaddrNetlink
 	sync.Mutex
+
+	// pfd is non nil when the socket is in non-blocking mode, and is used to wait for events on the socket.
+	pfd         *unix.PollFd
+	pollTimeout int64
 }
 
 func getNetlinkSocket(protocol int) (*NetlinkSocket, error) {
@@ -728,17 +732,22 @@ func Subscribe(protocol int, groups ...uint) (*NetlinkSocket, error) {
 		return nil, err
 	}
 
+	var pfd *unix.PollFd
 	// Sometimes (socket_linux.go:SocketGet), Subscribe is used to create a socket
-	// that subscirbed to no groups. So we don't need to set nonblock there.
+	// that subscribes to no groups. So we don't need to set nonblock there.
 	if len(groups) > 0 {
 		if err := unix.SetNonblock(fd, true); err != nil {
 			unix.Close(fd)
 			return nil, err
 		}
+		pfd = &unix.PollFd{Fd: int32(fd), Events: unix.POLLIN}
 	}
 
 	s := &NetlinkSocket{
-		fd: int32(fd),
+		fd:  int32(fd),
+		pfd: pfd,
+		// poll blocks infinitely by default.
+		pollTimeout: -1,
 	}
 	s.lsa.Family = unix.AF_NETLINK
 
@@ -791,6 +800,13 @@ func (s *NetlinkSocket) Receive() ([]syscall.NetlinkMessage, *unix.SockaddrNetli
 	if fd < 0 {
 		return nil, nil, fmt.Errorf("Receive called on a closed socket")
 	}
+	// The socket is in non-blocking mode.
+	if s.pfd != nil {
+		if _, err := unix.Poll([]unix.PollFd{*s.pfd}, int(atomic.LoadInt64(&s.pollTimeout))); err != nil {
+			return nil, nil, fmt.Errorf("Error polling the socket: %w", err)
+		}
+	}
+
 	var fromAddr *unix.SockaddrNetlink
 	var rb [RECEIVE_BUFFER_SIZE]byte
 	nr, from, err := unix.Recvfrom(fd, rb[:], 0)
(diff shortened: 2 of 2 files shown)


tnqn:
cc @aboch @kuroa-me

nl/nl_linux_test.go
```diff
@@ -75,7 +75,7 @@ func TestIfSocketCloses(t *testing.T) {
 		for {
 			_, _, err := sk.Receive()
 			// Receive returned because of a timeout and the FD == -1 means that the socket got closed
-			if err == unix.EAGAIN && nlSock.GetFd() == -1 {
+			if nlSock.GetFd() == -1 {
```
tnqn, line 78: The err is no longer `unix.EAGAIN` because `Receive` doesn't return when blocking on `unix.Recvfrom`, instead, it blocks on `unix.Poll`, then after 2s timeout, it calls `unix.Recvfrom` and gets "bad file descriptor"

gihnius:
When will this PR be merged?

aboch:
@tnqn please resolve the conflicts

tnqn:
> @tnqn please resolve the conflicts

Thanks for reminding, will do.

aboch:
@tnqn ping

tnqn:
@aboch sorry for the late response. I'm trying to verify my change does resolve the issue described in https://github.com/vishvananda/netlink/pull/941, which is why it changed `Subscribe` to non-blocking, but I'm not able to reproduce the issue, all IPv6 addresses events are received correctly. Thus I'm not sure if the change is really helpful.

@kuroa-me could you help me understand how to reproduce the issue? Do you have a reproducer that I can use to validate my change? The following is the code I used to test, but it has never failed even with 1000 IPs added or deleted in batch.
```
func TestAddrSubscribeWithOptionsIPv6(t *testing.T) {
	defer tearDown()

	ch := make(chan AddrUpdate, 5000)
	done := make(chan struct{})
	defer close(done)
	var lastError error
	defer func() {
		if lastError != nil {
			t.Fatalf("Fatal error received during subscription: %v", lastError)
		}
	}()
	if err := AddrSubscribeWithOptions(ch, done, AddrSubscribeOptions{
		ListExisting: false,
		ErrorCallback: func(err error) {
			lastError = err
		},
	}); err != nil {
		t.Fatal(err)
	}

	// Create a dummy interface to add IPs.
	dummy := &Dummy{
		LinkAttrs: LinkAttrs{Name: "dummy0"},
	}
	if err := LinkAdd(dummy); err != nil {
		t.Fatal(err)
	}
	defer LinkDel(dummy)

	var ipNets []*net.IPNet
	for i := 1; i <= 1000; i++ {
		ip, cidr, _ := net.ParseCIDR(fmt.Sprintf("2001:db8::%d/64", i))
		ipNet := &net.IPNet{IP: ip, Mask: cidr.Mask}
		if err := AddrAdd(dummy, &Addr{IPNet: ipNet}); err != nil {
			t.Fatalf("Failed to add address %s: %v", ipNet, err)
		}
		ipNets = append(ipNets, ipNet)
	}
	for _, ipNet := range ipNets {
		if !expectAddrUpdate(ch, true, ipNet.IP) {
			t.Fatalf("Add update for %v not received as expected", ipNet.IP)
		}
	}
	for _, ipNet := range ipNets {
		if err := AddrDel(dummy, &Addr{IPNet: ipNet}); err != nil {
			t.Fatalf("Failed to delete address %s: %v", ipNet, err)
		}
	}
	for _, ipNet := range ipNets {
		if !expectAddrUpdate(ch, false, ipNet.IP) {
			t.Fatalf("Del update for %v not received as expected", ipNet.IP)
		}
	}
}
```


```
# go test -count 10 -run TestAddrSubscribeWithOptionsIPv6 ./
ok      github.com/vishvananda/netlink  4.624s

kuroa-me:
Hmm, interesting. Please give me some time on trying to reproduce this, I could not find my old test suits. But I do remember that this relates to SLACC when the interface is being `ifup` instead of manully giving the interface address. Maybe difference lies there?

kuroa-me:
Most intriguing, SLACC does not affect netlink from acquring LLA. I will keep seeking what is missing...

Issue #528, opened 2020-03-18
Change type of Mark / Mask in Rule?


antoninbas:
`Rule` is defined as follows in [rule.go](https://github.com/vishvananda/netlink/blob/master/rule.go):
```
type Rule struct {
  // ...
  Mark              int
  Mask              int
  // ...
}
```
The default value for `Mark` and `Mask` is -1 when creating an instance of `Rule` with `NewRule`, and this translates in the corresponding attributes being omitted from the generated netlink message:
```
	if rule.Mark >= 0 {
		b := make([]byte, 4)
		native.PutUint32(b, uint32(rule.Mark))
		req.AddData(nl.NewRtAttr(nl.FRA_FWMARK, b))
	}
	if rule.Mask >= 0 {
		b := make([]byte, 4)
		native.PutUint32(b, uint32(rule.Mask))
		req.AddData(nl.NewRtAttr(nl.FRA_FWMASK, b))
	}
```

I have a package that uses `github.com/vishvananda/netlink` as follows:
```
ipRule := netlink.NewRule()
ipRule.Mark = 1 << 31
ipRule.Mask = 0xffffffff # same as using the default (-1), but writing this makes sense to be explicit
...
```
When I try to build this code for a 32-bit architecture (in my case `arm/v7`), I get errors like this:
```
constant 2147483648 overflows int
constant 4294967295 overflows int
```

Would it be possible to change the type of `Mark` and `Mask` in order to have better support for 32-bit architectures? I would suggest making these fields `*int32` values.

(I have limited knowledge of `ip rule` but I am assuming that `1 << 31` is a legit fwmark value).

antoninbas:
Fixed by #983

PR #983, opened 2024-06-14
rule: fix 32-bit platforms don't support adding rules with a mark 0xF0000000

qxoqx:
rule: fix 32-bit platforms don't support adding rules with a mark value of 0x80000000/0xF0000000 ~ 0xF0000000/0xF0000000

https://github.com/vishvananda/netlink/blob/4d4ba1473f21ca0ccdecc301adbe4315db42718b/rule.go#L14
The maximum value for an `int` type on a 32-bit platform is 0x7FFFFFFF. Since 0xF0000000 exceeds this limit, we need to use `uint` instead of `int` to handle these values.

I also tried to use unit tests under x86 architecture and failed
```
$ GOARCH=386 go test -v
# github.com/vishvananda/netlink [github.com/vishvananda/netlink.test]
./route_test.go:2336:14: cannot use 0xFFFFFFFF (untyped int constant 4294967295) as int value in assignment (overflows)
FAIL	github.com/vishvananda/netlink [build failed]
```
https://github.com/vishvananda/netlink/blob/4d4ba1473f21ca0ccdecc301adbe4315db42718b/route_test.go#L2336

### DIFF
--- a/route_linux.go
+++ b/route_linux.go
@@ -1448,7 +1448,7 @@ type RouteGetOptions struct {
 	VrfName  string
 	SrcAddr  net.IP
 	UID      *uint32
-	Mark     int
+	Mark     uint32
 	FIBMatch bool
 }
 
@@ -1557,7 +1557,7 @@ func (h *Handle) RouteGetWithOptions(destination net.IP, options *RouteGetOption
 
 		if options.Mark > 0 {
 			b := make([]byte, 4)
-			native.PutUint32(b, uint32(options.Mark))
+			native.PutUint32(b, options.Mark)
 
 			req.AddData(nl.NewRtAttr(unix.RTA_MARK, b))
 		}
--- a/rule.go
+++ b/rule.go
@@ -10,8 +10,8 @@ type Rule struct {
 	Priority          int
 	Family            int
 	Table             int
-	Mark              int
-	Mask              int
+	Mark              uint32
+	Mask              *uint32
 	Tos               uint
 	TunID             uint
 	Goto              int
@@ -51,8 +51,8 @@ func NewRule() *Rule {
 		SuppressIfgroup:   -1,
 		SuppressPrefixlen: -1,
 		Priority:          -1,
-		Mark:              -1,
-		Mask:              -1,
+		Mark:              0,
+		Mask:              nil,
 		Goto:              -1,
 		Flow:              -1,
 	}
--- a/rule_linux.go
+++ b/rule_linux.go
@@ -102,14 +102,14 @@ func ruleHandle(rule *Rule, req *nl.NetlinkRequest) error {
 		native.PutUint32(b, uint32(rule.Priority))
 		req.AddData(nl.NewRtAttr(nl.FRA_PRIORITY, b))
 	}
-	if rule.Mark >= 0 {
+	if rule.Mark != 0 || rule.Mask != nil {
 		b := make([]byte, 4)
-		native.PutUint32(b, uint32(rule.Mark))
+		native.PutUint32(b, rule.Mark)
 		req.AddData(nl.NewRtAttr(nl.FRA_FWMARK, b))
 	}
-	if rule.Mask >= 0 {
+	if rule.Mask != nil {
 		b := make([]byte, 4)
-		native.PutUint32(b, uint32(rule.Mask))
+		native.PutUint32(b, *rule.Mask)
 		req.AddData(nl.NewRtAttr(nl.FRA_FWMASK, b))
 	}
 	if rule.Flow >= 0 {
@@ -242,9 +242,10 @@ func (h *Handle) RuleListFiltered(family int, filter *Rule, filterMask uint64) (
 					Mask: net.CIDRMask(int(msg.Dst_len), 8*len(attrs[j].Value)),
 				}
 			case nl.FRA_FWMARK:
-				rule.Mark = int(native.Uint32(attrs[j].Value[0:4]))
+				rule.Mark = native.Uint32(attrs[j].Value[0:4])
 			case nl.FRA_FWMASK:
-				rule.Mask = int(native.Uint32(attrs[j].Value[0:4]))
+				mask := native.Uint32(attrs[j].Value[0:4])
+				rule.Mask = &mask
 			case nl.FRA_TUN_ID:
 				rule.TunID = uint(native.Uint64(attrs[j].Value[0:8]))
 			case nl.FRA_IIFNAME:
@@ -297,7 +298,7 @@ func (h *Handle) RuleListFiltered(family int, filter *Rule, filterMask uint64) (
 				continue
 			case filterMask&RT_FILTER_MARK != 0 && rule.Mark != filter.Mark:
 				continue
-			case filterMask&RT_FILTER_MASK != 0 && rule.Mask != filter.Mask:
+			case filterMask&RT_FILTER_MASK != 0 && !ptrEqual(rule.Mask, filter.Mask):
 				continue
 			}
 		}
@@ -320,4 +321,14 @@ func (pr *RuleUIDRange) toRtAttrData() []byte {
 	native.PutUint32(b[0], pr.Start)
 	native.PutUint32(b[1], pr.End)
 	return bytes.Join(b, []byte{})
+}
+
+func ptrEqual(a, b *uint32) bool {
+	if a == b {
+		return true
+	}
+	if (a == nil) || (b == nil) {
+		return false
+	}
+	return *a == *b
 }
--- a/rule_test.go
+++ b/rule_test.go
@@ -252,7 +252,7 @@ func runRuleListFiltered(t *testing.T, family int, srcNet, dstNet *net.IPNet) {
 				r.Family = family
 				r.Table = 1
 				RuleAdd(r)
-				
+
 				r.Priority = 32765 // Set priority for assertion
 				return r
 			},
@@ -325,7 +325,7 @@ func runRuleListFiltered(t *testing.T, family int, srcNet, dstNet *net.IPNet) {
 		},
 		{
 			name:       "returns rules filtered by Mask",
-			ruleFilter: &Rule{Mask: 0x5},
+			ruleFilter: &Rule{Mask: &[]uint32{0x5}[0]},
 			filterMask: RT_FILTER_MASK,
 			preRun: func() *Rule {
 				r := NewRule()
@@ -333,7 +333,7 @@ func runRuleListFiltered(t *testing.T, family int, srcNet, dstNet *net.IPNet) {
 				r.Priority = 1 // Must add priority and table otherwise it's auto-assigned
 				r.Family = family
 				r.Table = 1
-				r.Mask = 0x5
+				r.Mask = &[]uint32{0x5}[0]
 				RuleAdd(r)
 				return r
 			},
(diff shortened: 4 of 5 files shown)


antoninbas:
May be related to https://github.com/vishvananda/netlink/issues/528

route_linux.go
```diff
@@ -1448,7 +1448,7 @@ type RouteGetOptions struct {
 	VrfName  string
 	SrcAddr  net.IP
 	UID      *uint32
-	Mark     int
+	Mark     uint
```
aboch, line 1451: Could we make it uint32 instead?

rule.go
```diff
@@ -51,8 +51,8 @@ func NewRule() *Rule {
 		SuppressIfgroup:   -1,
 		SuppressPrefixlen: -1,
 		Priority:          -1,
-		Mark:              -1,
-		Mask:              -1,
+		Mark:              0,
+		Mask:              0,
```
antoninbas, line 55: Shouldn't the defaults be `0xFFFFFFFF` to preserve the existing behavior?

qxoqx: [Complete test source code](https://gist.github.com/qxoqx/452b0174907fe616c650dbf1a4c76cb0)
```
AddRule(200, 20, -1, 0xFFFFFFFF)
AddRule(200, 21, 0, 0xFFFFFFFF)
AddRule(200, 22, 1, 0xFFFFFFFF)
AddRule(200, 23, 0xFFFFFFFF, 0xFFFFFFFF)
/*
	20:     from all fwmark 0 lookup 200
	21:     from all fwmark 0 lookup 200
	22:     from all fwmark 0x1 lookup 200
	23:     from all fwmark 0xffffffff lookup 200
*/
```
if mark == -1 will not set FRA_FWMARK, so the mask will be ignored, the phenomenon looks the same as mark == 0, if mask == 0xffffffff, the result is different.
		so the default value of mark should be 0, even I think rule.Mark >= 0 can be modified to rule.Mark > 0

```
AddRule(200, 24, 0xFFFFFFFF, 0)
AddRule(200, 25, 0xFFFFFFFF, 1)
AddRule(200, 26, 0xFFFFFFFF, -1)
AddRule(200, 27, 0xFFFFFFFF, 0xFFFFFFFF)
/*
	24:     from all fwmark 0xffffffff/0 lookup 200
	25:     from all fwmark 0xffffffff/0x1 lookup 200
	26:     from all fwmark 0xffffffff lookup 200
	27:     from all fwmark 0xffffffff lookup 200
*/
```
The above results look like, when the mask is -1, it is the same as when the mask is 0xFFFFFFFF,
		so the default value of the mask should be changed from -1 to 0xFFFFFFFF, but I found the following scenario
```
AddRule(200, 30, 1, -1)
AddRule(200, 31, 1, 0xFFFFFFFF)
AddRule(200, 32, -1, -1)
AddRule(200, 33, -1, 0xFFFFFFFF) // <----
AddRule(200, 34, 0, -1)
AddRule(200, 35, 0, 0xFFFFFFFF) // <----
/*
	30:     from all fwmark 0x1 lookup 200
	31:     from all fwmark 0x1 lookup 200
	32:     from all lookup 200
	33:     from all fwmark 0 lookup 200
	34:     from all lookup 200
	35:     from all fwmark 0 lookup 200
*/
```
When mark == 1, mask == -1 is the same as mask == 0xFFFFFFFF, but when mark == 0, mask == -1 is different from mask == 0xFFFFFFFF (pref 32,33), the same problem with mark == -1. This makes me very confused.

route_linux.go
qxoqx: I found only one usage for RouteGetOptions.Mark. It ignores all values less than or equal to 0,  so I don't think this change will have any impact.

https://github.com/vishvananda/netlink/blob/4d4ba1473f21ca0ccdecc301adbe4315db42718b/route_linux.go#L1558-L1563

rule.go
qxoqx: I have resubmitted the code and made the following changes. Please review again.

rule.go
```
		Mark:              0,
		Mask:              0xFFFFFFFF,
```

rule_linux.go
``` 
	if rule.Mark > 0 {
		b := make([]byte, 4)
		native.PutUint32(b, uint32(rule.Mark))
		req.AddData(nl.NewRtAttr(nl.FRA_FWMARK, b))
	}
	if rule.Mask >= 0 && rule.Mask < 0xFFFFFFFF || (rule.Mask >= 0xFFFFFFFF && rule.Mark > 0) {
		b := make([]byte, 4)
		native.PutUint32(b, uint32(rule.Mask))
		req.AddData(nl.NewRtAttr(nl.FRA_FWMASK, b))
```

route_linux.go
aboch: This is not about that. `RouteGetOptions ` is an exposed structure so are its members. It represent data as it is passed to the kernel via netlink route attributes. It is useful to the user to know what data the kernel expect:

https://github.com/torvalds/linux/blob/1dfe225e9af5bd3399a1dbc6a4df6a6041ff9c23/net/ipv4/route.c#L3240

```
mark = tb[RTA_MARK] ? nla_get_u32(tb[RTA_MARK]) : 0;
```

qxoqx: Thank you for your review, it was my oversight earlier.

I updated the test code in [route_test.go](https://github.com/vishvananda/netlink/pull/983/files#diff-91f9c3332b2469d3f9d128f07bab664061bf5d8d35125b32690fd173b93d1be4) to use RouteGetWithOptions to retrieve routes with marks in the range 0x80000000 to 0xFFFFFFFF (e.g. 0xAFFFFFFF, 0xBFFFFFFF)
 I can successfully get routes with marks 0xAFFFFFFF or 0xBFFFFFFF.
```
sudo GOARCH=386 go test -v -run TestRouteFWMarkOption
```

I tried changing RouteGetOptions.Mark back to `int` type, the route could't be retrieved correctly.
`fwmark1` is of type `uint`, but `Mark` is of type `int`. Therefore, after type conversion, `Mark` will be less than 0.
```
...
fwmark1 := uint(0xAFFFFFFF)
...
routes, err = RouteGetWithOptions(dstIP, &RouteGetOptions{Mark: int(fwmark1)})
...
```
If `Mark` less than 0, This won't add the mark attribute to the conditions.
https://github.com/vishvananda/netlink/blob/4d4ba1473f21ca0ccdecc301adbe4315db42718b/route_linux.go#L1558-L1563

So I think it‘s necessary to change the type of `Mark` from `int` to `uint`.

Does this indirectly indicate that the kernel's mark, although of type `int`, doesn't care about the sign bit?

aboch:
@antoninbas please check

rule.go
```diff
@@ -51,8 +51,8 @@ func NewRule() *Rule {
 		SuppressIfgroup:   -1,
 		SuppressPrefixlen: -1,
 		Priority:          -1,
-		Mark:              -1,
-		Mask:              -1,
+		Mark:              0,
+		Mask:              0xFFFFFFFF,
```
antoninbas, line 55: Thinking about this some more, it seems that one issue with this approach is that it is not possible to represent the `0x0/0xFFFFFFFF` mark/mask pair, as this is what we use for the defaults (and we use it to mean "match any mark").

I cannot thing of a solution for this that does not involve making at least one of these 2 fields a pointer (`*uint32`).
For example, with a `uint32` `Mark` and a `*uint32` `Mask`, the following logic could be used:
* if `Mark == 0` and `Mask == nil`, omit both attributes
* else if `Mark != 0` and `Mask == nil`, set the mark attribute but not the mask attribute
* else set both attributes

cc @aboch

route_linux.go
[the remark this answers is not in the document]
qxoqx: I have changed mark `uint` to `uint32`
I have changed mask `uint` to `*uint32`

rule.go
qxoqx: I fell into the trap of thinking that -1 == 0xFFFFFFFF, so I missed the 0x0/0xFFFFFFFF mark/mask pair.
I have made the changes and added some necessary tests in rule_test.go. Please review again.

rule_linux.go
```diff
@@ -297,7 +298,8 @@ func (h *Handle) RuleListFiltered(family int, filter *Rule, filterMask uint64) (
 				continue
 			case filterMask&RT_FILTER_MARK != 0 && rule.Mark != filter.Mark:
 				continue
-			case filterMask&RT_FILTER_MASK != 0 && rule.Mask != filter.Mask:
+			case filterMask&RT_FILTER_MASK != 0 &&
+				(rule.Mask != filter.Mask && (rule.Mask == nil || filter.Mask == nil || *rule.Mask != *filter.Mask)):
```
antoninbas, line 302: I think this is correct, but I personally find it hard to read.
IMO, it's good to have a utility function for this. Something like this:
```go
func ptrEqual[T comparable](a, b *T) bool {
	if (a == nil) != (b == nil) {
		return false
	}
	if a == nil {
		return true
	}
	return *a == *b
}
```
Source: https://github.com/kubernetes/utils/blob/18e509b52bc8/ptr/ptr.go#L65

And then this case can become: `case filterMask&RT_FILTER_MASK != 0 && !ptrEqual(rule.Mask, filter.Mask)`

rule_test.go
```diff
@@ -474,5 +672,8 @@ func ruleEquals(a, b Rule) bool {
 		a.Invert == b.Invert &&
 		a.Tos == b.Tos &&
 		a.IPProto == b.IPProto &&
-		a.Protocol == b.Protocol
+		a.Protocol == b.Protocol &&
+		a.Mark == b.Mark &&
+		(a.Mask == b.Mask || a.Mask != nil && b.Mask != nil && *a.Mask == *b.Mask ||
+			(a.Mark != 0 && (a.Mask == nil && *b.Mask == 0xFFFFFFFF || b.Mask == nil && *a.Mask == 0xFFFFFFFF)))
```
antoninbas, line 678: hopefully the `ptrEqual` utility function suggested above can make this more readable

antoninbas (commented):
The API changes look good to me, but I am not a maintainer for this repo.
Hopefully users don't find it confusing that `Rule{Mark: 0}` means "match all marks" but `Rule{Mark:1}` means "match mark 1 exclusively". To match mark 0 exclusively, one needs to use `Rule{Mark: 0, Mask: &mask}` with `mask := uint32(0xffffffff)`.
If other believe that this is confusing, the only solution is to make `Mark` a pointer as well, which also impacts usability to some extent.

rule_linux.go
aboch: Agree.
On a similar line, I prefer
```
func ptrEqual(int *a, int* b) bool {
        if a == b {
		return true
	}
	if (a == nil) || (b == nil) {
		return false
	}
	return *a == *b
}
which would avoid a deeper comparison if the caller passed the same pointer for both input parameters.
```

qxoqx:
> The API changes look good to me, but I am not a maintainer for this repo. Hopefully users don't find it confusing that `Rule{Mark: 0}` means "match all marks" but `Rule{Mark:1}` means "match mark 1 exclusively". To match mark 0 exclusively, one needs to use `Rule{Mark: 0, Mask: &mask}` with `mask := uint32(0xffffffff)`. If other believe that this is confusing, the only solution is to make `Mark` a pointer as well, which also impacts usability to some extent.


```
~@ubuntu:~$ ip rule add fwmark 0 table 300 pref 120
~@ubuntu:~$ ip rule add fwmark 1 table 300 pref 120
~@ubuntu:~$ ip rule add fwmark 0/0xFFFFFFFF table 300 pref 130
~@ubuntu:~$ ip rule add fwmark 1/0xFFFFFFFF table 300 pref 130
~@ubuntu:~$ ip rule show
120:	from all lookup 300  <-------  mark 0 match all marks
120:	from all fwmark 0x1 lookup 300
130:	from all fwmark 0 lookup 300
130:	from all fwmark 0x1 lookup 300
```
I used the `ip` command to add rules with `fwmark:0`, `fwmark:1`, `fwmark:0/0xFFFFFFFF`, and `fwmark: 1/0xFFFFFFFF`. There's a similar confusing phenomenon here.  
As an API interface, should we replicate the behavior of the `ip` command, or should we avoid potentially confusing behavior internally in the API.
What should be the decision?
@antoninbas

rule_linux.go
qxoqx: I have add func `ptrEqual `

rule_test.go
qxoqx: I have add func ptrEqual

antoninbas:
@qxoqx I was unaware that there was a similar behavior with `ip rule`, thanks for pointing it out.
Then I would say the current version looks good to me, but again I am not a maintainer for this repo so the decision is not mine.

rule_test.go
```diff
@@ -325,15 +325,15 @@ func runRuleListFiltered(t *testing.T, family int, srcNet, dstNet *net.IPNet) {
 		},
 		{
 			name:       "returns rules filtered by Mask",
-			ruleFilter: &Rule{Mask: 0x5},
+			ruleFilter: &Rule{Mask: &[]uint32{0x5}[0]},
```
antoninbas, line 328: this is not elegant or idiomatic in my opinion, please write a helper function:
```go
func maskAsPtr(mask uint32) *uint32 {
        return &mask
}
```

Then this becomes `Mask: maskAsPtr(0x5)`

aboch:
Thanks @antoninbas 

LGTM

rule_test.go
qxoqx: `&[]uint32{0x5}[0]` is the only way I know to get a pointer to a simple type in a single line of code. 
It can simplify a lot of code in unit tests, especially if there isn't an existing function to get a pointer to the corresponding simple type.

The PR has been merged, so I'm thinking of not making any more changes for now.

qxoqx:
Thank you all for review the PR!

PR #984, opened 2024-06-15
Expose sockets attribute of nl.Handle

dpeckett:
For advanced use cases where the user might supply one or more existing sockets, eg. those sent using `SCM_RIGHTS` and such.


dpeckett:
While probably technically possible, there's a lot more additional pain needed to make this work.

PR #989, opened 2024-07-08
Add ConntrackDeleteFilters

aroradaman:
ConntrackDeleteFilters enables users to delete flow entries that match any of the specified filters. This allows users to delete multiple flow entries with a single dump table call.

### DIFF
--- a/conntrack_linux.go
+++ b/conntrack_linux.go
@@ -70,7 +70,13 @@ func ConntrackUpdate(table ConntrackTableType, family InetFamily, flow *Conntrac
 // ConntrackDeleteFilter deletes entries on the specified table on the base of the filter
 // conntrack -D [table] parameters         Delete conntrack or expectation
 func ConntrackDeleteFilter(table ConntrackTableType, family InetFamily, filter CustomConntrackFilter) (uint, error) {
-	return pkgHandle.ConntrackDeleteFilter(table, family, filter)
+	return pkgHandle.ConntrackDeleteFilters(table, family, filter)
+}
+
+// ConntrackDeleteFilters deletes entries on the specified table matching any of the specified filters
+// conntrack -D [table] parameters         Delete conntrack or expectation
+func ConntrackDeleteFilters(table ConntrackTableType, family InetFamily, filters ...CustomConntrackFilter) (uint, error) {
+	return pkgHandle.ConntrackDeleteFilters(table, family, filters...)
 }
 
 // ConntrackTableList returns the flow list of a table of a specific family using the netlink handle passed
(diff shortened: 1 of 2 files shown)


conntrack_linux.go
```diff
@@ -144,12 +150,15 @@ func (h *Handle) ConntrackDeleteFilter(table ConntrackTableType, family InetFami
...
-			req2.Execute(unix.NETLINK_NETFILTER, 0)
-			matched++
+		for _, filter := range filters {
+			if match := filter.MatchConntrackFlow(flow); match {
+				req2 := h.newConntrackRequest(table, family, nl.IPCTNL_MSG_CT_DELETE, unix.NLM_F_ACK)
+				// skip the first 4 byte that are the netfilter header, the newConntrackRequest is adding it already
+				req2.AddRawData(dataRaw[4:])
+				req2.Execute(unix.NETLINK_NETFILTER, 0)
+				matched++
+				break
```
aojea, line 160: this may use one comment to indicate the flow was already deleted so no need to process more filters and we can continue in the next flow

fasaxc:
I was just looking for a function like this; hope it makes it in!

conntrack_test.go
```diff
@@ -311,7 +311,7 @@ func TestConntrackTableDelete(t *testing.T) {
 
 	// Flush entries of groupB
 	var deleted uint
-	if deleted, err = h.ConntrackDeleteFilter(ConntrackTable, unix.AF_INET, filter); err != nil {
+	if deleted, err = h.ConntrackDeleteFilters(ConntrackTable, unix.AF_INET, filter); err != nil {
```
aojea, line 314: you may better add a new test or expand this to delete multiple flows with two different filters to exercise the new functionality

aroradaman: I'm also considering passing the errors back to the caller. Currently, `Execute()` can return an error, but it is not being captured. 
Also returning the error will be difficult in case of multiple filters, we would have to decide if we want to execute immediately in case of error, or try out all the filters and then return the error at the end.

conntrack_linux.go
```diff
@@ -133,9 +139,9 @@ func (h *Handle) ConntrackUpdate(table ConntrackTableType, family InetFamily, fl
 	return err
 }
 
-// ConntrackDeleteFilter deletes entries on the specified table on the base of the filter using the netlink handle passed
```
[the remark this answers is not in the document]
aboch: Agree, and it is taken care of above:

image

aboch:
LGTM

conntrack_linux.go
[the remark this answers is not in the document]
thaJeztah: This is indeed a breaking change; the singular one above only took care of the function, but looks like it missed adding a `Handle.ConntrackDeleteFilter` (singular) method on `Handle`; see

- https://github.com/moby/moby/pull/48368#issuecomment-2307059733

```
73.74 # github.com/docker/docker/libnetwork/iptables
73.74 libnetwork/iptables/conntrack.go:78:28: nlh.ConntrackDeleteFilter undefined (type *netlink.Handle has no field or method ConntrackDeleteFilter)
73.74 libnetwork/iptables/conntrack.go:84:28: nlh.ConntrackDeleteFilter undefined (type *netlink.Handle has no field or method ConntrackDeleteFilter)
73.74 libnetwork/iptables/conntrack.go:105:13: nlh.ConntrackDeleteFilter undefined (type *netlink.Handle has no field or method ConntrackDeleteFilter)
```

PR #1004, opened 2024-08-06
fix: updates SizeofVfVlanInfo to address #1003

### LINKED ISSUE #1003: netlink.LinkSetVfVlanQosProto fails with `invalid argument` but equivalent ip link command succeeds
When running `netlink.LinkSetVfVlanQosProto` on the latest commit of `netlink@v1.2.1-beta.2.0.20240713210050-d13535d71ed3` and kernel version `Linux 6.1.0-23-amd64 x86_64`, netlink fails with `invalid argument`. The equivalent `ip link` command suggested in the comments above the function succeeds.  I've added some code below to help reproduce the issue.

```go
package main

import (
	"flag"
	"log/slog"
	"os"

)

var (
	pfname = flag.String("pfname", "", "pf name to lookup link")
	vf     = flag.Int("vf", 0, "vf ID")
	vlan   = flag.Int("vlan", 0, "vlan ID")
	qos    = flag.Int("qos", 0, "qos")
	proto  = flag.Int("proto", 0, "proto")
)

func main() {
	flag.Parse()

	slog.Info("using values",
		slog.String("pfname", *pfname),
		slog.Int("vf_id", *vf),
		slog.Int("vlan_id", *vlan),
		slog.Int("qos", *qos),
		slog.Int("proto", *proto),
	)

	pfLink, err := netlink.LinkByName(*pfname)
	if err != nil {
		slog.Error("could not run LinkByName", slog.String("err", err.Error()))
		os.Exit(1)
	}

	err = netlink.LinkSetVfVlanQosProto(pfLink, *vf, *vlan, *qos, *proto)
	if err != nil {
		slog.Error("could not run LinkSetVfVlanQosProto", slog.String("err", err.Error()))
		os.Exit(1)
	}

	slog.Info("done")
}
```

When running 

```
> netlink-tester -qos 1 -pfname ens1f0 -proto 33024 -vf 10 -vlan 10

2024/08/05 13:28:25 INFO using values pfname=ens1f0 vf_id=10 vlan_id=10 qos=1 proto=33024
2024/08/05 13:28:25 ERROR could not run LinkSetVfVlanQosProto err="invalid argument"
```

But when we run the command suggested [here](https://github.com/vishvananda/netlink/blob/d13535d71ed3e430e26c7538cd6266d026b26f54/link_linux.go#L673) there is no issue.

```
 ip link set ens1f0 vf 10 vlan 10 qos 1 proto 802.1q

> works fine
``` 

If we `strace` the commands, it looks like `ip link` is using the `nlmsg_type` `RTM_NEWLINK`, while `netlink` is using `RTM_SETLINK`.

mega-alex:
This PR updates the `SizeOfVlanInfo` to be 16 rather than 14 to pad the attribute to be a multiple of 4 bytes, which is required by netlink when using `LinkSetVfVlanQosProto()`. This appears to have always been an issue, but was not correctly caught by netlink until [this CVE](https://lore.kernel.org/linux-cve-announce/2024053015-CVE-2024-36017-bf96@gregkh/T/) was addressed, which explains why older kernels don't have an issue with this. 

Closes #1003

### DIFF
--- a/nl/link_linux.go
+++ b/nl/link_linux.go
@@ -321,7 +321,7 @@ const (
 const (
 	SizeofVfMac        = 0x24
 	SizeofVfVlan       = 0x0c
-	SizeofVfVlanInfo   = 0x0e
+	SizeofVfVlanInfo   = 0x10
 	SizeofVfTxRate     = 0x08
 	SizeofVfRate       = 0x0c
 	SizeofVfSpoofchk   = 0x08


SchSeba (approved):

aboch:
LGTM

PR #1006, opened 2024-08-22
rule: Rule.Type to allow adding/listing unreachable (RTN_UNREACHABLE)…

oxtoacart:
… rules

Updates #710

### DIFF
--- a/rule.go
+++ b/rule.go
@@ -28,6 +28,7 @@ type Rule struct {
 	IPProto           int
 	UIDRange          *RuleUIDRange
 	Protocol          uint8
+	Type              uint8
 }
 
 func (r Rule) String() string {
@@ -41,8 +42,8 @@ func (r Rule) String() string {
 		to = r.Dst.String()
 	}
 
-	return fmt.Sprintf("ip rule %d: from %s to %s table %d",
-		r.Priority, from, to, r.Table)
+	return fmt.Sprintf("ip rule %d: from %s to %s table %d %s",
+		r.Priority, from, to, r.Table, r.typeString())
 }
 
 // NewRule return empty rules.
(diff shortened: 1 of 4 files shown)


rule.go
```diff
@@ -28,6 +28,12 @@ type Rule struct {
 	IPProto           int
 	UIDRange          *RuleUIDRange
 	Protocol          uint8
+
+	// Type is the unix.RTN_* rule type, such as RTN_UNICAST
+	// or RTN_UNREACHABLE.
+	// When adding a new rule, zero means automatic.
+	// Only supported on Linux.
```
aboch, line 35: Why does this element need all this explanation while all other ones don't?

oxtoacart: I'm happy to remove it if you prefer, it just seemed like useful documentation for any consumer of this new field.

aboch: Yes let's remove it, unless it deviates from iproute2.
Assumption is who uses this library is familiar with iproute2 and can always look for reference uses there and check kernek code as well. Also you are adding UT which shows an example, which is great.
Thank you for your contribution.

oxtoacart: Comment removed.

rule_linux.go
```diff
@@ -332,3 +332,34 @@ func ptrEqual(a, b *uint32) bool {
...
 }
+
+func (r Rule) typeString() string {
+	switch r.Type {
+	case unix.RTN_UNSPEC: // zero
+		return ""
+	case unix.RTN_UNICAST:
+		return ""
+	case unix.RTN_LOCAL:
+		return " local"
```
aboch, line 343: I am not sure about having the leading space in the string. Let's have the print function add that. Not a big deal if we have a trailing space in the route entry when type == RTN_UNSPEC

rule.go
```diff
@@ -41,8 +42,8 @@ func (r Rule) String() string {
 		to = r.Dst.String()
 	}
 
-	return fmt.Sprintf("ip rule %d: from %s to %s table %d",
-		r.Priority, from, to, r.Table)
+	return fmt.Sprintf("ip rule %d: from %s to %s table %d%s",
```
aboch, line 45: I am referring to this stringification of the IP Rule put a space here please

rule_linux.go
oxtoacart: Done. Note that on Mac, we will always have a trailing space.

aboch:
LGTM

oxtoacart:
Thanks for the review @aboch !

PR #1008, opened 2024-08-23
Handle: add ConntrackDeleteFilter alias for backward compat

thaJeztah:
- relates to https://github.com/vishvananda/netlink/pull/989#discussion_r1679358097
- relates to https://github.com/moby/moby/pull/48368#issuecomment-2307059733

Commit c96b03b4befaf5cc418386ad9eaf1a5d15101b1e changed the signature of this method to accept a list of filters and renamed it to ConntrackDeleteFilters (plural).

This patch adds back ConntrackDeleteFilter as an alias, and marks it as deprecated in favor of the new version.

### DIFF
--- a/conntrack_linux.go
+++ b/conntrack_linux.go
@@ -69,6 +69,8 @@ func ConntrackUpdate(table ConntrackTableType, family InetFamily, flow *Conntrac
 
 // ConntrackDeleteFilter deletes entries on the specified table on the base of the filter
 // conntrack -D [table] parameters         Delete conntrack or expectation
+//
+// Deprecated: use [ConntrackDeleteFilter] instead.
 func ConntrackDeleteFilter(table ConntrackTableType, family InetFamily, filter CustomConntrackFilter) (uint, error) {
 	return pkgHandle.ConntrackDeleteFilters(table, family, filter)
 }
@@ -137,6 +139,14 @@ func (h *Handle) ConntrackUpdate(table ConntrackTableType, family InetFamily, fl
 
 	_, err = req.Execute(unix.NETLINK_NETFILTER, 0)
 	return err
+}
+
+// ConntrackDeleteFilter deletes entries on the specified table on the base of the filter using the netlink handle passed
+// conntrack -D [table] parameters         Delete conntrack or expectation
+//
+// Deprecated: use [Handle.ConntrackDeleteFilters] instead.
+func (h *Handle) ConntrackDeleteFilter(table ConntrackTableType, family InetFamily, filter CustomConntrackFilter) (uint, error) {
+	return h.ConntrackDeleteFilters(table, family, filter)
 }
 
 // ConntrackDeleteFilters deletes entries on the specified table matching any of the specified filters using the netlink handle passed
--- a/conntrack_unspecified.go
+++ b/conntrack_unspecified.go
@@ -11,6 +11,9 @@ type InetFamily uint8
 // ConntrackFlow placeholder
 type ConntrackFlow struct{}
 
+// CustomConntrackFilter placeholder
+type CustomConntrackFilter struct{}
+
 // ConntrackFilter placeholder
 type ConntrackFilter struct{}
 
@@ -29,7 +32,15 @@ func ConntrackTableFlush(table ConntrackTableType) error {
 
 // ConntrackDeleteFilter deletes entries on the specified table on the base of the filter
 // conntrack -D [table] parameters         Delete conntrack or expectation
+//
+// Deprecated: use [ConntrackDeleteFilter] instead.
 func ConntrackDeleteFilter(table ConntrackTableType, family InetFamily, filter *ConntrackFilter) (uint, error) {
+	return 0, ErrNotImplemented
+}
+
+// ConntrackDeleteFilters deletes entries on the specified table matching any of the specified filters
+// conntrack -D [table] parameters         Delete conntrack or expectation
+func ConntrackDeleteFilters(table ConntrackTableType, family InetFamily, filters ...CustomConntrackFilter) (uint, error) {
 	return 0, ErrNotImplemented
 }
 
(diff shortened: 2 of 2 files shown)


thaJeztah:
@aboch @aojea @aroradaman  PTAL

aroradaman (commented):
LGTM

aojea:
heh https://github.com/vishvananda/netlink/pull/989#discussion_r1679358097

/lgtm

conntrack_linux.go
```diff
@@ -196,10 +204,11 @@ type ProtoInfo interface {
 type ProtoInfoTCP struct {
 	State uint8
 }
+
 // Protocol returns "tcp".
-func (*ProtoInfoTCP) Protocol() string {return "tcp"}
+func (*ProtoInfoTCP) Protocol() string { return "tcp" }
 func (p *ProtoInfoTCP) toNlData() ([]*nl.RtAttr, error) {
-	ctProtoInfo := nl.NewRtAttr(unix.NLA_F_NESTED | nl.CTA_PROTOINFO, []byte{})
+	ctProtoInfo := nl.NewRtAttr(unix.NLA_F_NESTED|nl.CTA_PROTOINFO, []byte{})
```
thaJeztah, line 211: oh! just now notice the formatting changes (I had to push and jump in a meeting); probably my IDE doing a `gofumpt` 😞 

let me know if you prefer having this as a separate commit 😅

thaJeztah: I reverted the formatting changes; will open a separate PR to do a format of the code.

antoninbas:
Thanks for taking care of this. I hope @aboch can release v1.2.2 promptly with this change so that people can upgrade to a new release without having to deal with breaking changes.

conntrack_linux.go
```diff
@@ -139,6 +139,14 @@ func (h *Handle) ConntrackUpdate(table ConntrackTableType, family InetFamily, fl
 	return err
 }
 
+// ConntrackDeleteFilter deletes entries on the specified table on the base of the filter using the netlink handle passed
+// conntrack -D [table] parameters         Delete conntrack or expectation
+//
+// Deprecated: use [Handle.ConntrackDeleteFilters] instead.
```
antoninbas, line 145: why is this being marked as deprecated, when the top-level `ConntrackDeleteFilter` function is not?

thaJeztah: My thinking here was that the new one effectively replaces the existing one. So deprecation as indication for users to move over.


I'm fine removing though (honestly I think the change in signature probably would've been fine without a rename, but I guess it could break someone trying to match against an interface)

antoninbas: My point was that if this method is marked as deprecated, then `ConntrackDeleteFilter` should also be marked as deprecated, or it's inconsistent

https://github.com/vishvananda/netlink/blob/65a253d3751cff6da8274035e4548974a874f477/conntrack_linux.go#L72

thaJeztah:
> Thanks for taking care of this. I hope @aboch can release v1.2.2 promptly with this change so that people can upgrade to a new release without having to deal with breaking changes.

We haven't found time yet to bisect/dig into, but I should mention that we also ran into other issues with this update in Moby (docker engine); https://github.com/moby/moby/pull/48368#issuecomment-2307375318

antoninbas:
> > Thanks for taking care of this. I hope @aboch can release v1.2.2 promptly with this change so that people can upgrade to a new release without having to deal with breaking changes.
> 
> We haven't found time yet to bisect/dig into, but I should mention that we also ran into other issues with this update in Moby (docker engine); [moby/moby#48368 (comment)](https://github.com/moby/moby/pull/48368#issuecomment-2307375318)

This is likely because of this change: https://github.com/vishvananda/netlink/commit/acdc658b8613655ddb69f978e9fb4cf413e2b830
It was made a year ago but I guess there hasn't been an actual release of the netlink library for a while. You will probably need to update the moby code to check if `route.Dst` is "all zeros" instead of nil.

More generally, there has been quite a lot of changes to the API since the 1.2.0 release. New functions were added but there were also some "breaking" changes, most of them more subtle than just the removal of `ConntrackDeleteFilter`. The behavior of some API functions was changed (which is the issue you are running into), the type of some fields was changed (https://github.com/vishvananda/netlink/pull/983), etc. It seems that at the very least a new minor release (1.3.0) would have been more appropriate than a patch release (1.2.1).

conntrack_linux.go
thaJeztah: Ah! Sorry, misinterpreted your comment 😅 

Yes, that makes sense; I pushed, and deprecated that one as well


While working on that, I also found that there c96b03b4befaf5cc418386ad9eaf1a5d15101b1e did not add stubs for non-supported platforms, so I added those as well.

thaJeztah:
Thanks @antoninbas ! That sounds indeed like it's related; thanks for that pointer, I think we should be able to work with that 🫶 


> More generally, there has been quite a lot of changes to the API since the 1.2.0 release. New functions were added but there were also some "breaking" changes

Yes, we stuck to a tagged version (to not force others to update), but the beta was probably tagged a bit too soon (either that, or had been around for too long, and gathered too many changes since). 1.3.0 definitely would've given more flexibility (keeping the option open to tag 1.2.x versions if needed).

aboch:
Thanks @thaJeztah for adding back the Handle version of the original method.
Why do we need to deprecate the methods which take an individual filter? Do you see a problem in leaving them available?

thaJeztah:
Mostly to keep the option open to reduce the public API in future, and to reduce maintenance (admitted, for these that's fairly limited). Deprecated doesn't have to mean "remove" immediately. Happy to remove the deprecation though if that's no concern.

From a public API perspective, while looking at this, even started wondering if this change was strictly needed; the existing function accepts a custom filter, which must match an interface; a utility to construct a "compound" filter probably would've given the same (and more), e.g. `MultiFilter(filter1, filter2)` or even `AndFilter(filter1, filter2, ...)` 🙈

PR #1012, opened 2024-08-30
Fix SetSendTimeout/SetReceiveTimeout

robmry:
Before https://github.com/vishvananda/netlink/pull/793, netlink sockets were blocking.

Send/receive timeouts were implemented using `SO_SNDTIMEO`/`SO_RCVTIMEO` on the socket descriptor. But, that doesn't work now the socket is non-blocking.

Instead, set deadlines on the file read/write.

### DIFF
--- a/nl/nl_linux.go
+++ b/nl/nl_linux.go
@@ -4,6 +4,7 @@ package nl
 import (
 	"bytes"
 	"encoding/binary"
+	"errors"
 	"fmt"
 	"net"
 	"os"
@@ -11,6 +12,7 @@ import (
 	"sync"
 	"sync/atomic"
 	"syscall"
+	"time"
 	"unsafe"
 
 	"github.com/vishvananda/netns"
@@ -656,9 +658,11 @@ func NewNetlinkRequest(proto, flags int) *NetlinkRequest {
 }
 
 type NetlinkSocket struct {
-	fd   int32
-	file *os.File
-	lsa  unix.SockaddrNetlink
+	fd             int32
+	file           *os.File
+	lsa            unix.SockaddrNetlink
+	sendTimeout    int64 // Access using atomic.Load/StoreInt64
+	receiveTimeout int64 // Access using atomic.Load/StoreInt64
 	sync.Mutex
 }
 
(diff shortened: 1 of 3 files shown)


nl/nl_linux.go
```diff
@@ -847,16 +895,16 @@ func (s *NetlinkSocket) Receive() ([]syscall.NetlinkMessage, *unix.SockaddrNetli
 
 // SetSendTimeout allows to set a send timeout on the socket
 func (s *NetlinkSocket) SetSendTimeout(timeout *unix.Timeval) error {
-	// Set a send timeout of SOCKET_SEND_TIMEOUT, this will allow the Send to periodically unblock and avoid that a routine
-	// remains stuck on a send on a closed fd
-	return unix.SetsockoptTimeval(int(s.fd), unix.SOL_SOCKET, unix.SO_SNDTIMEO, timeout)
+	duration := (time.Duration(timeout.Sec) * time.Second) + (time.Duration(timeout.Usec) * time.Microsecond)
+	atomic.StoreInt64(&s.sendTimeout, int64(duration))
```
corhere, line 899: The `unix.Timeval` type already has a methods to convert to and from a `time.Duration` -- just without naming the type, [presumably because it's context-dependent whether a timeval represents a duration or timestamp.](https://go.dev/issues/2534#issuecomment-66060429)
```suggestion
	atomic.StoreInt64(&s.sendTimeout, timeout.Nano())
```

nl/nl_linux_test.go
```diff
@@ -97,6 +97,42 @@ func TestIfSocketCloses(t *testing.T) {
...
+	// Even if the test fails because the timeout doesn't work, closing the
+	// socket at the end of the test should result in an EAGAIN (as long as
+	// TestIfSocketCloses completed, otherwise this test will leak the
+	// goroutine running the Receive).
+	defer nlSock.Close()
+
+	// Set a short timeout on the Receive and a much longer timeout on
+	// the test so that, even if there are neighbour changes, it's very
+	// likely there will be a timeout.
+	nlSock.SetReceiveTimeout(&unix.Timeval{Sec: 0, Usec: 100000})
```
corhere, line 114:
```suggestion
	nlSock.SetReceiveTimeout(&(unix.NsecToTimeval(100*time.Millisecond)))
```

nl/nl_linux.go
robmry: Oh, that's handy! Thank you.

nl/nl_linux_test.go
robmry: Done.

robmry:
I'm not sure who needs to review this ... @aboch, could you take a look?

PR #1014, opened 2024-09-05
capture and return errors in ConntrackDeleteFilters

aroradaman:
`ConntrackDeleteFilters` does not capture and return errors for the actual netlink call to delete the flow entry.
This PR captures the error and returns it.

### DIFF
--- a/conntrack_linux.go
+++ b/conntrack_linux.go
@@ -6,6 +6,7 @@ import (
 	"errors"
 	"fmt"
 	"net"
+	"strings"
 	"time"
 
 	"github.com/vishvananda/netlink/nl"
@@ -158,6 +159,7 @@ func (h *Handle) ConntrackDeleteFilters(table ConntrackTableType, family InetFam
 	}
 
 	var matched uint
+	var errMsgs []string
 	for _, dataRaw := range res {
 		flow := parseRawData(dataRaw)
 		for _, filter := range filters {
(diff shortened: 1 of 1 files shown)


conntrack_linux.go
```diff
@@ -158,21 +159,27 @@ func (h *Handle) ConntrackDeleteFilters(table ConntrackTableType, family InetFam
...
 		for _, filter := range filters {
 			if match := filter.MatchConntrackFlow(flow); match {
 				req2 := h.newConntrackRequest(table, family, nl.IPCTNL_MSG_CT_DELETE, unix.NLM_F_ACK)
 				// skip the first 4 byte that are the netfilter header, the newConntrackRequest is adding it already
 				req2.AddRawData(dataRaw[4:])
-				req2.Execute(unix.NETLINK_NETFILTER, 0)
-				matched++
-				// flow is already deleted, no need to match on other filters and continue to the next flow.
-				break
+				if _, err = req2.Execute(unix.NETLINK_NETFILTER, 0); err != nil {
```
aboch, line 170: For readability, you may want to handle the successful case first:
```
if _, err = req2.Execute(unix.NETLINK_NETFILTER, 0); err == nil {
  matched++
 // flow is already deleted, no need to match on other filters and continue to the next flow.
 break
 }
 errMsgs = append(errMsgs, fmt.Sprintf("failed to delete conntrack flow '%s': %s", flow.String(), err.Error()))
 ```

aboch:
LGTM

PR #1015, opened 2024-09-08
bugfix: parse ipv4 src/dst error

Charlie17Li:
### Introduction

When I use tc-pedit module to modify the src_ipv4/dst_ipv4, I check the result but the src_ipv4/dst_ipv4 is always "0".

For example, I want to set the dst_ipv4 to “172.17.0.2”, and I check the result with `tc filter show dev eth0`:

```
# actual
action order 1: pedit action ...
   key #1 at ipv4+16 val 00000000 mask 00000000
```

but the result is expeted to this:
```
# expect, the val is "172.17.0.2" not "0"
action order 1: pedit action ...
   key #1 at ipv4+16 val ac110002 mask 00000000
```

### DIFF
--- a/nl/tc_linux.go
+++ b/nl/tc_linux.go
@@ -1533,7 +1533,7 @@ func (p *TcPedit) SetIPv6Dst(ip6 net.IP) {
 }
 
 func (p *TcPedit) SetIPv4Src(ip net.IP) {
-	u32 := NativeEndian().Uint32(ip[:4])
+	u32 := NativeEndian().Uint32(ip.To4())
 
 	tKey := TcPeditKey{}
 	tKeyEx := TcPeditKeyEx{}
@@ -1549,7 +1549,7 @@ func (p *TcPedit) SetIPv4Src(ip net.IP) {
 }
 
 func (p *TcPedit) SetIPv4Dst(ip net.IP) {
-	u32 := NativeEndian().Uint32(ip[:4])
+	u32 := NativeEndian().Uint32(ip.To4())
 
 	tKey := TcPeditKey{}
 	tKeyEx := TcPeditKeyEx{}


Charlie17Li:
I just remove the unit test, because I found that parsing the pedit action is not supported now.

https://github.com/vishvananda/netlink/blob/b1ce50cfa9bea7652382b8c5ef6083f8d3f5f853/filter_linux.go#L800-L820

Charlie17Li:
cc @aboch @chent1996

nl/tc_linux.go
```diff
@@ -1533,7 +1533,7 @@ func (p *TcPedit) SetIPv6Dst(ip6 net.IP) {
 }
 
 func (p *TcPedit) SetIPv4Src(ip net.IP) {
-	u32 := NativeEndian().Uint32(ip[:4])
+	u32 := NativeEndian().Uint32(ip[12:16])
```
aboch, line 1536: Would it not be better to force the input `ip` into v4 notation with `ip.To4()` rather than assuming the passed ip is always in IPv4 in IPv6 notation?

Charlie17Li: good idea! I will fix it.

chent1996: sorry, I only used ipv6 so I didn't check ipv4 before. But thank you for fixing this :)

nl/tc_linux.go
```diff
@@ -1460,15 +1460,15 @@ func (p *TcPedit) SetIPv6Src(ip6 net.IP) {
 
 func (p *TcPedit) SetDstIP(ip net.IP) {
 	if ip.To4() != nil {
-		p.SetIPv4Dst(ip)
+		p.SetIPv4Dst(ip.To4())
```
aboch, line 1463: @Charlie17Li 
I think you need to bring back your original change along with the `To4()` modification.

`SetIPv4Dst` and `SetIPv4Src` are public methods. It is better the `To4()` change is done inside those two methods.

Charlie17Li: ok

aboch:
Thank you @Charlie17Li .
Is it possible to add a new UT or extend an existing one to program and then read back this filter and make sure what is programmed in kernel is what we wanted to program?

aboch:
LGTM

PR #1023, opened 2024-10-01
fix: add missing CLOEXEC flag

smira:
Some calls were already using it, some were not, but fix the remaining ones.

Without this flag, the file descriptor would to the child process after fork/exec.


kolyshkin (commented):
LGTM(nb)

PR #1027, opened 2024-11-05
Add support for the `expires` option of `ip route`

albertjin:
This is implemented through the attributes RTA_EXPIRES and RTA_CACHEINFO.

### DIFF
--- a/route.go
+++ b/route.go
@@ -45,6 +45,17 @@ type Encap interface {
 	Equal(Encap) bool
 }
 
+type RouteCacheInfo struct {
+	Users   uint32
+	Age     uint32
+	Expires int32
+	Error   uint32
+	Used    uint32
+	Id      uint32
+	Ts      uint32
+	Tsage   uint32
+}
+
 //Protocol describe what was the originator of the route
 type RouteProtocol int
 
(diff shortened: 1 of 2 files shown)


route.go
```diff
@@ -115,6 +128,9 @@ func (r Route) String() string {
 	elems = append(elems, fmt.Sprintf("Flags: %s", r.ListFlags()))
 	elems = append(elems, fmt.Sprintf("Table: %d", r.Table))
 	elems = append(elems, fmt.Sprintf("Realm: %d", r.Realm))
+	if r.Expires != 0 {
+		elems = append(elems, fmt.Sprintf("Expires: %dsec", r.Expires))
```
aboch, line 132: should we put a space b/w before "sec"?

albertjin: This format is inherited from the `ip route` command. Note that the value can be negative because the kernel removes expired entries in a garbage collecting style.

albertjin:
> Thanks @albertjin pls squash the two commits into one and push again. This project follows a "1 commit per functionality/fix" approach.

I made a mistake in squashing and gave up. Simply recreated a pull request https://github.com/vishvananda/netlink/pull/1029

Issue #1030, opened 2024-11-18
`ipset` revision usage is inconsistent with `ipset` binary from netfilter


bleggett:
Example:

For `hash:ip,port`, this library defaults to `revision=1`: https://github.com/vishvananda/netlink/blob/main/ipset_linux.go#L398.

But the `ipset` binary code comments indicate we should be using at least `revision=3` here (since this library supports using comments with that set type) https://github.com/Olipro/ipset/blob/9f145b49100104d6570fe5c31a5236816ebb4f8f/kernel/net/netfilter/ipset/ip_set_hash_ipport.c#L30

For `hash:ip`, this library defaults to `revision=0`. But the ipset binary code comments indicate we should be using at least `revision=2` here (since this library supports using comments with this set type as well): https://github.com/Olipro/ipset/blob/9f145b49100104d6570fe5c31a5236816ebb4f8f/kernel/net/netfilter/ipset/ip_set_hash_ip.c#L28

In general, we should probably track the set-level revisions that the `ipset` binary does.

aboch:
Thanks @bleggett 
Please push a PR

PR #1031, opened 2024-11-18
Mimic `ipset` C code for determining correct default ipset revision for `hash:ip{port,net,etc}`

bleggett:
Fixes https://github.com/vishvananda/netlink/issues/1030

It takes both desired features AND set type into account when determining the "correct" default revision, which is how it actually works in the `ipset` userspace binary.

e.g.

`hash:ip,port,net`

- with `comment` support? Revision 5.
- with `counter` support? Revision 4.

etc etc.

A more exhaustive mapping would be ideal, but this PR only does default revision mappings for the ipset types that previously had revision defaulting logic, and not all possible ipset types globally.


aboch:
LGTM

PR #1034, opened 2024-11-26
Fix parsing 4-bytes attribute

Asphaltt:
What if the data length of attribute is 4? The attribute will be ignored, because `i+4 < len(data)`.


aboch:
@adrianchiris pls check

adrianchiris (approved):

PR #1037, opened 2024-12-05
disable broadcast if broadcast is set to net.IPv4zero

WeidiDeng:
Another go at fixing [329](https://github.com/vishvananda/netlink/issues/329).

According to [comment](https://github.com/vishvananda/netlink/issues/329#issuecomment-373133170), broadcast address auto-calculation is now a feature, and there is no intention to revert that. However, sometimes users do not want broadcast to be enabled.

Currently this can be done by setting `Broadcast` to `net.IPv4zero`. But that's not a public api of the netlink. And in my test, it will cause an error to be generated in the kernel log like `attribute type 4 has an invalid length.`.

I think we can use this as a special value to disable broadcast since it already works. But now we explicitly set it to nil when interacting with netlink.

This is an alternative to introducing a new function or extending the `Addr` struct.

### DIFF
--- a/addr_linux.go
+++ b/addr_linux.go
@@ -18,6 +18,7 @@ import (
 //
 // If `addr` is an IPv4 address and the broadcast address is not given, it
 // will be automatically computed based on the IP mask if /30 or larger.
+// If `net.IPv4zero` is given as the broadcast address, broadcast is disabled.
 func AddrAdd(link Link, addr *Addr) error {
 	return pkgHandle.AddrAdd(link, addr)
 }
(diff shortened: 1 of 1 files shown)


WeidiDeng:
@aboch Do you have time to take a look at this change? What's your opinion?

addr_linux.go
```diff
@@ -69,6 +74,7 @@ func AddrDel(link Link, addr *Addr) error {
 //
 // If `addr` is an IPv4 address and the broadcast address is not given, it
 // will be automatically computed based on the IP mask if /30 or larger.
+// If `net.IPv4zero` is given as the broadcast address, broadcast is disabled.
```
aboch, line 77: Does it make sense to specify this for a delete address operation?

aboch:
Thank you @WeidiDeng.
Deviating from iproute2 behavior is not good. We have chosen semantic versioning for this package to be able to introduce API breaking changes.

Let's do the right change, let's make the behavior of the addr add/del methods in line with iproute2 behavior.
Then we can release a 2.0 version,
WDYT?

addr_linux.go
WeidiDeng: No, it doesn't. I only added it for consistency with the other comments. It's added in this [commit](https://github.com/vishvananda/netlink/commit/aad0baef2856c220acdcb053b1d3d4391d8afd7b).

WeidiDeng: Maybe I should remove the part about broadcast auto computation as well.

WeidiDeng:
I think it's OK.

aboch:
Please squash the commits into one

WeidiDeng:
@aboch all commits are squashed

aboch:
LGTM

PR #1038, opened 2024-12-12
link_linux: Add deserialization of `IFF_RUNNING` flag

dylandreimerink:
I noticed that the decoded link flags never include `net.FlagRunning`. It seems deserialization logic was never added for this particular flag. Having access to the running flag is actually useful to detect when a link is in a NO-CARRIER state, which happens when a layer 1 issue occurs such as an unplugged or broken cable.

I also removed the comment since the function now deviates from the original.

### DIFF
--- a/link_linux.go
+++ b/link_linux.go
@@ -3041,7 +3041,6 @@ func parseMacvlanData(link Link, data []syscall.NetlinkRouteAttr) {
 	}
 }
 
-// copied from pkg/net_linux.go
 func linkFlags(rawFlags uint32) net.Flags {
 	var f net.Flags
 	if rawFlags&unix.IFF_UP != 0 {
@@ -3058,6 +3057,9 @@ func linkFlags(rawFlags uint32) net.Flags {
 	}
 	if rawFlags&unix.IFF_MULTICAST != 0 {
 		f |= net.FlagMulticast
+	}
+	if rawFlags&unix.IFF_RUNNING != 0 {
+		f |= net.FlagRunning
 	}
 	return f
 }


aboch:
Thank you @dylandreimerink 
Please check the test failure is related
image

dylandreimerink:
Hmm, it seems `net.FlagRunning` was [added](https://go-review.googlesource.com/c/go/+/413454) in go v1.20. But CI run go v1.17. Go v1.21 is currently the latest supported version according to the go [release policy](https://go.dev/doc/devel/release#policy), so would it be reasonable to bump the minimum go version?

dylandreimerink:
I bumped the Go version, which seems to have fixed the original failure. But now we have a new one:

```
=== RUN   TestRuleListFiltered/IPv6/returns_rules_filtered_by_fwmark_0/0xFFFFFFFF#01
    rule_test.go:592: Expected len: 1, got: 0

        --- PASS: TestRuleListFiltered/IPv6/returns_rules_filtered_by_fwmark_0/0xFFFFFFFF (0.00s)
        --- PASS: TestRuleListFiltered/IPv6/returns_rules_filtered_by_fwmark_0x1234/0 (0.00s)
        --- FAIL: TestRuleListFiltered/IPv6/returns_rules_filtered_by_fwmark_0/0xFFFFFFFF#01 (0.00s)
```

Interestingly the suite contains the exact same test twice (hence the #01 at the end of the last sub-test name). Both the name, and the test case is the same: [First](https://github.com/vishvananda/netlink/blob/main/rule_test.go#L387-L408) and [Second](https://github.com/vishvananda/netlink/blob/main/rule_test.go#L431-L452). But it is only on the second that fails. 

Not sure if this is test flakyness or something the Go bump might have caused :thinking: .

borkmann (approved):

aboch:
Thanks @dylandreimerink 
The `#01` suffix on those tests is there in older runs as well.

Can you please push a separate PR to bump go to 1.22. This project follows a 1 fix/change/feature per PR.

Thanks

dylandreimerink:
> Can you please push a separate PR to bump go to 1.22. This project follows a 1 fix/change/feature per PR.

Understandable. I will make a separate PR for the go version bump and rebase this PR once that one is in.

aboch:
please rebase to latest

aboch:
LGTM

PR #1039, opened 2024-12-15
Add support for ARP/ND Timestamps when retriving neighbors

jlamanna:
On Linux, Netlink provides NDA_CACHEINFO which carries timestamps about when ARP/ND was updated, used, and confirmed.

``` 
struct nda_cacheinfo {
    __u32         ndm_confirmed;
    __u32         ndm_used;
    __u32         ndm_updated;
    __u32         ndm_refcnt;
};
```

Expose these fields in the Neigh type

### DIFF
--- a/neigh.go
+++ b/neigh.go
@@ -19,6 +19,14 @@ type Neigh struct {
 	Vlan         int
 	VNI          int
 	MasterIndex  int
+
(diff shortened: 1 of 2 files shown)


aboch:
LGTM

PR #1048, opened 2025-01-13
fix various linting issues

thaJeztah:
More to come, but these were fairly trivial fixes.


conntrack_linux.go
```diff
@@ -641,9 +637,10 @@ func parseRawData(data []byte) *ConntrackFlow {
 		if nested, t, l := parseNfAttrTL(reader); nested {
 			switch t {
 			case nl.CTA_TUPLE_ORIG:
-				if nested, t, l = parseNfAttrTL(reader); nested && t == nl.CTA_TUPLE_IP {
+				if nested, t, _ = parseNfAttrTL(reader); nested && t == nl.CTA_TUPLE_IP {
 					parseIpTuple(reader, &s.Forward)
 				}
+				// TODO(thaJeztah): should this skipNfAttrValue ?
```
adrianchiris, line 643: i believe you can remove the TODO, setting wildcard is ok here since l (len) is not used here.

neigh_linux.go
```diff
@@ -120,12 +120,6 @@ func (h *Handle) NeighAppend(neigh *Neigh) error {
 	return h.neighAdd(neigh, unix.NLM_F_CREATE|unix.NLM_F_APPEND)
 }
 
-// NeighAppend will append an entry to FDB
-// Equivalent to: `bridge fdb append...`
-func neighAdd(neigh *Neigh, mode int) error {
```
adrianchiris, line 125: seems like this should be "NeighAdd" and not removed :)

adrianchiris:
minor nits otherwise lgtm

neigh_linux.go
thaJeztah: Hm... good call; possibly? Possibly someone forgot to export it?

thaJeztah: Oh, nope, that's not it; the exported `NeighAdd` is already there;
https://github.com/vishvananda/netlink/blob/cb48698f2590ca1a31b4ec83bc6598f3aa246aed/neigh_linux.go#L87-L91

What happend is that;

- the original implementation used the non-exported `neighAdd` to back various other functions (including the exported `NeighAdd`); https://github.com/vishvananda/netlink/commit/33e87186732a591e0fe542b3b5ab148088315309#diff-f4d304a4036276093514cde73db6556eb0e5975195683a892e3061302db9d090
- a later PR changed that to provide a handle, no longer using the non-exported `neighAdd` function; https://github.com/vishvananda/netlink/commit/14f41c27fae046385d2f8fcee41fd8e876ffd009
- which later got replaced https://github.com/vishvananda/netlink/commit/f116a3048ac2560afd2cec21c82bafa10762578c

So it looks like a leftover from the second step, and it's indeed no longer used.

conntrack_linux.go
thaJeztah: removed

thaJeztah:
> also please squash to a single commit

I rebased and removed the TODO, but didn't squash; if possible I'd rather keep the commits separate as I tried to group changes in logical commits, which can provide much more context when looking at "blame" view than a giant squashed commit without context for the specific change. Let me know if that's a problem.

adrianchiris:
The guidelines are one commit per PR, else its separate PRs.

all commits in this pr are covering the same area (lint fixes) so IMO its fine to squash.

thaJeztah:
Hm, is that documented somewhere? I'm not sure I agree with that; yes, PR should be an atomic / related set of changes, but requiring that to be a single commit would enforce larger changes to either be a long range of stacked PRs or bad practice and a commit that's too large to review. Commits should be logically grouped (and each commit should build), but I don't agree that a PR should always have a single commit.

neigh_linux.go
adrianchiris: ack, sounds legit.

PR #1051, opened 2025-01-20
Add IFLA_PARENT_DEV_NAME / IFLA_PARENT_DEV_BUS_NAME to links

akerouanton:
These attributes are supported since kernel v5.14 (see [1]). Here's what iproute2 shows:

```
$ ip -d link show eth0
4: eth0: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 65535 qdisc pfifo_fast state UP mode DEFAULT group default qlen 1000
    ... parentbus virtio parentdev virtio0
```

[1]: https://github.com/torvalds/linux/commit/00e77ed8e64d5f271c1f015c7153545980d48a76

### DIFF
--- a/link.go
+++ b/link.go
@@ -56,6 +56,8 @@ type LinkAttrs struct {
 	Vfs            []VfInfo // virtual functions available on link
 	Group          uint32
 	PermHWAddr     net.HardwareAddr
+	ParentDev      string
+	ParentDevBus   string
 	Slave          LinkSlave
 }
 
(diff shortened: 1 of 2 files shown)


thaJeztah (approved):
LGTM

aboch:
Thanks @akerouanton 
Would it be possible to add a new or extend and existing UT function to cover these changes?

akerouanton:
@aboch Testing this one is quite challenging because you need a physical interface to have a 'parent' reported, and there's no way to guarantee that the same parent will be available across CI, developers' machine, etc… or even if there'll be at least one interface with a 'parent'.

I'd need to mock the netlink response to test that, but it seems no other tests are using mocks.

aboch:
LGTM

PR #1049, opened 2025-01-14
.github/workflows: Bump CI Go version to v1.22

dylandreimerink:
Update the Go version we test against to Go v1.22 which is currently the oldest version still receiving security updates. This PR is in preparation for #1038

PR #1052, opened 2025-01-21
go.mod: bump to Go 1.20

tklauser:
PR https://github.com/vishvananda/netlink/pull/1038 started using `net.FlagRunning` which was introduced in Go 1.20, see https://go.dev/doc/go1.20#netpkgnet. Bump the minimum required Go version in `go.mod` accordingly and run `go mod tidy`.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module github.com/vishvananda/netlink
 
-go 1.12
+go 1.20
 
 require (
 	github.com/vishvananda/netns v0.0.4
--- a/go.sum
+++ b/go.sum
@@ -1,5 +1,4 @@
 github.com/vishvananda/netns v0.0.4 h1:Oeaw1EM2JMxD51g9uhtC0D7erkIjgmj8+JZc26m1YX8=
 github.com/vishvananda/netns v0.0.4/go.mod h1:SpkAiCQRtJ6TvvxPnOSyH3BMl6unz3xZlaprSwhNNJM=
-golang.org/x/sys v0.2.0/go.mod h1:oPkhp1MJrh7nUepCBck5+mAzfO9JrbApNNgaTdGDITg=
 golang.org/x/sys v0.10.0 h1:SqMFp9UcQJZa+pmYuAKjd9xq1f0j5rLcDIk0mj4qAsA=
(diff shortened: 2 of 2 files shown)


thaJeztah:
I guess, if we wanted to keep compatibility with go < 1.20, the alternative would be to define `net.FlagRunning` in a `//go:build !go1.20` file? (it's only the `FlagRunning` const that's missing?

@aboch @akerouanton for https://github.com/vishvananda/netlink/pull/1038, do you think it would make sense to have a "pre-go1.20" file?

akerouanton:
PR https://github.com/vishvananda/netlink/pull/1049 bumped the version of Go used on GHA to v1.22, so I guess it'd make sense to bump the minimum Go version to v1.22.

If this looks too drastic to @aboch, we should probably downgrade the version used on GHA to keep it aligned with `go.mod`.

thaJeztah:
> PR https://github.com/vishvananda/netlink/pull/1049 bumped the version of Go used on GHA to v1.22, so I guess it'd make sense to bump the minimum Go version to v1.22.

No, I don't think we should aggressively update the minimum version, because we know this module is used in many projects (either as a direct, or indirect dependency). For such modules, we should try to keep the requirements as low as possible. Given, go1.12 as minimum is really low (but if nothing in this module requires anything more recent, then that's perfectly fine).

I think the correct approach (more so, because CI on this repository runs really fast) would be to change to using a matrix in GitHub actions) to test against;

- minimum version specified in go.mod (i.e., verify what we advertise as minimum required version to actually work)
- actively supported go versions (stable, oldstable  (latest + latest -1))

We _could_ test all versions in between (with CI being reasonably short), but that may be a bit much.

adrianchiris:
adding my 2-cent here.

I think we should align with Go's support policy [1] where they support N and N-1 of the go toolchain.

some examples from popular go libs that are updating their min go version [2][3][4]

[1] https://go.dev/doc/devel/release
[2] https://github.com/grpc/grpc-go/blob/master/go.mod
[3] https://github.com/kubernetes-sigs/controller-runtime/blob/main/go.mod
[4] https://github.com/gin-gonic/gin/blob/master/go.mod

PR #1058, opened 2025-02-05
conntrack: prevent potential memory leak

aroradaman:
Currently, the ConntrackDeleteFilters captures all flow entries it fails to delete and reports them as errors. This behavior can potentially lead to memory leaks in high-traffic systems, where thousands of conntrack flow entries are cleared in a single batch. With this commit, instead of returning all the un-deleted flow entries, we now return a single error message for all of them.

ref: https://github.com/kubernetes/kubernetes/issues/129982

### DIFF
--- a/conntrack_linux.go
+++ b/conntrack_linux.go
@@ -5,8 +5,8 @@ import (
 	"encoding/binary"
 	"errors"
 	"fmt"
+	"io/fs"
 	"net"
-	"strings"
 	"time"
 
 	"github.com/vishvananda/netlink/nl"
@@ -159,7 +159,7 @@ func (h *Handle) ConntrackDeleteFilter(table ConntrackTableType, family InetFami
 // ConntrackDeleteFilters deletes entries on the specified table matching any of the specified filters using the netlink handle passed
 // conntrack -D [table] parameters         Delete conntrack or expectation
 func (h *Handle) ConntrackDeleteFilters(table ConntrackTableType, family InetFamily, filters ...CustomConntrackFilter) (uint, error) {
-	var errMsgs []string
+	var finalErr error
 	res, err := h.dumpConntrackTable(table, family)
 	if err != nil {
 		if !errors.Is(err, ErrDumpInterrupted) {
@@ -167,9 +167,10 @@ func (h *Handle) ConntrackDeleteFilters(table ConntrackTableType, family InetFam
 		}
 		// This allows us to at least do a best effort to try to clean the
 		// entries matching the filter.
-		errMsgs = append(errMsgs, err.Error())
+		finalErr = err
 	}
 
+	var totalFilterErrors int
 	var matched uint
 	for _, dataRaw := range res {
 		flow := parseRawData(dataRaw)
@@ -178,19 +179,20 @@ func (h *Handle) ConntrackDeleteFilters(table ConntrackTableType, family InetFam
 				req2 := h.newConntrackRequest(table, family, nl.IPCTNL_MSG_CT_DELETE, unix.NLM_F_ACK)
 				// skip the first 4 byte that are the netfilter header, the newConntrackRequest is adding it already
 				req2.AddRawData(dataRaw[4:])
-				if _, err = req2.Execute(unix.NETLINK_NETFILTER, 0); err == nil {
+				if _, err = req2.Execute(unix.NETLINK_NETFILTER, 0); err == nil || errors.Is(err, fs.ErrNotExist) {
 					matched++
 					// flow is already deleted, no need to match on other filters and continue to the next flow.
 					break
+				} else {
+					totalFilterErrors++
 				}
-				errMsgs = append(errMsgs, fmt.Sprintf("failed to delete conntrack flow '%s': %s", flow.String(), err.Error()))
 			}
 		}
 	}
-	if len(errMsgs) > 0 {
-		return matched, fmt.Errorf(strings.Join(errMsgs, "; "))
+	if totalFilterErrors > 0 {
+		finalErr = errors.Join(finalErr, fmt.Errorf("failed to delete %d conntrack flows with %d filters", totalFilterErrors, len(filters)))
 	}
-	return matched, nil
+	return matched, finalErr
 }
 
 func (h *Handle) newConntrackRequest(table ConntrackTableType, family InetFamily, operation, flags int) *nl.NetlinkRequest {


conntrack_linux.go
```diff
@@ -159,15 +158,17 @@ func (h *Handle) ConntrackDeleteFilter(table ConntrackTableType, family InetFami
 // ConntrackDeleteFilters deletes entries on the specified table matching any of the specified filters using the netlink handle passed
 // conntrack -D [table] parameters         Delete conntrack or expectation
 func (h *Handle) ConntrackDeleteFilters(table ConntrackTableType, family InetFamily, filters ...CustomConntrackFilter) (uint, error) {
-	var errMsgs []string
```
aojea, line 162: so the problem is that we have two nested loops

```
for _, dataRaw := range res {
 ...
		for _, filter := range filters {
```

that makes this array can be in the worse case (num_conntrack_entries x num_filters) and there is no point on returning that granularity, so some ideas
1. we report the total number of errors per filter, as filter will never be very large
2. we report just the total number of errors

Independently, we need to avoid the array and also wrap the ErrDumpInterrupted error so consumers can cast it

@aboch @aroradaman WDYT about something like

```diff
diff --git a/conntrack_linux.go b/conntrack_linux.go
index c566b2a..c09b65e 100644
--- a/conntrack_linux.go
+++ b/conntrack_linux.go
@@ -6,7 +6,6 @@ import (
        "errors"
        "fmt"
        "net"
-       "strings"
        "time"
 
        "github.com/vishvananda/netlink/nl"
@@ -159,7 +158,7 @@ func (h *Handle) ConntrackDeleteFilter(table ConntrackTableType, family InetFami
 // ConntrackDeleteFilters deletes entries on the specified table matching any of the specified filters using the netlink handle passed
 // conntrack -D [table] parameters         Delete conntrack or expectation
 func (h *Handle) ConntrackDeleteFilters(table ConntrackTableType, family InetFamily, filters ...CustomConntrackFilter) (uint, error) {
-       var errMsgs []string
+       var finalErr error
        res, err := h.dumpConntrackTable(table, family)
        if err != nil {
                if !errors.Is(err, ErrDumpInterrupted) {
@@ -167,9 +166,10 @@ func (h *Handle) ConntrackDeleteFilters(table ConntrackTableType, family InetFam
                }
                // This allows us to at least do a best effort to try to clean the
                // entries matching the filter.
-               errMsgs = append(errMsgs, err.Error())
+               finalErr = err
        }
 
+       var totalFilterErrors int
        var matched uint
        for _, dataRaw := range res {
                flow := parseRawData(dataRaw)
@@ -182,15 +182,16 @@ func (h *Handle) ConntrackDeleteFilters(table ConntrackTableType, family InetFam
                                        matched++
                                        // flow is already deleted, no need to match on other filters and continue to the next flow.
                                        break
+                               } else {
+                                       totalFilterErrors++
                                }
-                               errMsgs = append(errMsgs, fmt.Sprintf("failed to delete conntrack flow '%s': %s", flow.String(), err.Error()))
                        }
                }
        }
-       if len(errMsgs) > 0 {
-               return matched, fmt.Errorf(strings.Join(errMsgs, "; "))
+       if totalFilterErrors > 0 {
+               finalErr = errors.Join(finalErr, fmt.Errorf("failed to delete %d conntrack flows with %d filters", totalFilterErrors, len(filters)))
        }
-       return matched, nil
+       return matched, finalErr
 }
 ```

Comment (aroradaman) [reply conntrack_linux.go:162]: Make sense to pass the number of entries which failed to delete, I have updated the code with the same.

File: conntrack_linux.go [hunk 2]
```diff
@@ -182,15 +182,16 @@ func (h *Handle) ConntrackDeleteFilters(table ConntrackTableType, family InetFam
 					matched++
```
aojea, line 182:
```suggestion
                          if _, err = req2.Execute(unix.NETLINK_NETFILTER, 0); err == nil || errors.Is(err, fs.ErrNotExist) {
					matched++
```

conntrack_linux.go
```diff
@@ -182,15 +182,16 @@ func (h *Handle) ConntrackDeleteFilters(table ConntrackTableType, family InetFam
...
 					// flow is already deleted, no need to match on other filters and continue to the next flow.
 					break
+				} else {
```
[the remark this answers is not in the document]
aojea: just the check for ENOENT and it LGTM

conntrack_linux.go
```diff
@@ -178,19 +179,20 @@ func (h *Handle) ConntrackDeleteFilters(table ConntrackTableType, family InetFam
 				req2 := h.newConntrackRequest(table, family, nl.IPCTNL_MSG_CT_DELETE, unix.NLM_F_ACK)
 				// skip the first 4 byte that are the netfilter header, the newConntrackRequest is adding it already
 				req2.AddRawData(dataRaw[4:])
-				if _, err = req2.Execute(unix.NETLINK_NETFILTER, 0); err == nil {
+				if _, err = req2.Execute(unix.NETLINK_NETFILTER, 0); err == nil || errors.Is(err, fs.ErrNotExist) {
```
aojea, line 182: fore reference, this is how golang std lib recommends to do this check https://pkg.go.dev/os#IsNotExist

aojea:
/lgtm

/assign @aboch 

PTAL this is causing disruption and a big impact on performance and scalability https://github.com/kubernetes/kubernetes/issues/129982

conntrack_linux.go
aroradaman: I compared the err with syscall.ENONET, fs.ErrNotExist and unix.ENONET. Only got a match on fs.ErrNotExist .

https://github.com/vishvananda/netlink/compare/main...aroradaman:netlink:debug?expand=1
```
####################################################### Netlink ConntrackDeleteFilters
error occurred no such file or directory false true false
length 1808537
####################################################### Netlink ConntrackDeleteFilters
```

aboch:
LGTM

aojea:
Thanks 🙇

conntrack_linux.go
```diff
@@ -178,19 +179,20 @@ func (h *Handle) ConntrackDeleteFilters(table ConntrackTableType, family InetFam
...
+					totalFilterErrors++
 				}
-				errMsgs = append(errMsgs, fmt.Sprintf("failed to delete conntrack flow '%s': %s", flow.String(), err.Error()))
 			}
 		}
 	}
-	if len(errMsgs) > 0 {
-		return matched, fmt.Errorf(strings.Join(errMsgs, "; "))
+	if totalFilterErrors > 0 {
+		finalErr = errors.Join(finalErr, fmt.Errorf("failed to delete %d conntrack flows with %d filters", totalFilterErrors, len(filters)))
```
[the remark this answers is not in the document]
thaJeztah: Yes, this breaks the module; 
https://github.com/vishvananda/netlink/blob/655392bc778acd41a35de75f804fe7088e922e2b/go.mod#L3


For this one we may be able to just use the old approach (at least, I _think_ it would give mostly the same for now)
 (there was another change also requiring newer go versions)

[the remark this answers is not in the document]
aojea: yeah, but 1.20 was EOL one year ago https://github.com/devcontainers/images/issues/944 ,

PR #1060, opened 2025-02-18
Add `OifIndex` option for `RouteGetWithOptions`

dylandreimerink:
The `RouteGetWithOptions` function currently has a `Oif` option which gets translated from link name to link index via a `LinkByName` call. This adds unnecessary overhead when the link index is already known.

This commit adds a new `OifIndex` option to `RouteGetWithOptions` which can be specified instead of `Oif` to skip the internal link index translation.

### DIFF
--- a/route_linux.go
+++ b/route_linux.go
@@ -1533,6 +1533,7 @@ type RouteGetOptions struct {
 	Iif      string
 	IifIndex int
 	Oif      string
+	OifIndex int
 	VrfName  string
 	SrcAddr  net.IP
 	UID      *uint32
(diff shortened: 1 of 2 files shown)


borkmann (approved):

dylandreimerink:
cc @aboch

route_linux.go
```diff
@@ -1533,6 +1533,7 @@ type RouteGetOptions struct {
 	Iif      string
 	IifIndex int
 	Oif      string
+	OifIndex int
```
aboch, line 1536: I was wondering why not making this `uint32` then I see LifIndex is already there as `int`.

LGTM

PR #1062, opened 2025-02-21
vxlan: Fix parseVxlanData for source port range

borkmann:
binary.Read() != nil check means error case, so the vxlan.Port{Low,High} are never populated. Fix the check.


borkmann:
Cc @aboch

aboch:
LGTM

PR #1067, opened 2025-03-26
Add support for MTU Lock

trozet:
When adding a route with "mtu lock <mtu>" path MTU discovery (PMTUD) will not be tried and packets will be sent without DF bit set. Upon receiving an ICMP needs frag due to PMTUD, the kernel will not install a cached route and lower the MTU.


aboch:
LGTM

PR #1077, opened 2025-04-25
geneve: Support setting/getting source port range

borkmann:
Add support for geneve feature to specify source port range, see kernel commits:

- e1f95b1992b8 ("geneve: Allow users to specify source port range")
- 5a41a00cd5d5 ("geneve, specs: Add port range to rt_link specification")

This is exactly equivalent on what is done in case of vxlan today.


borkmann:
Cc @tklauser @aboch ptal

tklauser (approved):
Thanks

aboch:
LGTM

PR #1078, opened 2025-04-27
add support for flags and qos maps of vlan links

gwenya:
Adds all the vlan flags and the ingress and egress qos maps.

The flags are pointers to bools so that they can be modified individually without affecting other flags.


gwenya:
I don't quite understand which test is failing there

PR #1083, opened 2025-05-08
qdisc: fix wrong type info of tc_sfq_qopt

lwintermelon:
see https://github.com/vishvananda/netlink/blob/9d88d8385bf99155c1e66be2877126518ef85282/nl/tc_linux.go#L1132-L1146


aboch:
LGTM

PR #1088, opened 2025-06-16
Add SRv6 support for End.DT46

wnagele:
Simple one to add the End.DT46 SRv6 action.


wnagele:
Looks like the Linux kernel added some restrictions in a newer release which breaks this test case. Relevant commit:
https://github.com/torvalds/linux/commit/1ddf9916ac09313128e40d6581cef889c0b4ce84#diff-a96154251770a71528224b271a6ef513db473b3e82af7070ca56031f5a316ce7R464

I am not familiar enough with the code base to say if this test case should be adjusted, removed, etc. Once fixed in another PR I am happy to rebase here.

aboch:
LGTM

Issue #1097, opened 2025-06-29
Add option to skip VF information fetching for performance optimization


pasteley:
### Summary

The netlink library currently always requests Virtual Function (VF) information when making `RTM_GETLINK` calls, even when this data is never used by the calling application. This causes significant performance degradation on systems with many VFs, particularly affecting applications like Cilium that make frequent netlink calls.

### Problem Description
**Current Behavior**

- All `RTM_GETLINK` calls in `link_linux.go` unconditionally set the `RTEXT_FILTER_VF` flag
- This triggers the kernel's `rtnl_fill_vfinfo()` function for every VF on the interface
- On systems with 64 VFs (common for SR-IOV), this results in 64 expensive kernel calls per netlink operation

### Performance Impact

We measured severe performance degradation in production environments:

**Problematic scenario:**

- 12 netlink calls → 768 VF information queries (64 VFs × 12 calls)
- Each VF query: 1-2ms
- Total time: ~866ms per operation
- 99.8% of application wait time spent on unused VF data fetching

**Healthy comparison:**

- Same hardware with optimized path: ~157ms total (5.5× improvement)
- Per-VF query: 128-256μs (8× faster per call)

PR #1098, opened 2025-06-29
feat: option to disable VF polling

pasteley:
Fixes: #1097

### DIFF
--- a/handle_linux.go
+++ b/handle_linux.go
@@ -12,13 +12,24 @@ import (
 // Empty handle used by the netlink package methods
 var pkgHandle = &Handle{}
 
-// Handle is an handle for the netlink requests on a
+type HandleOptions struct {
+	lookupByDump  bool
+	collectVFInfo bool
+}
+
+// Handle is a handle for the netlink requests on a
 // specific network namespace. All the requests on the
 // same netlink family share the same netlink socket,
 // which gets released when the handle is Close'd.
 type Handle struct {
-	sockets      map[int]*nl.SocketHandle
-	lookupByDump bool
+	sockets map[int]*nl.SocketHandle
+	options HandleOptions
+}
+
+// DisableVFInfoCollection configures the handle to skip VF information fetching
+func (h *Handle) DisableVFInfoCollection() *Handle {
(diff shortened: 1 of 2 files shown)


handle_linux.go
```diff
@@ -12,13 +12,26 @@ import (
 // Empty handle used by the netlink package methods
 var pkgHandle = &Handle{}
 
-// Handle is an handle for the netlink requests on a
+type HandleOptions struct {
+	lookupByDump bool
+	skipVFInfo   bool
```
aboch, line 17: Can you make this `enableVFInfo` so that it's off by default?

handle_linux.go
```diff
@@ -12,13 +12,26 @@ import (
...
 // which gets released when the handle is Close'd.
 type Handle struct {
-	sockets      map[int]*nl.SocketHandle
-	lookupByDump bool
+	sockets map[int]*nl.SocketHandle
+	options HandleOptions
+}
+
+// DisableVFInfo configures the handle to skip VF information fetching
+func (h *Handle) DisableVFInfo() *Handle {
```
aboch, line 32: EnableVFInfo() instead
and can you please also add a function to allow enabling lookupByDump ?

aboch: SGTM

aboch:
LGTM

handle_linux.go
pasteley: There isn’t really a single “proper” approach here—it really depends on the specific scenario and use case. Since this is a common library, there may already be users who rely on the current behavior. If it’s possible to avoid breaking changes, I think it would be best to do so. This way, we can maintain compatibility and avoid disrupting existing projects. Thanks for considering this!

lazysegtree:
@pasteley  Fix for TestXfrmStateWithPcpunumWithoutSADir, the test which is failing in your PR, was recently merged in main. 
Regarding other test that is failing, its a known issue. Should pass after retries
- https://github.com/vishvananda/netlink/issues/1108

Please rebase with main.

lazysegtree:
@aboch Anything else needed to merge ?

PR #1101, opened 2025-07-14
Add DevlinkSplitPort and DevlinkUnsplitPort

Charlie999:
Adds functions for splitting and un-splitting ports (for use on switches).


Charlie999:
Rebased

vishvananda:
The flaky test was fixed. If you can rebase and squash your changes into a single commit that would be great. Also is it possible to add a simple test for this?

Charlie999:
Rebased and squashed.

As for tests, I'm not sure it's possible to come up with a decent test without an actual switch chip.

I can confirm this patch works as we are using it on our dev SN2700s with no issue at Mythic Beasts.

Issue #1069, opened 2025-03-27
undefined socket functions for non-Linux platforms causing issues


zajdee:
Hello,

while trying to run a `golang-ci` linter on a macOS against an app using `netlink.SocketDiagTCPInfo` (added in #534), I'm getting a lint error that cannot be skipped:
```
undefined: netlink.SocketDiagTCPInfo (typecheck)
```

It seems that while many other functions have platform-agnostic code that returns `ErrNotImplemented`, this is not the case for many of the functions defined in `socket_linux.go` and `socket_xdp_linux.go`, except `SocketGet` and `SocketDestroy` (which live in `netlink_unspecified.go`).

To ease the linting, it would be useful to have the platform-agnostic socket functions defined too.

PR #1070, opened 2025-03-27
socket: add functions for unspecified platforms

zajdee:
This PR adds socket functions for unspecified (non-Linux) platforms and fixes #1069.

PR #1107, opened 2025-07-20
socket: add functions for unspecified platforms

zajdee:
This PR adds socket functions for unspecified (non-Linux) platforms and fixes https://github.com/vishvananda/netlink/issues/1069.


zajdee:
This is a follow up to the discussion in #1070, where Github just closed the PR for me.
The change is squashed into a single commit.
Note that the tests are also failing in the `main` branch, so the test failure doesn't seem to be caused by this change.

lazysegtree:
@zajdee Tests will be fixed by this 
- https://github.com/vishvananda/netlink/pull/1106

Yes, they are unrelated to your change, and doesn't block merge.

PR #1109, opened 2025-07-22
Clean up logging

tklauser:
Don't log from the library on (yet) unknown fou or ipset attributes.These may appear any time a new kernel is used which adds new attributes.

Also clean up some stray logging in tests.

See commits for details.


tklauser:
cc @aboch

aboch:
LGTM

PR #1110, opened 2025-07-22
Skip TestSocketXDPGetInfo on EPERM

tklauser:
Skip `TestSocketXDPGetInfo` rather than failing it when the user lacks permissions to create AF_XDP sockets.

This allows to run the full test suite as an unprivileged user again.

### DIFF
--- a/socket_xdp_linux_test.go
+++ b/socket_xdp_linux_test.go
@@ -4,6 +4,7 @@
 package netlink
 
 import (
+	"errors"
 	"os"
 	"testing"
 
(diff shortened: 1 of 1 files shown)


socket_xdp_linux_test.go
```diff
@@ -12,7 +13,9 @@ import (
 
 func TestSocketXDPGetInfo(t *testing.T) {
 	xdpsockfd, err := unix.Socket(unix.AF_XDP, unix.SOCK_RAW, 0)
-	if err != nil {
+	if errors.Is(err, unix.EPERM) {
```
aboch, line 16: I understand avoiding extra indentation, but I feel it is more readable is we test the error type only if the error is not nil. I am sure `error.Is` test handles the _nil_ use case, but I do not want to have to check.
```
if err != nil {
    if errors.Is(err, unix.EPERM) {
        t.Skipf("creating AF_XDP socket not permitted")
    }
     t.Fatal(err)
}
```

Yes, this is only test code, but quite often test code is looked upon to find examples.

tklauser:
The test failure looks unrelated. It seems `TestRuleListFiltered` is somewhat flaky.

aboch:
LGTM

PR #1112, opened 2025-08-01
Fixed IP6tnl encapsulation and ID handling issues:

vladimir131313:
- Added IP6tnlEncap processing
- Added proper IP6tnlEncap.ID assignment
- Enabled handling of IP6tnlEncap.TC and IP6tnlEncap.FLAGS for serialization support

In the current version, IP6tnlEncap.ID always applys 0 regardless of the actual e.ID value.
Route serialization with IP6tnlEncap doesn't work - the encap field always becomes nil.
For ID handling, binary.BigEndian is used since the kernel explicitly specifies BigEndian type when parsing this parameter: https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/net/ipv4/ip_tunnel_core.c?h=v6.12#n695

### DIFF
--- a/route_linux.go
+++ b/route_linux.go
@@ -669,7 +669,7 @@ func (e *IP6tnlEncap) Decode(buf []byte) error {
 	for _, attr := range attrs {
 		switch attr.Attr.Type {
 		case nl.LWTUNNEL_IP6_ID:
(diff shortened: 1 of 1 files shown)


aboch:
LGTM

PR #1115, opened 2025-08-17
feat: add additional devlink port attributes

adrianchiris:
Add the following devlink port attributes:

PortNumber: the physical port number
PfNumber: the PF number
VfNumber: the VF number (index)
SfNumber: the SF number (index)
ControllerNumber: the controller number
External: if set, indicates external controller

iproute2 ref: https://github.com/iproute2/iproute2/blob/0ad8fef322365b7bafd052f416fc972bea49d362/devlink/devlink.c#L4949

## Summary by CodeRabbit

- New Features
  - Expanded devlink port metadata: devices may now surface additional optional port identifiers (port/PF/VF/SF numbers), controller number, and external status when available, without changing existing behavior.

- Tests
  - Added comprehensive parsing tests covering physical, PF, VF, SF, and controller variants.
  - Improved test logs for clearer, structured port output.

- Chores
  - Added a testing assertion library dependency.


devlink_test.go
```diff
@@ -57,7 +114,7 @@ func TestDevLinkGetAllPortList(t *testing.T) {
 	}
 	t.Log("devlink port count = ", len(ports))
 	for _, port := range ports {
-		t.Log(*port)
+		logPort(t, port)
```
adrianchiris, line 117: as we have pointers in the struct, previous logging method wasnt ideal

vishvananda:
This looks good although I am a bit concerned about pulling in testify and all of the relevant deps for just handling the asserts. I accidentally pushed to your branch but i'll pr a fix in instead so you can squash it in.

vishvananda:
I don't feel super strongly about the testify dep although if we are going to include it, it might be nice go go through and switch all of the existing tests to use it. LGTM for now.

adrianchiris:
>I don't feel super strongly about the testify dep although if we are going to include it, it might be nice go go through and switch all of the existing tests to use it. LGTM for now.

ack, i will followup with moving the remaining devlink tests. (one step at a time :) )

PR #1119, opened 2025-08-26
test: Improve test reliability with proper cleanup and isolation

vishvananda:
Refactors test setup and teardown logic to use `t.Cleanup` instead of `defer`. This ensures that cleanup functions are correctly scoped to each subtest's lifecycle, improving test isolation and reliability.

The `setUpNetlinkTest` helper function is also improved to correctly save and restore the original network namespace, ensuring that tests do not leak state.

Additionally, a bug in the `netns` tests is fixed where a large namespace ID could overflow a 32-bit integer, causing spurious failures on some systems.

depends on #1118


netns_test.go
```diff
@@ -25,7 +26,7 @@ func TestNetNsIdByFd(t *testing.T) {
 	// In an attempt to avoid namespace id collisions, set this to something
 	// insanely high. When the kernel assigns IDs, it does so starting from 0
 	// So, just use our pid shifted up 16 bits
-	wantID := os.Getpid() << 16
+	wantID := (os.Getpid() << 16) & 0x7FFFFFFF
```
lazysegtree, line 29: @coderabbitai

(1) What are the usual values that os.Getpid will return ? Is it always less than 2^16 ?
(2) Why do we need do and it by 0x7FFFFFFF . I guess it is because if the PID had its 16th bit set, then `<<16` will overflow to negative integer . Is that correct ?

PR #1120, opened 2025-08-27
test(conntrack): ensure conntrack hooks for tests

vishvananda:
This change ensures that the necessary netfilter hooks are in place for conntrack tests to run reliably. Previously, the tests would fail in environments where the host's firewall was not configured to accept conntrack traffic.

This change introduces a new function, `ensureCtHooksInThisNS`, that uses `iptables` or `nftables` to install the necessary hooks. This function is called from `nsCreateAndEnter`, so all tests that use this function will have a properly configured netns.

This change also removes the `CI` environment variable check from the tests, as they are now expected to pass in CI environments.

## Summary by CodeRabbit

* **Tests**
  * Centralized setup of connection-tracking hooks within test network namespaces for consistent behavior.
  * Enabled relevant connection-tracking settings during namespace setup to reduce flakiness.
  * Runtime detection and use of available firewall tooling (iptables or nft), with graceful skipping when unsupported.
  * Removed scattered per-test conntrack configs and CI-specific skips in favor of a unified, idempotent namespace-scoped setup.

### DIFF
--- a/conntrack_test.go
+++ b/conntrack_test.go
@@ -8,6 +8,7 @@ import (
 	"fmt"
 	"net"
 	"os"
+	"os/exec"
 	"runtime"
 	"testing"
 	"time"
@@ -29,7 +30,7 @@ func CheckError(t *testing.T, err error) {
 }
 
 func udpFlowCreateProg(t *testing.T, flows, srcPort int, dstIP string, dstPort int) {
-	for i := 0; i < flows; i++ {
+	for i := range flows {
 		ServerAddr, err := net.ResolveUDPAddr("udp", fmt.Sprintf("%s:%d", dstIP, dstPort))
 		CheckError(t, err)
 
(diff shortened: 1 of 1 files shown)


conntrack_test.go
```diff
@@ -44,6 +45,57 @@ func udpFlowCreateProg(t *testing.T, flows, srcPort int, dstIP string, dstPort i
...
+		ipt := func(args ...string) {
+			cmd := exec.Command("iptables", args...)
+			if out, err := cmd.CombinedOutput(); err != nil {
+				t.Fatalf("iptables %v failed: %v\n%s", args, err, out)
+			}
+		}
+
+		// Minimal hooks so packets traverse conntrack in this netns.
+		// Check (-C); if absent, insert (-I). Idempotent on reruns.
+		if !iptExists("-C", "INPUT", "-m", "conntrack", "--ctstate", "NEW,ESTABLISHED", "-j", "ACCEPT") {
```
aboch, line 74: Given we are polluting the host with these rules, it would be better to add them to a custom iptables chain, and something that reconnects to this library UT
```
iptables -D INPUT -j vish_netlink
iptables -X vish_netlink
iptables -N vish_netlink
iptables -I vish_netlink -m conntrack --ctstate NEW, ESTABLISHED - j ACCEPT
```
and at test end we should cleanup
```
iptables -D INPUT -j vish_netlink
iptables -X vish_netlink
```

similarly for the nft table.

vishvananda: @aboch The rules will only be created in the network namespace used for the test so they will be gone when the namespace is dropped. I can add a cleanup step to remove the rules for safety but I’m not sure the value of using a separate table.

vishvananda: I modified the test to cleanup the rules it adds.

aboch: Oh, of course Sorry I missed the setup would only be executed inside a netns. Please disregard my comment.

aboch:
LGTM

PR #1121, opened 2025-08-27
tests: fix SEG6 route test on recent kernels

vishvananda:
The SEG6 route test was failing on newer kernels:

* Using loopback (`lo`) for SRv6 inline routes no longer preserves the `encap seg6` attribute. Switched to a dummy device (`eth0`) to ensure SRv6 routes are accepted.
* Removed the artificial `::` SID from the inline segment list, which caused the kernel to drop the SRH.
* Added cleanup to delete the dummy device after the test.
* Updated route checks to use `RouteListFiltered` with `RT_FILTER_OIF|RT_FILTER_DST` since modern kernels add extra auto-routes (e.g., fe80::/64), which broke the naive `len(routes) == 1` assertion.

These changes make `TestSEG6RouteAddDel` stable across recent kernels while still validating both INLINE (IPv6) and ENCAP (IPv4) SRv6 routes.

## Summary by CodeRabbit

- New Features
  - No user-facing changes in this release.

- Tests
  - Improved route-related tests for broader kernel compatibility by using a controlled dummy network interface.
  - Replaced environment-specific assumptions with explicit interface setup and cleanup.
  - Switched to filtered route discovery to avoid false positives and validated inline vs encap paths.
  - Added kernel-version guard for tests requiring newer kernel features and improved per-test cleanup messaging.

- Chores
  - Removed CI-specific skip logic and obsolete environment imports from tests.

### DIFF
--- a/route_test.go
+++ b/route_test.go
@@ -5,7 +5,6 @@ package netlink
 
 import (
 	"net"
-	"os"
 	"runtime"
 	"strconv"
 	"testing"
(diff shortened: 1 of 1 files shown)


route_test.go
```diff
@@ -1997,18 +1996,26 @@ func TestSEG6LocalEqual(t *testing.T) {
...
-	}
-	// add/del routes with LWTUNNEL_SEG6 to/from loopback interface.
+	// add/del routes with LWTUNNEL_SEG6 to/from interface.
 	// Test both seg6 modes: encap (IPv4) & inline (IPv6).
 	t.Cleanup(setUpSEG6NetlinkTest(t))
 
-	// get loopback interface and bring it up
-	link, err := LinkByName("lo")
+	// loopback doesn't work on recent kernels, so use a dummy
+	err := LinkAdd(&Dummy{LinkAttrs{Name: "eth0"}})
```
aboch, line 2004: Should we give the dummy device a name which is unlikely to conflict with a real device name on the host?
People can run go test on their linux VM where `eth0` may exist.

vishvananda: This is in a brand new isolated network namespace so there should be no conflict. I used eth0 because some of the other tests use that name.

PR #1122, opened 2025-08-27
tests: Stabilize TestLinkXdp against race condition

vishvananda:
The TestLinkXdp test was observed to be flaky in CI environments. The failure occurred because the test expects the second call to LinkSetXdpFdWithFlags with the XDP_FLAGS_UPDATE_IF_NOEXIST flag to fail with EBUSY, indicating that an XDP program is already attached.

Due to a race condition, the kernel had not always completed the attachment from the first LinkSetXdpFd call before the second call was made. This caused the second call to succeed (returning a nil error) when it should have failed. The test logic correctly interpreted this unexpected success as a failure, but logged the nil error, leading to confusing output.

This commit resolves the flakiness by introducing a retry loop with a short delay. The test now attempts the second call multiple times, giving the kernel sufficient time to update the link's XDP state. This ensures that the test reliably checks for the EBUSY error as intended.

## Summary by CodeRabbit

- Tests
  - Improved reliability of XDP/link attachment tests by adding brief retries to handle asynchronous kernel state updates.
  - Added additional teardown cleanup to ensure resources are released.
  - Tests now assert expected busy-state outcomes after retries to reduce flakiness on slower environments.
  - Clarifying comments added; no public APIs or runtime behaviors changed.

### DIFF
--- a/link_test.go
+++ b/link_test.go
@@ -2591,14 +2591,32 @@ func TestLinkXdp(t *testing.T) {
 	if err != nil {
 		t.Skipf("Loading bpf program failed: %s", err)
 	}
+	t.Cleanup(func() {
+		_ = unix.Close(fd)
+	})
+
 	if err := LinkSetXdpFd(testXdpLink, fd); err != nil {
 		t.Fatal(err)
 	}
-	if err := LinkSetXdpFdWithFlags(testXdpLink, fd, nl.XDP_FLAGS_UPDATE_IF_NOEXIST); !errors.Is(err, unix.EBUSY) {
-		t.Fatal(err)
+	t.Cleanup(func() {
+		_ = LinkSetXdpFd(testXdpLink, -1)
+	})
+	var err2 error
+	// It can take a moment for the kernel to update the link state, so we retry here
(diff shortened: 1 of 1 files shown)


link_test.go
```diff
@@ -2591,14 +2591,32 @@ func TestLinkXdp(t *testing.T) {
...
 	}
-	if err := LinkSetXdpFdWithFlags(testXdpLink, fd, nl.XDP_FLAGS_UPDATE_IF_NOEXIST); !errors.Is(err, unix.EBUSY) {
-		t.Fatal(err)
+	t.Cleanup(func() {
+		_ = LinkSetXdpFd(testXdpLink, -1)
+	})
+	var err2 error
+	// It can take a moment for the kernel to update the link state, so we retry here
+	attempts := 0
+	for attempts = 0; attempts < 10; attempts++ {
```
aboch, line 2607: Can you please `for ; attempts < 10; attempts++ {`

vishvananda: done

aboch:
LGTM

Issue #755, opened 2022-04-18
LinkByName crashing on v1.2.0-beta but works in master


DavidVentura:
Hi
Commit 5e915e0149386ce3d02379ff93f4c0a5601779d5 fixes a crash I was having on ARMv7 when calling `LinkByName` -- `SetExtAck` would crash with `protocol not available`

While this was probably a bug, should `EnableErrorMessageReporting` get documented more ?
Even if this is closed immediately (there's no bug in HEAD after all), this might help others with the same issues.

vishvananda:
I'll make a new beta tag that includes the fix from here:https://github.com/vishvananda/netlink/pull/752

vishvananda:
New release posted: https://github.com/vishvananda/netlink/releases/tag/v1.2.1-beta.2

dreamcat4:
could be related to this bug? https://github.com/moby/libnetwork/issues/2614

PR #1130, opened 2025-09-24
Always check for `EnableErrorMessageReporting` and `SetExtAck()` when creating Netlink sockets.

### LINKED ISSUE #1129: `nl.EnableErrorMessageReporting = true` does not enable kernel error messages.
It loos like `nl.EnableErrorMessageReporting` is designed to enable the Netlink's `EXT_ACK` part of the protocol, which provides readable English kernel error messages in case of error back to the user. Kernel documentation highly recommends to have it enabled,

from https://www.kernel.org/doc/html/v6.6/userspace-api/netlink/intro.html#extended-ack

> Extended ACKs greatly improve the usability of Netlink and should always be enabled, appropriately parsed and reported to the user.

To rephrase in terms of a classic bug report, considering code like this:

```go
func main() {
	nl.EnableErrorMessageReporting = true
        err := installFilter() // error: no qdisc installed previously to place the filter, err is `netlink` error.
        if err != nil {
            panic(err)
        }

}


```

One expects to observe a message

```
panic: invalid argument: Parent Qdisc doesn't exists
```

But instead one gets

```
panic: invalid argument
```

printed, that is an error based solely on the integer error code.

We debugged this issue a bit and realized, that while all the support for these messages is in place, the ext acks are not always requested. I'll send a PR, let's discuss the details there.

peterk-akamai:
Please see https://github.com/vishvananda/netlink/issues/1129 with description of the issue from the user perspective.

In technical tems, the netlink library does great job of managing the ExtAcks, but the subtle issue (and arguably a bug) here is that not all code paths that create new Netlink sockets check and `SetExtAck()`.

One can also observe with strace, that indeed socket has no `NETLINK_EXT_ACK` option set and the extended error messages are not communicated on the protocol level.

This PR fixes this issue by always checking `EnableErrorMessageReporting` and setting the ext ack socket option when creating new sockets. We ignore the error here, but it's not ignored on the other code path (in `ExecuteIter`), maybe we should not ignore it  or maybe better ignore it in both places (i.e. if for example a binary is run on an old kernel with no support for ext acks, it still works  without readable errrors).

I also looked a bit into the alternatives: one reasonable one seems to be to add `Handle.SetExtAck()` and make the user to call it explicitly when creating handles. It seems like additional complexity and against the spirit of the kernel documentation, but still viable.

Please take a quick look at the PR.

fixes #1129.

### DIFF
--- a/nl/nl_linux.go
+++ b/nl/nl_linux.go
@@ -716,6 +716,11 @@ func getNetlinkSocket(protocol int) (*NetlinkSocket, error) {
 		return nil, err
 	}
 
+	if EnableErrorMessageReporting {
+		// ignore error
+		_ = s.SetExtAck(true)
+	}
+
 	return s, nil
 }
 


peterk-akamai:
I see comments by AI, the main one makes sense and I described it already

> Here we return on any SetExtAck error, but in getNetlinkSocket (Lines 719-723) we currently ignore errors

It would be nice to hear a human opinion before I go on with the change.

peterk-akamai:
See a real example of why it makes sense to ignore the error here: https://github.com/vishvananda/netlink/issues/755

aboch:
LGTM

PR #1133, opened 2025-10-05
Remove netlink subscription in socket_linux.go

aojea:
Removed netlink subscription that seems was a left over from the cleanup in https://github.com/vishvananda/netlink/commit/a57a7bd6b256ee7215ff5a934cf4c11d306547ab

## Summary by CodeRabbit

* **Bug Fixes**
  * Fewer rare errors when destroying sockets on Linux, improving reliability during teardown.

* **Refactor**
  * Simplified socket teardown flow to remove unnecessary subscription and cleanup steps, reducing complexity and potential failure points.

### DIFF
--- a/socket_linux.go
+++ b/socket_linux.go
@@ -261,11 +261,6 @@ func (h *Handle) SocketDestroy(local, remote net.Addr) error {
 		return ErrNotImplemented
 	}
 
-	s, err := nl.Subscribe(unix.NETLINK_INET_DIAG)
-	if err != nil {
-		return err
-	}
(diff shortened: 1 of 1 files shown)


aboch:
Wondering why we were subscribing during the destroy

aboch:
LGTM

aojea:
> Wondering why we were subscribing during the destroy

I do not know, but it also was a problem because IIUIC is not namespace aware, so it subscribe to the root namespace despite I was handling a different network namespace

PR #1140, opened 2025-11-10
nexthop: Support LWT encapsulation

YutaroHayakawa:
Add support for Light Weight Tunnel encapsulation in the nexthop object. Since the netlink encoding is exactly the same as the one for route, we can reuse most of the encoding/decoding code. What we needed to do was supporting NHA_ENCAP_TYPE and NHA_ENCAP types.

## Summary by CodeRabbit

* **New Features**
  * Added encapsulation support for nexthops with multiple encap types and ensured encapsulation is included when creating nexthops; route decoding now centralizes encap handling.
* **Bug Fixes**
  * Nexthop string output is now safe for nil receivers and shows encap presence or absence.
* **Tests**
  * Added tests for nexthop encapsulation and string handling to validate create/retrieve behavior.

### DIFF
--- a/nexthop.go
+++ b/nexthop.go
@@ -13,6 +13,7 @@ type Nexthop struct {
 	OIF       uint32
 	Gateway   net.IP
 	Protocol  RouteProtocol
+	Encap     Encap
 }
 
 func (h *Nexthop) String() string {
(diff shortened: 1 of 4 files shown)


YutaroHayakawa:
@aboch @vishvananda Please take a look.

aboch:
Thank you for adding the UT.

LGTM

PR #1145, opened 2025-11-20
fix: always set vxlan udpcsum attribute

### LINKED ISSUE #1144: Unable to set `noudpcsum` on vxlan device
There seem to be no way to create vxlan netdev with `noudpcsum` attribute. Adding a vxlan link with `UDPCSum: false` has no effect as the attribute is passed [only when its value is true](https://github.com/vishvananda/netlink/blob/main/link_linux.go#L1325-L1327). So the attribute is set to whatever is the default for OS, which happens to be `true` on Ubuntu 22.04 5.19.0-46-generic ([relevant patch](https://git.zx2c4.com/linux-rng/commit/include/net?id=6ceb31ca5f65acff299dbc3da5854d54e147b7d8)), making it impossible to opt out on link creation.

Unsetting this attribute for existing netdevs [isn't supported either](https://elixir.bootlin.com/linux/v5.19/source/drivers/net/vxlan/vxlan_core.c#L4040-L4048):
```
$ sudo ip link set vxlan1 type vxlan noudpcsum
Error: Cannot change UDP_CSUM flag.
```
so `LinkModify` also expectedly fails.

What would be the best way around this issue?

p-sherratt:
fixes #1144

we should always explicitly set this attribute for consistency between kernel versions, since the default udpcsum behaviour changes between versions.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved VXLAN network configuration to consistently communicate UDP checksum settings to the system, ensuring more reliable network tunnel behavior in all scenarios.

### DIFF
--- a/link_linux.go
+++ b/link_linux.go
@@ -1319,12 +1319,10 @@ func addVxlanAttrs(vxlan *Vxlan, linkInfo *nl.RtAttr) {
 	data.AddRtAttr(nl.IFLA_VXLAN_RSC, boolAttr(vxlan.RSC))
 	data.AddRtAttr(nl.IFLA_VXLAN_L2MISS, boolAttr(vxlan.L2miss))
 	data.AddRtAttr(nl.IFLA_VXLAN_L3MISS, boolAttr(vxlan.L3miss))
+	data.AddRtAttr(nl.IFLA_VXLAN_UDP_CSUM, boolAttr(vxlan.UDPCSum))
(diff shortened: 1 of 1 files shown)


aboch:
@p-sherratt I understand the change, it modifies the current contract, but this is not a main concern.

In https://github.com/vishvananda/netlink/issues/1144 you point out it is not possible to disable the UDP checksumming on an existing vxlan link via iproute2 commands and you point out to the reason, the kernel code explicitly not honoring such request.

In your change, you modify _addVxlanAttrs_ to always send to kernel the value of Vxlan.UDPcsum.
But the function is only called in the link modify path: _linkModify->addVxlanAttrs_

Won't kernel now always reject a linkModify call over a vxlan link?

What am I missing?

PR #1158, opened 2026-01-25
Add support for getting links in different namespaces

gwenya:
Rtnetlink supports getting links from different namespaces by sending `IFLA_TARGET_NETNSID` with a `RTM_GETLINK` request. 
This is particularly useful for getting the peer of a veth, as the kernel tells us the netnsid of the peer but opening a netlink socket in that namespace requires a file descriptor into the namespace, and getting such a file descriptor from a netnsid requires enumerating all namespaces.

This PR adds two functions: `LinkByNameAndNsid` and `LinkByIndexAndNsid`. These functions are duplicates of `LinkByName` and `LinkByIndex`, but specify the `IFLA_TARGET_NETNSID` property in the request. The function `LinkByNameAndNsid` does not support the fallback functionality of getting the link from a dump, as the dump only includes links in the current namespace.

I am writing the nsid with `nl.Uint32Attr()`, but it is actually a signed 32 bit int. I can't find any function to write signed int attributes, I'm not sure if it makes a difference since go's cast should keep the bits the same (I think).

## Summary by CodeRabbit

* **New Features**
  * Query network interfaces by index within a specific network namespace
  * Query network interfaces by name within a specific network namespace

* **Bug Fixes**
  * Missing-interface handling now returns a clear “not found” result instead of a raw error

* **Tests**
  * Added tests and a test helper to validate namespace-targeted lookups and teardown routines

### DIFF
--- a/link_linux.go
+++ b/link_linux.go
@@ -2113,6 +2113,60 @@ func (h *Handle) LinkByIndex(index int) (Link, error) {
 	return execGetLink(req)
 }
 
+// LinkByIndexAndNsid finds a link by index in a different namespace and returns a pointer to the object
+func LinkByIndexAndNsid(index int, nsid int) (Link, error) {
+	return pkgHandle.LinkByIndexAndNsid(index, nsid)
+}
+
+// LinkByIndexAndNsid finds a link by index in a different namespace and returns a pointer to the object
+func (h *Handle) LinkByIndexAndNsid(index int, nsid int) (Link, error) {
+	req := h.newNetlinkRequest(unix.RTM_GETLINK, unix.NLM_F_ACK)
+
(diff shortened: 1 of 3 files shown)


vanytsvetkov:
I may be missing some context, but why not define your own `Handle` (see the `pkgHandle` initialization) with a dedicated namespace and request the `veth` and other objects through it? Works in practice, we use it.

gwenya:
@aboch I am not entirely certain but I think it has to be under handle, since the netns IDs are relative to the namespace they are used from. I.e. namespace A might have ID 0 from the perspective of namespace B but ID 1 from the perspective of namespace C.

I do think that we probably don't actually need the method to get by name and netns id, since the only use case I am aware of is getting the peer of a veth interface, which would always be by index and netns id

aboch:
please fix the conflicts

PR #1161, opened 2026-01-29
Use binary.NativeEndian

tklauser:
Use the NativeEndian native-endian var provided by the Go standard library encoding/binary package since Go 1.21.

Comparing binary.NativeEndian to binary.{Bit,Little}Endian doesn't work though, so instead use golang.org/x/sys/cpu.IsBigEndian as suggested by https://github.com/golang/go/issues/57237.

/cc @vishvananda @aboch

## Summary by CodeRabbit

* **Refactor**
  * Simplified endianness detection mechanism by replacing internal caching and detection logic with a direct runtime check.
  * Removed unsafe code paths previously used for endianness computation.
  * Updated internal imports to use standard platform detection utilities.

* **Tests**
  * Updated test endianness checks to use standard platform detection method.

### DIFF
--- a/conntrack_test.go
+++ b/conntrack_test.go
@@ -5,7 +5,6 @@ package netlink
 
 import (
 	"bytes"
-	"encoding/binary"
 	"fmt"
 	"net"
 	"os"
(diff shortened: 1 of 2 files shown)


aboch:
Thanks @tklauser 

As I understand, this change implies go1.21 is now a requirement for future releases of vish netlink.
I guess it is fine since it's from 3 years ago and golang base packages keep changing.

Let's merge and see.

PR #1183, opened 2026-04-16
fix: route serializing nil IP address

lanphgphm:
# Problem
Currently, destination IP address is always appended & sent to the kernel. This doesn't make sense if the route type are among 'unreachable', 'blackhole', 'prohibit' where the request is dropped (no destination required).

## Example problematic usage
```go
rt := &netlink.Route{
    Table: tableID,
    Dst: &net.IPNet{
	    IP:   net.IP{},
	    Mask: net.IPMask{},
    },
    Priority: math.MaxInt32,
    Type:     unix.RTN_UNREACHABLE,
    Scope:    unix.RT_SCOPE_UNIVERSE,
}

netlink.RouteAdd(rt)
```

which when run with `strace` shows that the constructed message is 48 byte long: 
```bash
sendto(4, [{nlmsg_len=48, nlmsg_type=RTM_NEWROUTE, nlmsg_flags=NLM_F_REQUEST|NLM_F_ACK|NLM_F_EXCL|NLM_F_CREATE, nlmsg_seq=13, nlmsg_pid=0}, {rtm_family=AF_INET, rtm_dst_len=0, rtm_src_len=0, rtm_tos=0, rtm_table=0x65, rtm_protocol=RTPROT_BOOT, rtm_scope=RT_SCOPE_UNIVERSE, rtm_type=RTN_UNREACHABLE, rtm_flags=0}, [{nla_len=4, nla_type=RTA_DST}, [{nla_len=8, nla_type=RTA_PRIORITY}, 2147483647], [{nla_len=8, nla_type=RTA_OIF}, 0]]], 48, 0, {sa_family=AF_NETLINK, nl_pid=0, nl_groups=00000000}, 12) = 48
```

and the kernel complains "numerical result out of range" 

# The fix 
This patch aims to fix that by adding route type filter before appending the destination IP address.

## Summary by CodeRabbit

* **Refactor**
  * Improved Linux route request behavior for unreachable/blackhole/prohibit route types when the destination is unspecified, avoiding incorrect destination attributes and improving reliability.
* **Tests**
  * Added tests to exercise adding, listing, and deleting those route types with an empty destination to prevent regressions.

### DIFF
--- a/route_linux.go
+++ b/route_linux.go
@@ -929,7 +929,12 @@ func (h *Handle) prepareRouteReq(route *Route, req *nl.NetlinkRequest, msg *nl.R
 		} else {
 			dstData = route.Dst.IP.To16()
 		}
-		rtAttrs = append(rtAttrs, nl.NewRtAttr(unix.RTA_DST, dstData))
+		shouldSkipDst := (route.Type == unix.RTN_UNREACHABLE ||
+			route.Type == unix.RTN_BLACKHOLE ||
+			route.Type == unix.RTN_PROHIBIT) && (dstLen == 0)
+		if !shouldSkipDst {
+			rtAttrs = append(rtAttrs, nl.NewRtAttr(unix.RTA_DST, dstData))
+		}
 	} else if route.MPLSDst != nil {
 		family = nl.FAMILY_MPLS
 		msg.Dst_len = uint8(20)
--- a/route_test.go
+++ b/route_test.go
@@ -152,6 +152,63 @@ func TestRouteAddDel(t *testing.T) {
 	}
 }
 
+func TestRouteUnreachableEmptyDst(t *testing.T) {
+	t.Cleanup(setUpNetlinkTest(t))
+
+	// Test adding unreachable/blackhole/prohibit routes with empty Dst.IP
+	// These route types don't need RTA_DST to be serialized
+	testCases := []struct {
+		name      string
+		routeType int
+	}{
+		{"unreachable", unix.RTN_UNREACHABLE},
+		{"blackhole", unix.RTN_BLACKHOLE},
+		{"prohibit", unix.RTN_PROHIBIT},
+	}
+
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			route := &Route{
+				Table: 100,
+				Dst: &net.IPNet{
+					IP:   net.IP{},
+					Mask: net.IPMask{},
+				},
+				Priority: 100,
+				Type:     tc.routeType,
+				Scope:    unix.RT_SCOPE_UNIVERSE,
+				Family:   FAMILY_V4,
+			}
+
+			if err := RouteAdd(route); err != nil {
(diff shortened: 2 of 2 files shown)


route_linux.go
```diff
@@ -929,7 +929,12 @@ func (h *Handle) prepareRouteReq(route *Route, req *nl.NetlinkRequest, msg *nl.R
 		} else {
 			dstData = route.Dst.IP.To16()
 		}
-		rtAttrs = append(rtAttrs, nl.NewRtAttr(unix.RTA_DST, dstData))
+		shouldSkipDst := (route.Type == unix.RTN_UNREACHABLE ||
+			route.Type == unix.RTN_BLACKHOLE ||
+			route.Type == unix.RTN_PROHIBIT) && (dstLen == 0)
```
aboch, line 934: Why this condition: ` && (dstLen == 0)` ?

Can you please help me understand what request, in `ip route` command terms, would get to here?
This is testing on specific route types (unreachable, blackhole, prohibit) which I assume make sense only for `ip route add ...`. Is that correct? Then formulating such request with a dstLen==0, meaning a /0 PREFIX seems to be a bad request. Should be caught earlier.

Instead it is embeded here into checking whether or not skipping the dstdata population.

lanphgphm: Yes, this patch aims at `ip route add ...` commands. 

The equivalent commands in terms of `ip route` that would get to here are: 
```bash 
ip route add unreachable default 
ip route add blackhole default 
ip route add prohibit default 

ip route add unreachable 0.0.0.0/0
ip route add blackhole 0.0.0.0/0 
ip route add prohibit 0.0.0.0/0 

ip route add unreachable ::/0 
ip route add blackhole ::/0 
ip route add prohibit ::/0 
```
From my understanding of `iproute2`, the right IP address family (AF) still need to be set even if the message carries no destination address. `default` is an alias to 0.0.0.0/0 or ::/0, and the AF (v4 or v6) is inferred from the context:
```bash 
ip route add blackhole default           # IPv4 assumed                                                                              
ip -6 route add blackhole default       # IPv6 inferred from -6 flag                                                                        
ip route add blackhole default via ::1 # IPv6 inferred from gateway                                              
ip route add blackhole ::/0                 # IPv6 inferred from destination            
```
I understand your comments as suggestion to move the `dstLen` checking earlier so as to avoid unnecessary calculation. But finding & setting AF is not skippable in this case, so for now I would still propose to keep the patch as-is. 

Let me know if you disagree or if I miss your point.

aboch: Thank you for the explanation

lanphgphm: Hi, if there is any feedback/disagreement per this patch, would you let me know? I would like to use upstream netlink instead of maintaining a separate fork in the long run, and I'm willing to put in work for revisions if this contribution is not up to standards. 

In the case where you think this is a niche usage that might be problematic to merge, I'd like to know as well.

aboch:
LGTM

PR #1186, opened 2026-04-24
conntrack: handle flow Status

treuherz:
Currently the Status field of conntrack flows is ignored. This change:

* adds constants for the bits of the status field
* parses the Status field when dumping flows
* handles Status values when creating or updating flows
* adds status strings to ConntrackFlow.String() to match the conntrack
  cli

with unit tests for all of the above.

The change to String() could cause backwards compatibility issues. Let me know if 
that's a problem and I can revert those bits.

## Summary by CodeRabbit

* **New Features**
  * Conntrack flows now include and display status markers (e.g., unreplied, assured, offload) in human-readable output.
  * Filtering can now match or exclude flows by status bits; a new API lets callers add status criteria to filters.

* **Tests**
  * Expanded test coverage for status handling, parsing/formatting, and filter matching to ensure correct behavior.

### DIFF
--- a/conntrack_linux.go
+++ b/conntrack_linux.go
@@ -251,6 +251,7 @@ type ProtoInfoTCP struct {
 
 // Protocol returns "tcp".
 func (*ProtoInfoTCP) Protocol() string { return "tcp" }
+
 func (p *ProtoInfoTCP) toNlData() ([]*nl.RtAttr, error) {
 	ctProtoInfo := nl.NewRtAttr(unix.NLA_F_NESTED|nl.CTA_PROTOINFO, []byte{})
 	ctProtoInfoTCP := nl.NewRtAttr(unix.NLA_F_NESTED|nl.CTA_PROTOINFO_TCP, []byte{})
@@ -289,7 +290,6 @@ type IPTuple struct {
 // toNlData generates the inner fields of a nested tuple netlink datastructure
 // does not generate the "nested"-flagged outer message.
 func (t *IPTuple) toNlData(family uint8) ([]*nl.RtAttr, error) {
-
 	var srcIPsFlag, dstIPsFlag int
 	if family == nl.FAMILY_V4 {
 		srcIPsFlag = nl.CTA_IP_V4_SRC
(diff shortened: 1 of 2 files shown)


conntrack_linux.go
```diff
@@ -327,21 +346,39 @@ type ConntrackFlow struct {
...
+	}
+	if s.Status&ConntrackStatusHWOffload != 0 {
+		repl = "[HW_OFFLOAD] "
+	} else if s.Status&ConntrackStatusOffload != 0 {
+		repl = "[OFFLOAD] "
+	} else if s.Status&ConntrackStatusAssured != 0 {
+		repl = "[ASSURED] "
+	}
+	return
+}
```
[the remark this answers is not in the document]
treuherz: I think you are incorrect, and that they are mutually exclusive in the conntrack cli. This output is handled by the `__snprintf_status_assured` function in the libnetfilter_conntrack repo, under src/conntrack/snprintf_default.c. That function reads:

```c
static int __snprintf_status_assured(char *buf,
                     unsigned int len,
                     const struct nf_conntrack *ct)
{
    int size = 0;

    if (ct->status & IPS_HW_OFFLOAD)
        size = snprintf(buf, len, "[HW_OFFLOAD] ");
    else if (ct->status & IPS_OFFLOAD)
        size = snprintf(buf, len, "[OFFLOAD] ");
    else if (ct->status & IPS_ASSURED)
        size = snprintf(buf, len, "[ASSURED] ");

    return size;
}
```

Only one of (or none of) `HW_OFFLOAD`, `OFFLOAD` or `ASSURED` will be printed.

conntrack_linux.go
```diff
@@ -327,21 +346,39 @@ type ConntrackFlow struct {
 	TimeStart  uint64
 	TimeStop   uint64
 	TimeOut    uint32
+	Status     uint32
 	Labels     []byte
 	ProtoInfo  ProtoInfo
 }
 
+func (s *ConntrackFlow) statusStrings() (orig, repl string) {
+	if s.Status&ConntrackStatusSeenReply == 0 {
```
aboch, line 355: Is `ConntrackStatusSeenReply ` mutually exclusive with any other status bit?

conntrack_linux.go
```diff
@@ -327,21 +346,39 @@ type ConntrackFlow struct {
...
+		orig = "[UNREPLIED] "
+	}
+	if s.Status&ConntrackStatusHWOffload != 0 {
+		repl = "[HW_OFFLOAD] "
+	} else if s.Status&ConntrackStatusOffload != 0 {
+		repl = "[OFFLOAD] "
+	} else if s.Status&ConntrackStatusAssured != 0 {
+		repl = "[ASSURED] "
+	}
+	return
```
aboch: Would it make sense to print the hex value of status for all other bits we are not providing a string form for?

conntrack_linux.go
aboch: oh I see, this is for orig only

conntrack_linux.go
treuherz: Sounds good, I can add that. I didn’t know how closely you wanted to follow the conntrack cli’s output. Any preferences on where in the string it goes? `%#x` vs `%#b`?

aboch: However, just retain the brakets "[\<value\>]"

aboch:
LGTM

treuherz:
Great, thank you. Do you know when you’re next going to cut a release?

PR #1190, opened 2026-06-03
rule: Changes IPProto type to u8

tsheinen:
netlink defines[1] the type of FRA_IP_PROTO to be a u8 but the definition in rule.go treats this field as a u32. This causes messages to appear in kernel logs with "attribute type 22 has an invalid length." This is exercised in TestRuleAddDel, although it does not cause the test to fail because netlink parses it permissively.

[1]: https://elixir.bootlin.com/linux/v6.19.14/source/net/core/fib_rules.c#L862

## Summary by CodeRabbit

* **Refactor**
  * Improved handling of protocol values in network rule creation and listing for better compatibility with the kernel, resulting in more reliable rule application and retrieval.

### DIFF
--- a/rule.go
+++ b/rule.go
@@ -25,7 +25,7 @@ type Rule struct {
 	Invert            bool
 	Dport             *RulePortRange
 	Sport             *RulePortRange
-	IPProto           int
+	IPProto           int // IPProto is represented as uint8 in the kernel and will be truncated before sending
 	UIDRange          *RuleUIDRange
 	Protocol          uint8
 	Type              uint8
--- a/rule_linux.go
+++ b/rule_linux.go
@@ -154,9 +154,7 @@ func ruleHandle(rule *Rule, req *nl.NetlinkRequest) error {
 	}
 
 	if rule.IPProto > 0 {
-		b := make([]byte, 4)
-		native.PutUint32(b, uint32(rule.IPProto))
-		req.AddData(nl.NewRtAttr(nl.FRA_IP_PROTO, b))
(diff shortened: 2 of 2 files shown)


vishvananda:
coderabbit's suggestion is correct here, we don't want to change the public interface here as it may break callers.

tsheinen:
> coderabbit's suggestion is correct here, we don't want to change the public interface here as it may break callers.

I considered that, but was of the opinion that it was an easily fixed breaking change and that it would be worse to silently change the semantics outside of the type system. If we need to narrow u32 to u8 before sending it to the kernel, we'll either have to return an error or silently discard upper bits. I suppose that given the kernel is already silently discarding larger values, doing it in userspace as well can't really change semantics. 

Do you have a preference on how to handle the narrowing?

aboch:
I agree it's unfortunate the Rule struct has `IPProto` as an `int`, but also I the issue is minor. The changes required to not break the existing API, which I am afraid are creating a new Rule2 struct, may not be worth in this case.

tsheinen:
I amended the commit to not change the type of `IPProto` and instead truncate it in userspace.

aboch:
@tsheinen Have you tested this on a big endian OS?

tsheinen:
I haven't. It's been tested on amd64 linux and whatever is tested in CI (which seems to also be amd64 linux). I can do so though, do you have a preference on which arch or just that it's big-endian? I did some light searching and found a few options for running github actions inside qemu (https://github.com/uraimo/run-on-arch-action). I could also include a ci action to run the tests inside s390x linux or similar.

PR #1192, opened 2026-06-08
nl: pool receive buffers before Go 1.26

mozhuli:
**Motivation**

fix excessive allocations in NetlinkSocket.Receive when building with Go versions before 1.26.

**Description**
Before Go 1.26, the receive buffer used by Receive may escape and be heap allocated on every receive call. Under high-frequency netlink receive workloads, such as kube-proxy/IPVS sync loops, this can cause significant allocation pressure and GC overhead.

To avoid this for pre-Go 1.26 builds, this PR uses a `sync.Pool` to reuse large receive buffers.

For Go 1.26 and newer, Receive keeps using a local stack buffer, since the newer compiler can handle this case without the same allocation issue.  https://go.dev/doc/go1.26#compiler

The build tags are split as follows:
- go1.26: use the local receive buffer implementation.
- !go1.26: use the a `sync.Pool` to reuse large receive buffers.

**Compatibility**

Go 1.26 and newer keep the existing per-call local buffer behavior.
Go 1.25 and older use a `sync.Pool` to reuse large receive buffers.
Future Go versions, such as Go 1.27, will continue to match the go1.26 build tag and use the Go 1.26+ implementation.

### DIFF
--- a/nl/nl_linux.go
+++ b/nl/nl_linux.go
@@ -919,14 +919,7 @@ func (s *NetlinkSocket) Receive() ([]syscall.NetlinkMessage, *unix.SockaddrNetli
 	if err != nil {
 		return nil, nil, err
 	}
-	var (
-		deadline time.Time
-		fromAddr *unix.SockaddrNetlink
-		rb       [RECEIVE_BUFFER_SIZE]byte
-		nr       int
-		from     unix.Sockaddr
-		innerErr error
-	)
+	var deadline time.Time
 	receiveTimeout := atomic.LoadInt64(&s.receiveTimeout)
 	if receiveTimeout != 0 {
 		deadline = time.Now().Add(time.Duration(receiveTimeout))
@@ -934,13 +927,7 @@ func (s *NetlinkSocket) Receive() ([]syscall.NetlinkMessage, *unix.SockaddrNetli
 	if err := s.file.SetReadDeadline(deadline); err != nil {
 		return nil, nil, err
 	}
-	err = rawConn.Read(func(fd uintptr) (done bool) {
-		nr, from, innerErr = unix.Recvfrom(int(fd), rb[:], 0)
-		return innerErr != unix.EWOULDBLOCK
-	})
-	if innerErr != nil {
-		return nil, nil, innerErr
-	}
+	nl, from, err := receiveNetlinkMessages(rawConn)
 	if err != nil {
 		// The timeout was previously implemented using SO_RCVTIMEO on a blocking
 		// socket. So, continue to return EAGAIN when the timeout is reached.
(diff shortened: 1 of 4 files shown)


mozhuli:
fix issue: https://github.com/kubernetes/kubernetes/issues/135230

mozhuli:
@aojea please have a look if this make sense?

nl/nl_linux.go
```diff
@@ -739,8 +741,9 @@ func getNetlinkSocket(protocol int) (*NetlinkSocket, error) {
 		return nil, err
 	}
 	s := &NetlinkSocket{
-		fd:   int32(fd),
-		file: os.NewFile(uintptr(fd), "netlink"),
+		fd:            int32(fd),
+		file:          os.NewFile(uintptr(fd), "netlink"),
+		receiveBuffer: make([]byte, RECEIVE_BUFFER_SIZE),
```
aojea: This will allocate a fixed amount of memory to every single socket, so if we have N sockets idle we will allocate NxRECEIVE_BUFFER_SIZE memory, right?

The canonical way I see to handle this problem is to use a https://pkg.go.dev/sync#Pool assuming we escape to the heap

aojea:
@mozhuli it seems this is not needed since in new versions the golang compiler fixed the heap escape

```go
package main

import (
	"fmt"
	"runtime"
	"testing"

	"golang.org/x/sys/unix"
)

func main() {
	s, err := nl.Subscribe(unix.NETLINK_ROUTE, unix.RTNLGRP_NEIGH)
	if err != nil {
		panic(err)
	}
	defer s.Close()

	s.SetReceiveTimeout(&unix.Timeval{Sec: 0, Usec: 1})

	allocs := testing.AllocsPerRun(100, func() {
		s.Receive()
	})
	
	var m1, m2 runtime.MemStats
	runtime.ReadMemStats(&m1)
	for i := 0; i < 100; i++ {
		s.Receive()
	}
	runtime.ReadMemStats(&m2)

	fmt.Printf("Allocs per run: %v, Bytes per run: %v\n", allocs, (m2.TotalAlloc-m1.TotalAlloc)/100)
}

```


```sh
aojea@aojea:~/src/netlink$ GOTOOLCHAIN=go1.23.0 go run /tmp/test_alloc.go
Allocs per run: 7, Bytes per run: 66094
aojea@aojea:~/src/netlink$ GOTOOLCHAIN=go1.26.0 go run /tmp/test_alloc.go
Allocs per run: 1, Bytes per run: 112
```

aojea:
So, current implementation is preferred because avoid allocating space if unnecessary, what I do not know are the policies of this project for old golang versions, in that case typically people build different behaviors using the golang tags , per example https://github.com/golang/net/tree/master/http2 

https://github.com/golang/net/blob/master/http2/client_priority_go126.go
https://github.com/golang/net/blob/master/http2/client_priority_go127.go

mozhuli:
> So, current implementation is preferred because avoid allocating space if unnecessary, what I do not know are the policies of this project for old golang versions, in that case typically people build different behaviors using the golang tags , per example https://github.com/golang/net/tree/master/http2
> 
> https://github.com/golang/net/blob/master/http2/client_priority_go126.go https://github.com/golang/net/blob/master/http2/client_priority_go127.go

@aojea updated, is this make sense?

go1.26: use the local receive buffer implementation.
!go1.26: use the a sync.Pool to reuse large receive buffers.

aojea:
I'm not maintainer here, I need to review , but please align first with current maintainers like @aboch

mozhuli:
@aboch please help review if this make sense? https://github.com/kubernetes/kubernetes/issues/135230

nl/nl_linux.go
mozhuli: right, Updated with `sync.Pool` to reuse receive buffers before go1.26 @aojea

PR #1194, opened 2026-06-10
netkit: serialize peer NumRxQueues/NumTxQueues on create

aarcamp:
addNetkitAttrs emitted the peer's name, MTU, GSO sizes and namespace into the IFLA_NETKIT_PEER_INFO nest, but not its queue counts, so a peer created via SetPeerAttrs always came up with the default single rx/tx queue. veth already serializes PeerNumRxQueues/PeerNumTxQueues; do the same for netkit.

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced peer network interface configuration by adding support for receive and transmit queue attribute settings during device modifications.


borkmann (approved):

aboch:
LGTM

PR #1196, opened 2026-06-21
fix: normalize IPv4 addresses and fix byte alignment in Multipath Via attribute

### LINKED ISSUE #1064: RouteAdd with Multipath specifying "Via" fails with IP obtained from netlink.ParseIP().IP
I encountered this issue/bug while testing the library.

```go
	ecmpRoute := &netlink.Route{
		// LinkIndex: br.Index,
		Dst: &net.IPNet{
			IP:   net.ParseIP("1.1.1.1"), // TODO: LOADBALANCER IP!!!
			Mask: net.CIDRMask(32, 32),
		},
		MultiPath: nil,
	}
        [...]
		tapAddr, err := netlink.ParseAddr(tap.Addr)
		[...]
		ecmpRoute.MultiPath = append(ecmpRoute.MultiPath, &netlink.NexthopInfo{
			Via: &netlink.Via{
				AddrFamily: netlink.FAMILY_V4,
				Addr:       tapAddr.IP.To4(), // strangely IP must be set with To4() otherwise will fail :shrug:
			},
		})
	[...]
	if err := netlink.RouteAdd(ecmpRoute); err != nil {
        // Throws: invalid argument
```

I'd be happy to solve that issue if someone can point me where I should start looking into?

Otherwise I'd believe the library is very nice. What I would really like is more context to errors. I'm usually adding context to errors when propagating errors. The following example should demonstrate my point.

```go
if err != nil {
    return errors.Join(err, errors.New("<some context>")) // instead of `return err`
}
```

With that I can easily grep into the netlink repo and check where error the error originated.

Anyway, it's a great repo and I'd be happy to contribute.

barisaydogdu:
Fixes an issue where passing an IPv4 address generated by `net.ParseIP()` to a Multipath `Via` attribute causes the kernel to reject the route with an `invalid gateway` error. 

The problem is that `net.ParseIP()` returns a 16-byte slice, and `Via.Encode()` was writing all 16 bytes to the `RTA_VIA` attribute. The kernel strictly expects 4 bytes for `AF_INET`. 

This PR simply adds a `.To4()` conversion inside `Via.Encode()` for `FAMILY_V4` to ensure we only serialize the correct 4 bytes. I also included a test (`TestRouteMultiPathViaIPv4Mapped`) to cover this case.

Fixes #1064

## Summary by CodeRabbit

* **Bug Fixes**
  * Normalized IPv4 `Via` address encoding by converting inputs to proper IPv4 form, returning an error when an IPv4 conversion isn’t possible to better match kernel behavior.

* **Tests**
  * Added a kernel-version-gated regression test for multipath routes using IPv4-mapped `Via` addresses (add/verify/delete).
  * Added unit tests for `Via` encoding with IPv4 `AddrFamily`, validating the encoded family and IPv4 bytes for `1.1.1.1`.

### DIFF
--- a/route_linux.go
+++ b/route_linux.go
@@ -780,11 +780,20 @@ func (v *Via) Family() int {
 
 func (v *Via) Encode() ([]byte, error) {
 	buf := &bytes.Buffer{}
+
 	err := binary.Write(buf, native, uint16(v.AddrFamily))
 	if err != nil {
 		return nil, err
 	}
-	err = binary.Write(buf, native, v.Addr)
+	addr := v.Addr
+	if v.AddrFamily == nl.FAMILY_V4 {
+		ipv4 := addr.To4()
+		if ipv4 == nil {
+			return nil, fmt.Errorf("invalid IPv4 address in Via struct")
+		}
+		addr = ipv4
+	}
+	err = binary.Write(buf, native, addr)
 	if err != nil {
 		return nil, err
 	}
--- a/route_test.go
+++ b/route_test.go
@@ -2472,6 +2472,101 @@ func TestRouteViaAddDel(t *testing.T) {
 	}
 }
 
+// TestRouteMultiPathViaIPv4Mapped verifies that adding a Multipath route
+// with a 16-byte IPv4-mapped address is correctly normalized to 4 bytes
+// during encoding, preventing kernel invalid gateway rejections.
+func TestRouteMultiPathViaIPv4Mapped(t *testing.T) {
+	minKernelRequired(t, 5, 4)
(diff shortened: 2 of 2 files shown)


barisaydogdu:
Hi @vishvananda, @aboch 

The PR is ready for review. All CI checks are passing, and full lifecycle regression tests are added for the Multipath Via attribute. Whenever you have a moment, I'd appreciate your feedback.

cc:  @alexandremahdhaoui  (This should resolve the invalid argument error you encountered in #1064).

aboch:
LGTM

PR #1174, opened 2026-03-18
 handle: add NewHandleWithOptions(), deprecate DisableVFInfoCollection()

ti-mo:
Repurpose HandleOptions as a declarative configuration mechanism for Handle. It can be passed to NewHandleWithOptions.

It also removes the unreleased Handle.RetryInterrupted and deprecates DisableVFInfoCollection in favor of its HandleOptions counterpart.

## Summary by CodeRabbit

* **New Features**
  * One-time global configuration for handle initialization; subsequent attempts now return an error.
  * New options to control retry behavior, explicit network-namespace selection, and VF-info collection.

* **Changes**
  * Consolidated handle creation through an options-based path.
  * Lookup logic now tracks dump-based lookups atomically and respects the VF-info collection toggle when listing/looking up links.

* **Bug Fixes**
  * Test added to validate one-time configuration semantics.

* **Deprecated**
  * Fluent VF-info disable method replaced by the new option.

PR #1197, opened 2026-07-02
*: remove pkgHandle, remove HandleOptions.NetNS, add NewHandleAtWithOptions

ti-mo:
https://github.com/vishvananda/netlink/pull/1174 as it was merged had some problems. I never really vetted it against Cilium as I had to take a break from work and switch tasks. It had a few conceptual mistakes, mostly stemming from my lack of understanding of `pkgHandle`. The most problematic being ConfigureHandle() locking all global functions to the netns of the caller of ConfigureHandle, making e.g. LinkList operate in the main netns.

This PR proposes removing `pkgHandle` outright, as it only serves as a zero placeholder to promote its methods to package-global functions, and its presence being rather confusing. It also never actually holds any sockets, as doing so would break global function calls from other network namespaces, a common pattern throughout the ecosystem. Also, package-global state of this kind is not a good idea.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved consistency by standardizing how the active handle is obtained for many networking operations (links, routes, neighbors, filters, bridges, sockets, etc.), including subscription/dump request construction.
  * Corrected several wrapper delegations to call the right operations (e.g., link setup and several VDPA queries).

* **Breaking Changes / Behavior**
  * `HandleOptions` no longer includes a `NetNS` field; namespace selection is now passed explicitly when creating handles.

* **Tests**
  * Updated affected unit tests and retry/handle setup expectations.

### DIFF
--- a/addr_linux.go
+++ b/addr_linux.go
@@ -31,7 +31,7 @@ const (
 // will be automatically computed based on the IP mask if /30 or larger.
 // If `net.IPv4zero` is given as the broadcast address, broadcast is disabled.
 func AddrAdd(link Link, addr *Addr) error {
-	return pkgHandle.AddrAdd(link, addr)
+	return pkgHandle().AddrAdd(link, addr)
 }
 
 // AddrAdd will add an IP address to a link device.
@@ -54,7 +54,7 @@ func (h *Handle) AddrAdd(link Link, addr *Addr) error {
 // will be automatically computed based on the IP mask if /30 or larger.
 // If `net.IPv4zero` is given as the broadcast address, broadcast is disabled.
 func AddrReplace(link Link, addr *Addr) error {
-	return pkgHandle.AddrReplace(link, addr)
+	return pkgHandle().AddrReplace(link, addr)
 }
 
 // AddrReplace will replace (or, if not present, add) an IP address on a link device.
@@ -73,7 +73,7 @@ func (h *Handle) AddrReplace(link Link, addr *Addr) error {
 //
 // Equivalent to: `ip addr del $addr dev $link`
 func AddrDel(link Link, addr *Addr) error {
-	return pkgHandle.AddrDel(link, addr)
+	return pkgHandle().AddrDel(link, addr)
 }
 
 // AddrDel will delete an IP address from a link device.
@@ -191,7 +191,7 @@ func (h *Handle) addrHandle(link Link, addr *Addr, req *nl.NetlinkRequest) error
 // If the returned error is [ErrDumpInterrupted], results may be inconsistent
 // or incomplete.
 func AddrList(link Link, family int) ([]Addr, error) {
-	return pkgHandle.AddrList(link, family)
+	return pkgHandle().AddrList(link, family)
 }
 
 // AddrList gets a list of IP addresses in the system.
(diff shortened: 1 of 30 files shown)


ti-mo:
@coderabbitai review

ti-mo:
@coderabbitai review

ti-mo:
@aboch PTAL

aboch:
@ti-mo 

> pkgHandle outright, as it only serves as a zero placeholder to promote its methods to package-global functions

Some context about this 10 year old thing. Initially this library only had pkg level methods. Then @mrjana and I encountered the runaway thread issue while operating across nw namespaces in Docker's libnetwork which causes havoc. Golang did not have yet a fix for it. We resorted to the handle thing which would allow the caller to pass the net ns; the first handle's method call would open the nl socket in the current namespace causing the handle to stick to it. 

To retain the original pkg level methods, a pkg level handle was created. 
As a, undocumented maybe, feature, as long as the first package level method call is done in the default netns, every other pkg level call from any netns would operate in the default netns.

Now, I do not know the current state and reliability of golang threads stitching to OS threads, but I see your change would generate a new handle each time a pkg level method is called. As a side effect, I see we lose the undocumented behavior above.

People like you who use this library in their everyday work/products are in the best position to tell whether or not retaining that behavior matters.

ti-mo:
Hi @aboch, thanks for the reply. I was also working on netlink-related stuff during that era. Fortunately, the runtime gave us the necessary tools in 1.10 with `runtime.LockOSThread`. (looks like the WeaveWorks domain is now being squatted :( Martynas' blog still has a copy: https://lambda.lt/blog/2018/go_netns_issue_fix.html)

>  As a, undocumented maybe, feature, as long as the first package level method call is done in the default netns, every other pkg level call from any netns would operate in the default netns.

I don't think this has ever been true. At least as far back as 2016, appc/cni was calling package-level API from netns-locked goroutines: https://github.com/cilium/cilium/commit/ffb26f3bc24#diff-110a0dfd47350c0dcaee4f867828908c6943b766eee7b100ab3f38f0bd816545R107-R111:

```go
       err = ns.WithNetNS(hostNS, false, func(_ *os.File) error {
               hostVeth, err := netlink.LinkByName(hostVethName)
               if err != nil {
                       return fmt.Errorf("failed to lookup %q in %q: %v", hostVethName, hostNS.Name(), err)
               }
```

This requires every call to dial a new socket in order to work.

Looking at the code, `pkgHandle` has `sockets == nil`. `newNetlinkRequest` explicitly checks if h.sockets == nil and returns a plain request with no Sockets map attached. Then, in `executeIter`, `req.Sockets` is nil, so it falls through to `getNetlinkSocket(sockType)` (a brand-new socket opened in the calling thread's current netns) and closes it with `defer s.Close()` on return. So pkgHandle never holds a socket, and each global function operates in whatever netns the calling goroutine's OS thread happens to be in.

This patch makes this behaviour explicit and removes the ambiguity around having this globally-shared variable that's essentially unused.

ti-mo:
@aboch Ping, resolved a merge conflict.

aboch:
LGTM

2 commits, but for two separate things

PR #1198, opened 2026-07-06
route: support v4-mapped IPv6 nexthops on V6 routes

### LINKED ISSUE #1114: IPv4-mapped IPv6 address cannot be used as route gateway
Netlink still makes use of Golang's `net` package, instead of the newer `netip` package to represent prefixes and addresses. In multiple instances netlink checks if a user-given address is a IPv4 or an IPv6 address via `nl.GetIPFamily()`. This function calls `ip.To4()` and if this returns anything other than `nil` we identify `ip` to be a IPv4 address. The problem is that the function contract is not that `.To4` returns not-nil for IPv4 addresses but it returns non-nil if the address _can be represented_ as IPv4. The second definition includes IPv4-mapped IPv6 addresses, while the first does not. 

Example:
I want to create the following route entry:
```
2300:5800:ab00::2 via ::ffff:100.95.128.2 dev br3 proto bgp onlink 
```

With `ip` one can do this by calling:
```
ip route add 2300:5800:ab00::2 via ::ffff:100.95.128.2 dev br3 proto bgp onlink
```

One would assume that the following Go code works:
```go
nlh.RouteAdd(
			&netlink.Route{
				LinkIndex: remoteBridge.Attrs().Index,	// some index
				Dst:       route.Dst,					// 2300:5800:ab00::2/128
				Gw:        hostAddress.AsSlice(),		//::ffff:100.95.128.2
				Flags:     int(unix.RTNH_F_ONLINK),		// onlink
				Protocol:  186,							// bgp
			},
		)
```

But it returns `gateway, source, and destination ip are not the same IP family` since it translates the `Gw` to an IPv4 first. Moreover, there's no way to override this interpretation with e.g., another `Family` field.

Note that this behavior exists in multiple places around the code and not only in `nl.GetIPFamily()`. For example, in `neighHandle()` there's an direct call to `.To4` to again decide if an address is a IPv4 or IPv6.

related issue: https://github.com/vishvananda/netlink/issues/908

Interestingly, changing Gw with a Via struct also doesn't work:
```go
nlh.RouteAdd(
			&netlink.Route{
				LinkIndex: remoteBridge.Attrs().Index,	// some index
				Dst:       route.Dst,					// 2300:5800:ab00::2/128
				Via: &netlink.Via{
					AddrFamily: netlink.FAMILY_V6,
					Addr:       hostAddress.AsSlice(), //::ffff:100.95.128.2
				},	
				Flags:     int(unix.RTNH_F_ONLINK),		// onlink
				Protocol:  186,							// bgp
			},
		)
```
As this returns `invalid argument`. Though I haven't debugged the reason for this one yet.

fdomain:
This is an alternative proposal to fix #1114. The initial plan was to migrate to netip #1181, but considering the size of the change, we believe it would be more acceptable to make minor fixes to handle IPv4-mapped IPv6 address as gw.

Change:

A v4-mapped IPv6 address (::ffff:a.b.c.d) is byte-identical to its IPv4 form as a net.IP, so `nl.GetIPFamily` reports FAMILY_V4 for it. When such an address was used as a route's gateway on a V6 route it was encoded as a 4-byte AF_INET nexthop, and combined with a V6 destination it failed outright with "gateway... not the same IP family".

Resolve this locally in `prepareRouteReq` rather than changing `GetIPFamily` (which has no route context and must keep reporting V4 for the common net.ParseIP("1.2.3.4") case used by many other callers):

- Honor route.Family, which was previously ignored on the add path, so a gateway-only route (no destination to pin the family) can still be requested as V6.
- When the caller explicitly requests a V6 route via route.Family, let a 16-byte v4-mapped gateway (route.Gw and multipath nexthops) conform to it and be encoded as a 16-byte AF_INET6 nexthop.

Support is opt-in via route.Family so existing callers see no change: a v4-mapped gateway without route.Family set is still treated as V4 and rejected on a V6 route, exactly as before. The conform is also gated on a 16-byte slice so that an explicit 4-byte IPv4 gateway still errors rather than being silently reinterpreted; a 4-byte slice is the only unambiguous "I meant IPv4" signal available at the net.IP level.

Scope is limited to the gateway/nexthop, the real-world case (verified against the kernel: `ip -6 route` reports the nexthop as ::ffff:a.b.c.d). v4-mapped addresses are a socket-API representation construct and are not routable as IPv6 destinations, so Dst and Src are left unchanged.

Fixes #1114

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed route requests so v4-mapped IPv6 gateways are handled correctly when an IPv6 route family is explicitly requested.
  * Tightened validation to reject inconsistent gateway/address-family combinations instead of producing incorrect or ambiguous results.
  * Confirmed standard IPv4 gateway behavior remains unchanged.
* **Tests**
  * Added coverage to verify correct gateway encoding and error cases for both single-path and multipath routes.

### DIFF
--- a/route_linux.go
+++ b/route_linux.go
@@ -919,12 +919,33 @@ func (h *Handle) routeHandleIter(route *Route, req *nl.NetlinkRequest, msg *nl.R
 	return req.ExecuteIter(unix.NETLINK_ROUTE, 0, f)
 }
 
+// gwIPFamily returns the IP family to use for a gateway address on the given
+// route. A v4-mapped IPv6 gateway (::ffff:a.b.c.d, 16 bytes) reports FAMILY_V4
+// but is a valid V6 nexthop, so it is conformed to V6 only when the caller
+// explicitly requested a V6 route via route.Family. This makes the behavior
+// opt-in rather than silently reinterpreting an ambiguous address, and
+// requiring 16 bytes leaves an explicit 4-byte IPv4 gateway reporting V4.
+func gwIPFamily(route *Route, gw net.IP) int {
+	gwFamily := nl.GetIPFamily(gw)
+	if route.Family == FAMILY_V6 && gwFamily == FAMILY_V4 && len(gw) == net.IPv6len {
+		return FAMILY_V6
+	}
+	return gwFamily
+}
+
 func (h *Handle) prepareRouteReq(route *Route, req *nl.NetlinkRequest, msg *nl.RtMsg) error {
 	if req.NlMsghdr.Type != unix.RTM_GETROUTE && (route.Dst == nil || route.Dst.IP == nil) && route.Src == nil && route.Gw == nil && route.MPLSDst == nil {
 		return fmt.Errorf("either Dst.IP, Src.IP or Gw must be set")
 	}
 
 	family := -1
+	if route.Family != 0 {
+		// AF_UNSPEC (0) means unset. Honor an explicitly requested family so
+		// that ambiguous addresses (e.g. a v4-mapped IPv6 gateway, which is
+		// byte-identical to its IPv4 form as a net.IP) are encoded for the
+		// intended family even when no destination pins it down.
+		family = route.Family
+	}
 	var rtAttrs []*nl.RtAttr
 
 	if route.Dst != nil && route.Dst.IP != nil {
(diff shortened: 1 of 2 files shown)


fdomain:
Hi @vishvananda @aboch ,

This PR is ready for review, but the build is failing due to a seemingly unrelated issue. The same failure is visible on the following [PR](https://github.com/vishvananda/netlink/actions/runs/28803203057/job/85411800420?pr=1199).

Thanks in advance for taking a look at my change whenever you have time!

aboch:
LGTM

PR #1157, opened 2026-01-22
addr: add IFA_PROTO support for address protocol/origin

cgoncalves:
Add support for the `IFA_PROTO` netlink attribute (kernel 5.18+) which indicates the protocol or origin of an IP address.
This is similar to `RTPROT_*` for routes.

## Summary by CodeRabbit

* **New Features**
  * Added address protocol/origin support. New `Protocol` field enables setting and reading address protocol information (kernel 5.18+).
  * New constants available for address protocol values: `IFAPROT_UNSPEC`, `IFAPROT_KERNEL_LO`, `IFAPROT_KERNEL_RA`, `IFAPROT_KERNEL_LL`.

* **Tests**
  * Added test coverage for the new address protocol functionality with kernel version compatibility handling.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

PR #1201, opened 2026-07-16
addr: propagate Protocol field in AddrUpdate subscriptions

cgoncalves:
The IFA_PROTO attribute was parsed in parseAddr but not included in the AddrUpdate struct or its construction in addrSubscribeAt, so address subscription consumers never saw the protocol value.

IFA_PROTO support added in https://github.com/vishvananda/netlink/pull/1157

## Summary by CodeRabbit

* **New Features**
  * Address-change notifications now include the protocol value associated with each address update.
  * Applications can read this protocol metadata for both address additions and deletions, providing additional context when processing network address events.

### DIFF
--- a/addr_linux.go
+++ b/addr_linux.go
@@ -322,6 +322,7 @@ type AddrUpdate struct {
 	Scope       int
 	PreferedLft int
 	ValidLft    int
+	Protocol    int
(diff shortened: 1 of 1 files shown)


aboch:
for some reason I cannot force a rebuild. Can you please force push something.

PR #1209, opened 2026-08-06
handle: use per-request sockets for VF-only configuration

l8huang:
# description
PR https://github.com/vishvananda/netlink/pull/1174 changed ConfigureHandle to configure the package-wide `pkgHandle` through NewHandleWithOptions, which creates persistent netlink sockets in the `pkgHandle`.

Because `pkgHandle.sockets` remains bound to the network namespace where it was created, the following operations could target the wrong namespace after the calling OS thread changed namespaces.

This PR returns a handle without nil `sockets` when the caller only disables VF information collection:
```
NewHandleWithOptions(HandleOptions{
    DisableVFInfoCollection: true,
})
```

## Summary by CodeRabbit

* **New Features**
  * Handles configured only to disable virtual-function information collection can now operate without a persistent socket.
  * Network namespace behavior is now clearly documented for persistent and socketless handles.

* **Bug Fixes**
  * Preserved existing socket-based behavior when retry, namespace, or netlink family options are configured.

* **Tests**
  * Added coverage for option propagation, socket allocation, cleanup, retries, namespaces, and netlink family settings.

### DIFF
--- a/handle_linux.go
+++ b/handle_linux.go
@@ -58,8 +58,15 @@ type HandleOptions struct {
 	// [ErrDumpInterrupted].
 	RetryInterrupted bool
 
-	// NetNS specifies the network namespace to operate on. If not set, the
-	// current network namespace will be used.
+	// NetNS specifies the network namespace in which the Handle's persistent
+	// netlink sockets are created. A persistent socket remains bound to its
+	// creation namespace even if the calling OS thread later changes namespaces.
+	//
+	// When NetNS is non-nil, persistent sockets are created in that namespace.
+	// When NetNS is nil, persistent sockets are created in the calling OS thread's
+	// current namespace when the Handle is created. A socketless Handle instead
+	// creates a temporary socket for each operation in the calling OS thread's
+	// current namespace when the operation is executed.
 	NetNS *netns.NsHandle
 }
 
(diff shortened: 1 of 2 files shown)


thediveo:
I simply don't understand this PR and I'm afraid that because of its convoluted AI generated description that it would make the situation worse by complicating the current API hidden state and behavior even more.

l8huang:
@thediveo updated PR description, hope it's easier to understand now

handle_linux.go
```diff
@@ -199,7 +206,13 @@ func NewHandleAtFrom(newNs, curNs netns.NsHandle) (*Handle, error) {
 }
 
 // NewHandleWithOptions returns a Handle created using the specified options.
+// If DisableVFInfoCollection is the only option set and no netlink families are
+// specified, the returned Handle creates a socket for each request in the
+// caller's current network namespace instead of keeping persistent sockets.
 func NewHandleWithOptions(opts HandleOptions, nlFamilies ...int) (*Handle, error) {
+	if opts.DisableVFInfoCollection && !opts.RetryInterrupted && opts.NetNS == nil && len(nlFamilies) == 0 {
```
l8huang, line 213: I don't know how widely is NewHandleWithOptions() used now, so just want to open a small window to allow disabling VFInfoCollection without init Handle.sockets.

l8huang:
@vishvananda @aboch could you please take a look? Thanks.

PR #1212, opened 2026-08-16
Fix: Prevent buffer aliasing in netlink route destination parsing

TotyLabs:
Resolves buffer aliasing in deserializeRoute() that can cause corrupted IP addresses when netlink attributes contain oversized Len fields.
Root Cause: deserializeRoute() assigned byte slices from the netlink message buffer directly to route addresses. When RTA_DST, RTA_GATEWAY, or RTA_PREFSRC attributes contain oversized length fields, the resulting slices can include bytes beyond the expected IP address length and incorporate adjacent buffer data.
Solution:
Validate address attributes against the expected length for their address family.
Clamp oversized address data to 4 bytes for IPv4 or 16 bytes for IPv6.
Copy the validated address data into new allocations to avoid retaining references to the netlink message buffer.
Prevent corrupted route information from propagating to callers.
Impact: This issue was observed downstream in Cilium issue #47956, where corrupted route destination CIDRs caused iptables rule reconciliation to fail. In the reported case, bytes corresponding to the egress interface name were interpreted as an IPv4 address.
Testing: Added regression coverage for malformed netlink messages with oversized route-address attributes and verified that IPv4/IPv6 route addresses are constrained to the expected lengths. Normal netlink messages remain unaffected.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Linux route handling by validating gateway, preferred-source, and destination address lengths.
  * Correctly rejects undersized IPv4 and IPv6 route addresses.
  * Safely truncates oversized addresses while preserving correct address and mask widths.
  * Prevents route address data from being unintentionally altered during processing.
  * Improves handling of empty destinations for special routes.
  * Increases reliability when processing malformed or inconsistently sized route information.

### DIFF
--- a/route_linux.go
+++ b/route_linux.go
@@ -1389,9 +1389,73 @@ func deserializeRoute(m []byte) (Route, error) {
 	for _, attr := range attrs {
 		switch attr.Attr.Type {
 		case unix.RTA_GATEWAY:
-			route.Gw = net.IP(attr.Value)
+			// Validate gateway address length matches address family.
+			// For special routes (blackhole, unreachable, prohibit), allow empty/malformed gateway.
+			// Reject undersized attributes for standard routes; clamp oversized; allocate independent storage.
+			gwValue := attr.Value
+			expectedLen := 0
+			if msg.Family == nl.FAMILY_V4 {
+				expectedLen = 4
+			} else if msg.Family == nl.FAMILY_V6 {
+				expectedLen = 16
+			}
+
+			// For special route types, allow empty/undersized gateway.
+			isSpecialRoute := msg.Type == unix.RTN_BLACKHOLE ||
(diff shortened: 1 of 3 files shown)


HadrienPatte:
> Testing: Added regression coverage for malformed netlink messages with oversized route-address attributes

I don't see any new tests in this PR

TotyLabs:
@coderabbitai resume

TotyLabs:
@coderabbitai rewiew

TotyLabs:
@coderabbitai rewiew

TotyLabs:
@coderabbitai rewiew

TotyLabs:
Hi! Just following up in case this got buried. Is there anything else you'd like me to adjust or test before this can move forward? Thanks!

PR #1214, opened 2026-08-27
link: reject oversized TUN/TAP names in LinkAdd

FakeLearne:
## Summary

`LinkAdd` for TUN/TAP devices silently truncated interface names longer than 15 bytes. Callers then stored a name that did not match the device the kernel actually created, and later lookups/ioctls failed with confusing errors (`ERANGE` / `EINVAL`) instead of failing at create time.

This change rejects those names up front, the same way `Tuntap.AddQueues` and `unix.NewIfreq` already do.

## Problem

Linux `IFNAMSIZ` is 16, including a trailing NUL, so a user-visible name may be at most 15 bytes. `ip link add` reports `"name" too long`.

TUN/TAP `LinkAdd` did not follow that. It built a local `ifReq` and copied the name with:

```go
copy(req.Name[:15], base.Name)
```

For a 16-byte name such as `z192.168.195.183`:

1. `LinkAdd` succeeds.
2. The kernel device is actually `z192.168.195.18` (last byte dropped).
3. `link.Attrs().Name` is later refreshed from the ioctl result, but many callers keep the original string they passed in.
4. A later `LinkByName` / `unix.NewIfreq` / `TUNSETIFF` using the original 16-byte name fails. `unix.NewIfreq` returns `EINVAL` because `len(name) >= IFNAMSIZ`. Netlink `IFLA_IFNAME` is limited to `IFNAMSIZ-1`, so the kernel can also return `ERANGE` (`numerical result out of range`).

That split is hard to debug: create looked fine, and the failure showed up on a completely different path (for example reopening a pooled TAP fd).

The same repository already avoids this in `Tuntap.AddQueues` by calling `unix.NewIfreq`, which documents that a trailing NUL must fit.

## Change

In `linkModify`’s TUN/TAP branch:

- Validate the name with `unix.NewIfreq` and return an error if it does not fit (`EINVAL`, wrapped with the invalid name).
- Issue `TUNSETIFF` via `unix.IoctlIfreq` instead of a hand-rolled `ifReq` + `copy(Name[:15], …)`.
- Keep a per-queue copy of the `Ifreq` (same idea as the old `localReq` value copy) so one `TUNSETIFF` cannot leak flags into the next queue.
- Still write the kernel-returned name back onto `link.Attrs().Name` for empty names and templates such as `tap%d`.

Empty names (kernel-assigned) and names of length 1–15 are unchanged.

The unexported `ifReq` type is removed; it is unused after this. Exported `IFNAMSIZ` / `SizeOfIfReq` stay as part of the public API.

## Test

`TestLinkAddTuntapNameTooLong` asserts that a 16-byte TUN/TAP name fails `LinkAdd` with `unix.EINVAL`. `unix.NewIfreq` rejects the name before any ioctl, so the test does not need `CAP_NET_ADMIN` or a netns.

## Compatibility

This is a behavior change for invalid names only: `LinkAdd` now returns an error instead of creating a truncated device. Truncation was never documented, and callers that relied on it already had a name mismatch with the kernel.

Bond/ethtool ioctl helpers that still `copy(..., [:IFNAMSIZ-1])` are left alone; they are not the TUN/TAP create path.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved TUN/TAP interface creation reliability on Linux.
  - Interface names that exceed the supported length are now rejected with a clear validation error before setup begins.
  - Errors during multi-queue interface setup now include the affected queue index, making failures easier to diagnose.
  - Interface names are now reported consistently after successful creation.

### DIFF
--- a/link_linux.go
+++ b/link_linux.go
@@ -1493,47 +1493,48 @@ func (h *Handle) linkModify(link Link, flags int) error {
 		}
 
 		queues := tuntap.Queues
-
-		var fds []*os.File
-		var req ifReq
-		copy(req.Name[:15], base.Name)
-
-		req.Flags = uint16(tuntap.Flags)
+		flags := uint16(tuntap.Flags)
 
 		if queues == 0 { //Legacy compatibility
 			queues = 1
 			if tuntap.Flags == 0 {
-				req.Flags = uint16(TUNTAP_DEFAULTS)
+				flags = uint16(TUNTAP_DEFAULTS)
 			}
 		} else {
 			// For best peformance set Flags to TUNTAP_MULTI_QUEUE_DEFAULTS | TUNTAP_VNET_HDR
 			// when a) KVM has support for this ABI and
 			//      b) the value of the flag is queryable using the TUNGETIFF ioctl
 			if tuntap.Flags == 0 {
-				req.Flags = uint16(TUNTAP_MULTI_QUEUE_DEFAULTS)
+				flags = uint16(TUNTAP_MULTI_QUEUE_DEFAULTS)
 			}
 		}
+		flags |= uint16(tuntap.Mode)
 
-		req.Flags |= uint16(tuntap.Mode)
+		req, err := unix.NewIfreq(base.Name)
+		if err != nil {
+			return fmt.Errorf("tuntap name %q is invalid: %w", base.Name, err)
+		}
+		req.SetUint16(flags)
+
+		var fds []*os.File
 		const TUN = "/dev/net/tun"
 		for i := 0; i < queues; i++ {
-			localReq := req
+			iterReq := *req
 			fd, err := unix.Open(TUN, os.O_RDWR|syscall.O_CLOEXEC, 0)
 			if err != nil {
 				cleanupFds(fds)
 				return err
 			}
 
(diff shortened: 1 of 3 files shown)


link_linux.go
```diff
@@ -1493,47 +1493,52 @@ func (h *Handle) linkModify(link Link, flags int) error {
...
-				req.Flags = uint16(TUNTAP_MULTI_QUEUE_DEFAULTS)
+				flags = uint16(TUNTAP_MULTI_QUEUE_DEFAULTS)
 			}
 		}
+		flags |= uint16(tuntap.Mode)
+
+		// Reject names that do not fit in ifr_name with a trailing NUL
+		// (IFNAMSIZ-1). The previous copy(req.Name[:15], ...) truncated
+		// silently, so LinkAdd succeeded while the kernel device used a
+		// different name than the caller stored.
```
aboch: Can you please remove this comment? 

The method documentations says it all already, not to mention the returned error msg below.

Image

I would like to reserve comments for what strictly necessary.
Otherwise there will be soon more comment lines than code.

FakeLearne: Removed the comment.

aboch: @FakeLearne pl squash your commits into one

FakeLearne: Done

PR #1215, opened 2026-08-29
ipset: avoid panic on non-errno errors

rs:
## Summary

Avoid an unchecked `syscall.Errno` assertion in `ipsetExecute`.

`NetlinkRequest.Execute` may also return dump-interruption errors,
validation errors, socket/file errors, or wrapped errno values carrying
extended acknowledgment text. The previous assertion panicked for those
errors.

Only direct private ipset errno values are now converted to
`nl.IPSetError`. All other errors are returned unchanged.

## Testing

- Added coverage for private and ordinary errno values.
- Added coverage ensuring wrapped and ordinary errors retain their identity.
- Added a closed-netlink-socket regression test reproducing the previous panic.
- Ran the complete privileged ipset test suite, including IPv4 and IPv6 cases.

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented certain network configuration errors from causing application crashes.
  * Improved handling and reporting of unexpected error types.
  * Preserved the original error when it cannot be converted to a recognized IP set error.

* **Tests**
  * Added coverage for standard, wrapped, and unexpected error scenarios.

### DIFF
--- a/ipset_linux.go
+++ b/ipset_linux.go
@@ -470,13 +470,15 @@ func getIpsetDefaultRevision(typename string, featureFlags uint32) uint8 {
 
 func ipsetExecute(req *nl.NetlinkRequest) (msgs [][]byte, err error) {
 	msgs, err = req.Execute(unix.NETLINK_NETFILTER, 0)
+	return msgs, ipsetError(err)
+}
 
-	if err != nil {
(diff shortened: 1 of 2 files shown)


aboch:
LGTM

PR #1221, opened 2026-09-14
Add an option to Zero Counters when listing conntrack entries

imbstack:
> [!IMPORTANT]
> Remember, this is a public fork but the PR is being made internal to its own branch. If we like this and test it out internally, we can make a new PR into the upstream. 

Currently this library always uses `IPCTNL_MSG_CT_GET` to list conntrack entries but the kernel also has `IPCTNL_MSG_CT_GET_CTRZERO` which resets the counters on reading which is how we scrape safely from netmon.

This change adds that as an option to a new `WithOptions` version of the list function and threads it through. I understand and hand-wrote all of the code but had Claude generate some tests around this.

## Summary by CodeRabbit

- **New Features**
  - Added configurable conntrack table listing.
  - You can optionally reset flow counters while retrieving conntrack entries.
  - Standard listings continue to preserve existing counters.
  - Counter reset requests are rejected for unsupported table types.
  - Added consistent APIs across supported and unsupported platforms; unsupported platforms report that the operation is unavailable.

- **Documentation**
  - Improved formatting and readability of conntrack protocol documentation.


imbstack:
Sorry, meant to open this on my own fork!

